PR #7, opened 2023-06-05
Add pull request template

ori-sky:
This patch adds a pull request template in order to make it clearer what information should be included when opening pull requests with code contributions.

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,0 +1,32 @@
+# Overview
+
+*Provide a brief overview of what your changes do, summarizing their effects
+and consequences.*
+
+# Reason for change
+
+*Describe how the current behaviour of the project is causing problems for you
+or is otherwise unsatisfactory for your use case.*
+
+# Description of change
+
+*Describe the intended behaviour your changes are meant to introduce to the
+project and explain how they resolve the problem stated above. Detail any
+relevant changes that may affect other users of the project, such as
+compilation options, runtime flags, expected inputs and outputs, API entry
+points, etc.*
+
+*If you have added new testing, provide details on what tests you have added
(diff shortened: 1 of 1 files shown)


pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+affects any compilation options or runtime flags, or changes expected
+input/output values, please document these effects here.*
+
+# Anything else we should know?
+
+* If there's any other relevant information we should know that may help us in understanding and verifying your patch, please include it here.
+
+# Checklist
+
+- [ ] I agree to follow the project [Code of Conduct](https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/CODE_OF_CONDUCT.md).
```
coldav, line 23: I'm not sure we need a checklist. Is it possible to just add these as a comment?

pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+effects and consequences of the change.*
+
+# Current behaviour
+
+*Describe the current behaviour of the project and how it differs from the
+behaviour your change is proposing.*
+
+# Expected behaviour
+
+*Describe the behaviour your change is proposing to the project. If your change
```
coldav, line 13: These two fields blend into other - the current behaviour talks about how it differs - maybe that part should be dropped.
Also i think Expected is wrong - Describe the behaviour change may be a more accurate title.

pull_request_template.md
ori-sky: Thanks

pull_request_template.md
ori-sky: I've updated the sections now, does this look clearer?

coldav (approved):

PR #8, opened 2023-06-05
Remove old GitLab CI file

ori-sky:
This patch removes a GitLab CI file that was left over from before the project was published as open-source. This file is no longer in use so should be removed.


frasercrmck (approved):
LGTM

PR #12, opened 2023-06-06
[compiler] Allow muxc to compile OpenCL C inputs

frasercrmck:
This is part of the transition away from snapshots, by enhancing `muxc`'s capabilities for testing OpenCL C inputs, and running customized pass pipelines on them.

The changes to `muxc` itself are mostly moving code around to account for different input options, and breaking functions up to return their errors directly. The workhorse is the new `BaseModule::compileOpenCLCToIR` which has been created out of the guts of `BaseModule::compileOpenCLC`. The latter method wasn't suitable as-is, because it relied and set too much internal state to be suitable, and `muxc` wasn't able to fetch the resulting `llvm::Module` from it.

No users or derived users of the stock `BaseModule` should be affected by this, but I've bumped the mux spec and main version to 3.0.0.

I've moved `compiler/test/passes/lit` to `compiler/test/lit/passes` for better hierarchy, and to fit 'frontend' tests underneath the main lit suite.

`muxc` now output text by default, which is by far the most common use of `muxc` in testing. Indeed, we don't have any in-tree uses of `muxc` that output object files.


modules/compiler/tools/muxc/muxc.cpp
```diff
@@ -44,11 +49,28 @@ static cl::opt<std::string> OutputFilename(
...
+    cl::init(true));
+
+static cl::opt<CodeGenFileType> FileType(
+    "filetype", cl::init(CGFT_AssemblyFile), cl::desc("Choose a file type:"),
+    cl::values(clEnumValN(CGFT_AssemblyFile, "asm", "Emit a textual file"),
+               clEnumValN(CGFT_ObjectFile, "obj", "Emit a binary object file"),
+               clEnumValN(CGFT_Null, "null",
+                          "Emit nothing, for performance testing")));
+
+static cl::opt<int> DeviceIdx(
```
coldav, line 63: I think this is sufficiently big, it should be mentioned in the commit message.

coldav (dismissed):

coldav (approved):

PR #17, opened 2023-06-07
[host] Move pass pipelines into HostPassMachinery

frasercrmck:
This centralizes the concept and construction of host pass pipelines
into HostPassMachinery, which means that muxc can easily run the same
pipelines.

We provide two pipeline components to muxc:

* host-late-passes
* host-kernel-passes<unique-name>

The `host-late-passes` pipeline runs the equivalent of
`HostModule::getLateTargetPasses`, which doesn't do much on host.

The `host-kernel-passes` runs the 'real' passes which are usually run
during jitting or offline compilation. This optionally takes
`unique-name` as a parameter, which schedules the pass to make the
kernel's name globally unique (if you pass a unique name...).

This requires the pass machinery to know about compiler options, so we
provide that functionality to all BaseModulePassMachinery instances.
These can be set multiple times between compilations in case a pass
machinery is being reused.

This functionality could be easily extended to riscv but that can come
in a future update.


modules/compiler/targets/host/source/HostPassMachinery.cpp
```diff
@@ -147,6 +166,180 @@ void HostPassMachinery::registerPassCallbacks() {
...
+
+  PM.addPass(vecz::RunVeczPass());
+
+  compiler::BaseModule::addSnapshotPassIfEnabled(
+      PM, "host", HOST_SNAPSHOT_VECTORIZED, snapshots);
+
+  addLateBuiltinsPasses(PM, tuner);
+
+  compiler::utils::HandleBarriersOptions HBOpts;
+  // Barriers with opt disabled are broken on 32 bit for some llvm versions, see
```
coldav, line 262: I wonder if this is still true on 15+ - although I appreciate this is just moving code

coldav (approved):

modules/compiler/targets/host/source/HostPassMachinery.cpp
frasercrmck: Good question - I might try dropping this and running CI over it. The ticket isn't super descriptive.

PR #18, opened 2023-06-07
Remove snapshots and the cl_codeplay_program_snapshot extension

frasercrmck:
These were originally intended for internal developer use, and don't
make sense as a public extension. They have been internally superceded
by more LLVM-like debugging options, which offer more flexibility.


modules/compiler/targets/host/include/host/host_pass_machinery.h
```diff
@@ -45,6 +47,22 @@ class HostPassMachinery final : public compiler::BaseModulePassMachinery {
   void registerPasses() override;
 
   void printPassNames(llvm::raw_ostream &) override;
+
+  bool handlePipelineElement(llvm::StringRef,
+                             llvm::ModulePassManager &AM) override;
+
+  /// @brief Returns an optimization pass pipepline to either all kernels in a
```
coldav, line 54: pipepline ->pipeline

frasercrmck: resolved in other PR: #17

PR #20, opened 2023-06-12
Added workflow for html documentation

coldav:
# Overview

Added a pull request workflow for testing documentation

# Reason for change

There is no current PR testing for building the documentation.

# Description of change

 This adds a workflow which will build it and create an artifact, but not deploy it which will require a different workflow.

# Checklist

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,45 @@
+# Simple workflow for building documentation
+name: Build documentation
+
+# On a pull request this will build the html documentation
+# and create an artifact. It does this by using the cmake
+# build target doc_html. Note that this requires installing a little
+# more than is technically required for building the documentation in
+# order to satisfy our current cmake configurations e.g. spirv-tools
+
+on:
+ pull_request:
+
+  # Allows you to run this workflow manually from the Actions tab
+ workflow_dispatch:
+
+jobs:
+  # Build job
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+
+      - name: Install apt package
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y doxygen virtualenv
+
+      - name: Install prerequisites
+        run: python -m pip install -r doc/requirements.txt
+
+      - name: Build Documentation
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.9
+
+      - name: Install apt package
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y doxygen spirv-tools virtualenv
```
frasercrmck, line 29: We need spirv-tools to build the documentation?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Install prerequisites
+        run: python3 -m pip install -r doc/requirements.txt
```
frasercrmck, line 32: Do we need `python3` here or will `setup-python` create `python = python3` for us? Just trying to reduce the number of places we're encoding the version.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Build Documentation
+        working-directory: ${{github.workspace}}
+        run: |
+          cmake -DCA_ENABLE_API=cl -DCMAKE_BUILD_TYPE=Debug -DCA_RUNTIME_COMPILER_ENABLED=OFF -DCA_CL_ENABLE_OFFLINE_KERNEL_TESTS=OFF -DOCL_EXTENSION_cl_khr_il_program=OFF -Bbuild_doc
+          cmake --build build_doc  --target doc_html
```
frasercrmck, line 38: Double space before `--target`

.github/workflows/docs.yml
```diff
@@ -0,0 +1,43 @@
+# Simple workflow for building documentation
+name: Build documentation
+
+# On a pull request this will build the html documentation
+# and create an artifact. It does this by using the standard
```
frasercrmck, line 5: nit: "standard" build target made it sound to me like a github thing. I think we could be clearer here.

.github/workflows/docs.yml
coldav: it seems to be required for the cmake - I agree that it shouldn't be needed. We may want to detach the documentation from the build in the future.

frasercrmck (approved):

PR #30, opened 2023-06-14
[spirv-ll] Support an 'online' lit testing mode

frasercrmck:
This adds two modes to the spirv-ll lit testing infrastructure: offline and online.

The offline mode is the default, and is essentially the current behaviour - running `check-spirv-ll-lit` will build all the necessary SPIR-V binaries ahead of time and run tests which expect those to be in place.

The new online mode allows faster turnaround times on testing, and allows the project to be built without the necessary tools known at build time. In this mode, the tools for spvasm/glsl are checked for independently at runtime and a feature is enabled for each if the tools are resolved. Under this feature, a conditional RUN line is enabled to assemble the spvasm/glsl file to SPIR-V. This uses lit's built-in `%if` syntax.

A bonus for the online mode is that test binaries can now more easily be assembled individually, e.g., taking certain options on a per-test basis.

This behaviour is more intuitive for developers as iterating on individual test files no longer requires specific build targets to be built before the changes to the SPIR-V binary are made apparent.It also suits building the project in a system where the tools aren't necessarily required at build time.

Lit can be told to switch to either mode at runtime, using `--param
spirv-ll-online=1/true/0/false/`. Omitting the parameter leaves the test
suite in the mode in which it was configured at build-time.

### DIFF
--- a/modules/compiler/spirv-ll/test/CMakeLists.txt
+++ b/modules/compiler/spirv-ll/test/CMakeLists.txt
@@ -15,35 +15,41 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 set(SUITE spirv-ll)
 
-find_program(glslangValidator_EXECUTABLE
-  NAMES glslangValidator PATHS ENV VULKAN_SDK PATH_SUFFIXES bin
-  CMAKE_FIND_ROOT_PATH_BOTH)
-if(glslangValidator_EXECUTABLE STREQUAL
-    glslangValidator_EXECUTABLE-NOTFOUND)
-  set(GLSL_UNSUPPORTED True)
-  message(WARNING "${SUITE} glsl lit tests unsupported: glslangValidator not found")
-else()
-  set(GLSL_UNSUPPORTED False)
-endif()
+ca_option(CA_ASSEMBLE_SPIRV_LL_LIT_TESTS_OFFLINE STRING
+  "Offline-assemble spirv-ll lit spvasm/glsl test binaries" ON)
 
-find_package(SpirvTools COMPONENTS spirv-as)
-if(TARGET spirv::spirv-as)
-  # Create custom commands to assemble .spvasm files in .spv.
-  set(SPVASM_UNSUPPORTED False)
-  # Get the found spirv-as version, takes the form: v<year>.<release>
-  execute_process(
-    COMMAND ${SpirvTools_spirv-as_EXECUTABLE} --version
-    OUTPUT_VARIABLE SpirvAsVersionOutput)
-  string(REGEX MATCH "v20[0-9][0-9].[0-9]+"
-    SpirvAsVersion ${SpirvAsVersionOutput})
-  string(REGEX MATCH "20[0-9][0-9]"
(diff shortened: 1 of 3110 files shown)


modules/compiler/spirv-ll/test/CMakeLists.txt
```diff
@@ -15,35 +15,41 @@
...
-else()
-  set(GLSL_UNSUPPORTED False)
-endif()
+ca_option(CA_ASSEMBLE_SPIRV_LL_LIT_TESTS_OFFLINE STRING
+  "Offline-assemble spirv-ll lit spvasm/glsl test binaries" ON)
+
+set(GLSL_UNSUPPORTED False)
+set(SPVASM_UNSUPPORTED False)
+
+# Check for tools at build time if asked to assembly everything offline.
```
coldav, line 24: assembly -> assemble

modules/compiler/spirv-ll/test/README.md
```diff
@@ -3,4 +3,44 @@
...
+time.
+
+This mode requires `spirv-as` to build the `spvasm` test binaries, and
+`glslangValidator` to build the `glsl` test binaries. Both are optional: if
+either tool is not found the tests will be disabled at runtime.
+
+### Online Mode
+
+In offline mode - when `CA_ASSEMBLE_SPIRV_LL_LIT_TESTS_OFFLINE`
+is disable in cmake - the test suite will build its SPIR-V binaries at test
```
coldav, line 23: disabled

modules/compiler/spirv-ll/test/README.md
```diff
@@ -3,4 +3,44 @@
...
+
+It is possible to switch between modes at runtime by setting the
+``spirv-ll-online`` parameter:
+
+```
+> ca-lit --param spirv-ll-online=0 modules/compiler/spirv-ll/test
+> ca-lit --param spirv-ll-online=1 modules/compiler/spirv-ll/test
+```
+
+Explicitly enabling the option with a truthy value forcibly enables the online
```
coldav, line 41: Is there any way to make this happen with the check-*-lit target?

modules/compiler/spirv-ll/test/README.md
```diff
@@ -3,4 +3,44 @@
...
+mode. Explicitly disabling the option with a falsey value forcibly enables
```
coldav, line 42: false

coldav (approved):

modules/compiler/spirv-ll/test/README.md
frasercrmck: yes, with `LIT_OPTS` - my mind blanked when writing this and just moved on. Added now.

modules/compiler/spirv-ll/test/README.md
frasercrmck: I mean `falsy` as in anything that can be evaluated as false: 0, off, false, etc. That's how the parameter is interpreted. I have `falsey` but apparently `falsy` is more common so I will use that.

coldav: I'd never heard of it, I'm not sure it's a term used in the context of cmake variables, but I'm okay if you want to go with it.

PR #32, opened 2023-06-15
[spirv-ll] Add support for SPV_INTEL_optnone

MaryaSharf:
# Overview

# Reason for change

To be able to run tests with yet to release oneAPI packages, support for SPV_INTEL_optnone is required

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/modules/compiler/spirv-ll/source/builder_core.cpp
+++ b/modules/compiler/spirv-ll/source/builder_core.cpp
@@ -1652,9 +1652,11 @@ cargo::optional<Error> Builder::create<OpFunction>(const OpFunction *op) {
 
   if (op->FunctionControl() & spv::FunctionControlInlineMask) {
     linkage = llvm::Function::LinkageTypes::LinkOnceODRLinkage;
-    llvm::Function *llvmFunction = IRBuilder.GetInsertBlock()->getParent();
-    llvmFunction->addFnAttr(llvm::Attribute::OptimizeNone);
-    llvmFunction->addFnAttr(llvm::Attribute::NoInline);
+    if (module.isExtensionEnabled("SPV_INTEL_optnone")) {
+      llvm::Function *llvmFunction = IRBuilder.GetInsertBlock()->getParent();
+      llvmFunction->addFnAttr(llvm::Attribute::OptimizeNone);
+      llvmFunction->addFnAttr(llvm::Attribute::NoInline);
+    }
   } else if (auto linkageInfo = getLinkage(module, op->IdResult())) {
     if (linkageInfo->first == spv::LinkageTypeImport ||
         linkageInfo->first == spv::LinkageTypeExport ||
--- a/modules/compiler/spirv-ll/source/builder_core.cpp
+++ b/modules/compiler/spirv-ll/source/builder_core.cpp
@@ -1652,6 +1652,9 @@ cargo::optional<Error> Builder::create<OpFunction>(const OpFunction *op) {
 
   if (op->FunctionControl() & spv::FunctionControlInlineMask) {
     linkage = llvm::Function::LinkageTypes::LinkOnceODRLinkage;
+    llvm::Function *llvmFunction = IRBuilder.GetInsertBlock()->getParent();
+    llvmFunction->addFnAttr(llvm::Attribute::OptimizeNone);
+    llvmFunction->addFnAttr(llvm::Attribute::NoInline);
   } else if (auto linkageInfo = getLinkage(module, op->IdResult())) {
     if (linkageInfo->first == spv::LinkageTypeImport ||
         linkageInfo->first == spv::LinkageTypeExport ||
--- a/modules/compiler/spirv-ll/source/builder_core.cpp
+++ b/modules/compiler/spirv-ll/source/builder_core.cpp
@@ -1946,13 +1946,11 @@ cargo::optional<Error> Builder::create<OpFunction>(const OpFunction *op) {
 
   if (op->FunctionControl() & spv::FunctionControlOptNoneINTELMask) {
     SPIRV_LL_ASSERT(
-            module.isExtensionEnabled("SPV_INTEL_optnone"),
-            "SPV_INTEL_optnone must be enabled.");
-    if (module.isExtensionEnabled("SPV_INTEL_optnone")) {
-      llvm::Function *llvmFunction = IRBuilder.GetInsertBlock()->getParent();
-      llvmFunction->addFnAttr(llvm::Attribute::OptimizeNone);
-      llvmFunction->addFnAttr(llvm::Attribute::NoInline);
-    }
+            module.hasCapability(spv::CapabilityOptNoneINTEL),
+            "CapabilityOptNoneINTEL must be enabled.");
+    llvm::Function *llvmFunction = IRBuilder.GetInsertBlock()->getParent();
+    llvmFunction->addFnAttr(llvm::Attribute::OptimizeNone);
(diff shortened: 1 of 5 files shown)


modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -1652,6 +1652,11 @@ cargo::optional<Error> Builder::create<OpFunction>(const OpFunction *op) {
 
   if (op->FunctionControl() & spv::FunctionControlInlineMask) {
     linkage = llvm::Function::LinkageTypes::LinkOnceODRLinkage;
+    if (module.isExtensionEnabled("SPV_INTEL_optnone")) {
```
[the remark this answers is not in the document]
frasercrmck: For validation, note that the spec says that the `OptNoneINTEL` `Function Control` bit is dependent on the `OptNoneINTEL` capability, not the extension. The extension does enable use of the capability, but I don't know if that's as important to validate. My knowledge of the extent of our SPIR-V validation coverage is a little weak, I'm afraid. Maybe we can/should catch invalid capabilities (that haven't declared the appropriate extension) where capablities are declared in the module?

Here I think the validation can be inverted; that is, checking whether the `OptNone` `FunctionControl`  bit has been set first, _then_ asserting that the Capability is enabled. We want to catch invalid binaries that set the bit without declaring the capability.

frasercrmck (commented):
In general the PR also needs some lit tests. You'll need to add code to `spirv-ll-tool` to recognize that capability on the command line, and maybe to enable the `SPV_INTEL_optnone` extension in the "enable all" option?

modules/compiler/spirv-ll/source/builder_core.cpp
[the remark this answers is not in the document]
MaryaSharf: I am still learning about the spirv, but I'be updated the code according to what I understood from your comments. 
I'll have a look at the test next.

modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -1949,6 +1944,17 @@ cargo::optional<Error> Builder::create<OpFunction>(const OpFunction *op) {
 
   SPIRV_LL_ASSERT_PTR(function);
 
+  if (op->FunctionControl() & spv::FunctionControlOptNoneINTELMask) {
+    SPIRV_LL_ASSERT(
+            module.isExtensionEnabled("SPV_INTEL_optnone"),
+            "SPV_INTEL_optnone must be enabled.");
+    if (module.isExtensionEnabled("SPV_INTEL_optnone")) {
```
frasercrmck: Can probably get rid of this `if` given we've asserted above.

frasercrmck (commented):
Looks good otherwise, just needs some tests. Copy/pasting any of the existing tests that deal with FunctionControl bits. Maybe grep for a simple test that uses `DontInline` or `Inline` and update it to use `OptNoneINTEL`?

modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -1949,6 +1944,17 @@ cargo::optional<Error> Builder::create<OpFunction>(const OpFunction *op) {
 
   SPIRV_LL_ASSERT_PTR(function);
 
+  if (op->FunctionControl() & spv::FunctionControlOptNoneINTELMask) {
+    SPIRV_LL_ASSERT(
```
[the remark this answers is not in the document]
MaryaSharf: I am confused, isn't that what I did?

modules/compiler/spirv-ll/source/builder_core.cpp
MaryaSharf: I think you're looking at my previous commit. I'll rebase and clean up.

modules/compiler/spirv-ll/source/builder_core.cpp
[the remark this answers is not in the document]
frasercrmck: Oh sorry, yeah maybe I was looking at an old version.

modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -1944,6 +1944,15 @@ cargo::optional<Error> Builder::create<OpFunction>(const OpFunction *op) {
...
+            module.hasCapability(spv::CapabilityOptNoneINTEL),
+            "CapabilityOptNoneINTEL must be enabled.");
+    llvm::Function *llvmFunction = IRBuilder.GetInsertBlock()->getParent();
```
frasercrmck, line 1951: in case my comment was lost - this can just be `function->addFnAttr` ...

modules/compiler/spirv-ll/test/spvasm/intel_opt_none.spvasm
```diff
@@ -0,0 +1,36 @@
...
+; RUN: %if online-spirv-as %{ spirv-as --target-env %spv_tgt_env -o %spv_file_s %s %}
+; RUN: spirv-ll-tool -a OpenCL -b 64 %spv_file_s | FileCheck %s
+               OpCapability Addresses
+               OpCapability Linkage
+               OpCapability Kernel
+               OpCapability OptNoneINTEL
+          %1 = OpExtInstImport "OpenCL.std"
+               OpMemoryModel Physical64 OpenCL
+               OpEntryPoint Kernel %4 "foo"
+               OpExecutionMode %4 VecTypeHint 0
```
frasercrmck, line 25: You don't need this `OpExecutionMode`.

modules/compiler/spirv-ll/test/spvasm/intel_opt_none.spvasm
```diff
@@ -0,0 +1,36 @@
...
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+; RUN: %if online-spirv-as %{ spirv-as --target-env %spv_tgt_env -o %spv_file_s %s %}
+; RUN: spirv-ll-tool -a OpenCL -b 64 %spv_file_s | FileCheck %s
+               OpCapability Addresses
+               OpCapability Linkage
```
frasercrmck, line 19: I don't think you need this capability

modules/compiler/spirv-ll/test/spvasm/intel_opt_none.spvasm
```diff
@@ -0,0 +1,36 @@
...
+          %6 = OpString "kernel_arg_type.foo."
+               OpSource OpenCL_C 102000
+               OpName %entry "entry"
+       %void = OpTypeVoid
+          %3 = OpTypeFunction %void
+          %4 = OpFunction %void OptNoneINTEL %3
+      %entry = OpLabel
+               OpReturn
+               OpFunctionEnd
+; CHECK-LABEL: foo
```
frasercrmck, line 35: best to check the attributes on `foo` with something like `CHECK: define void @foo() [[ATTRS:#[0-9]+]]` (adjusting for the actual signature of `foo`) and then `CHECK: attributes [[ATTRS]] = { noinline optnone }` to ensure you're definitely matching the attributes on the right function

frasercrmck (approved):
Looks good with that feedback addressed

MaryaSharf:
closing in favour of https://github.com/codeplaysoftware/oneapi-construction-kit/pull/35

PR #33, opened 2023-06-19
Remove fuzzy_testing scripts

ori-sky:
# Overview

This patch removes the scripts for fuzzy testing as these are not currently in use and trigger security vulnerability alerts from Dependabot.

# Reason for change

Dependabot alerts us to potential security vulnerabilities in our fuzzy testing scripts, so we have decided to remove these scripts for the time being in order to resolve the alerts.

# Description of change

Fuzzy testing has been removed.

# Checklist


coldav (approved):

coldav:
> Were these scripts the only use of `FuzzCL`? Maybe we should deprecate or remove that too?

I haven't found any other uses.

frasercrmck:
> > Were these scripts the only use of `FuzzCL`? Maybe we should deprecate or remove that too?
> 
> I haven't found any other uses.

I might be mistaken - FuzzCL is run by `scripts/testing/run_fuzzcl.py` which this PR _isn't_ removing.

frasercrmck:
> > > Were these scripts the only use of `FuzzCL`? Maybe we should deprecate or remove that too?
> > 
> > 
> > I haven't found any other uses.
> 
> I might be mistaken - FuzzCL is run by `scripts/testing/run_fuzzcl.py` which this PR _isn't_ removing.

... which might suggest the PR title is misleading as we're not removing *all* fuzzy testing scripts

PR #34, opened 2023-06-19
Migrate to compiler::KernelInfo and introduce compiler::ProgramInfo

frasercrmck:
[cl] Replace cl::binary::KernelInfo with compiler::KernelInfo

    We were duplicating this information across two places, which is overly
    cumbersome. Since the top-level compiler APIs are unconditionally
    included, even in 'offline' builds, we can safely use
    `compiler::KernelInfo` in all cases. At certain points we need to
    translate between CL and compiler representations, but we were doing
    that already.

----

    [compiler] Add compiler::ProgramInfo and migrate CL/UR APIs to that

    After replacing cl::binary::KernelInfo with compiler::KernelInfo, the
    program info callback we passed to Module::finalize was essentially
    redundant: it was copying compiler::KernelInfo instances to other
    compiler::KernelInfo instances.

    With that in mind, we can clean the codebase up further by introducing a
    central compiler::ProgramInfo and make Module::finalize populate that.
    Then clients can opt to use that directly (as CL and UR do) or they
    could translate the compiler data structures to their client ones.

### DIFF
--- a/modules/compiler/include/compiler/module.h
+++ b/modules/compiler/include/compiler/module.h
@@ -325,9 +325,9 @@ struct ArgumentType {
 struct KernelInfo {
   /// @brief Struct representing basic kernel argument information.
   struct ArgumentInfo {
-    AddressSpace address_qual;
-    KernelArgAccess access_qual;
-    std::uint32_t type_qual;
+    AddressSpace address_qual = compiler::AddressSpace::PRIVATE;
+    KernelArgAccess access_qual = compiler::KernelArgAccess::NONE;
+    std::uint32_t type_qual = 0;
     std::string type_name;
     std::string name;
   };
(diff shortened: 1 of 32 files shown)


modules/compiler/include/compiler/module.h
```diff
@@ -339,11 +339,117 @@ struct KernelInfo {
...
+  /// @brief Returns the reqd_work_group_size if present, else an all-zeros
+  /// array.
+  std::array<size_t, 3> getReqdWGSizeOrZero() const {
+    return reqd_work_group_size.value_or(std::array<size_t, 3>{0, 0, 0});
+  }
 };  // class KernelInfo
 
-/// @brief Kernel info callback type.
-using KernelInfoCallback = std::function<void(KernelInfo)>;
+/// @brief Class for managing program information.
```
coldav, line 353: I think this could say a little more, what kind of information. it seems to be a container of multiple kernel informations (and not much else).

coldav (approved):

modules/compiler/include/compiler/module.h
frasercrmck: good idea. yeah that's basically it. I stole it from the CL side of things, which doesn't need anything more than that.

PR #37, opened 2023-06-21
[compiler] Support cl_intel_required_subgroup_size

frasercrmck:
Note: this commit does not yet enable the extension by default! See
below.

This commit adds support for the cl_intel_required_subgroup_size OpenCL
extension.

* It supports device-specific lists of sub-group sizes by adding two new
  fields to the Mux device info available to targets: a pointer to a
  list of sub-group sizes, and a size for the list. The pointer should
  point to static/managed data.
* It supports the kernel data necessary for the extension by extending
  `compiler::KernelInfo` with the (optional) required sub-group size
  and the 'spill mem size'. The required sub-group size is
  encoded/decoded through the binary header. The spill size is not, but
  that's in keeping with the private mem size - this might indicate an
  oversight for offline-compiled binaries.

The rest of the work is plumbing that data into the extension's queries,
and adding testing coverage.

Two new compiler passes have been added to force the compiler into
failing if the required sub-group size is not supported by the device,
or if the requirement has not been satisfied by the compiler. Both of
these are always true as none of our targets advertise any sub-group
sizes, and never force the vectorizer into vectorizing to that size.

The `muxc` compiler tool can be given supported sub-group sizes via a
command-line option to ensure these compiler passes can be tested in a
device-agnostic way.

Note that this doesn't include support for this extension when the
program was compiled with SPIR-V (clCreateProgramWithILKHR). In a strict
reading of the extension spec, `CL_KERNEL_COMPILE_SUB_GROUP_SIZE_INTEL`
only reports the value of the attribute and thus has no meaning for
SPIR-V, but we can be charitable here.

It is expected that this property will manifest itself via the
`SubgroupSize` execution mode, but more work needs to be done here. For
that reason the extension is left disabled until SPIR-V support is
added.

### DIFF
--- a/doc/source/cl/extension.rst
+++ b/doc/source/cl/extension.rst
@@ -13,7 +13,7 @@ EXT
 Vendor
   Extension defined by a single vendor, but may be implemented by
   other vendors, e.g oneAPI Construction Kit implements vendor extension
-  ``cl_intel_unified_shared_memory``.
+  ``cl_intel_unified_shared_memory`` and ``cl_intel_required_subgroup_size``.
 
 oneAPI Construction Kit implements several Codeplay vendor extensions,
 specified under the ``extension`` directory. When adding a new vendor
@@ -49,6 +49,7 @@ be contiguous.
   extension/cl_codeplay_soft_math
   extension/cl_codeplay_wfv
   extension/cl_intel_unified_shared_memory
+  extension/cl_intel_required_subgroup_size
 
 OpenCL C 1.2 - ``khr_opencl_c_1_2``
 -----------------------------------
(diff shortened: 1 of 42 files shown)


doc/source/cl/extension.rst
```diff
@@ -351,3 +352,21 @@ replaced by a customer implementation where it can be accelerated.
...
+extension.
+
+This has a default implementation which does not report any available subgroup
+sizes for any device. The compiler will report an error for any kernel given
+the ``intel_reqd_sub_group_size`` attribute with a size that is not reported by
+the device. There is currently no handling of the attribute if a target does
+report a set of sub-group sizes, as no in-tree target does so.
+
+Furthermore, all kernels report the same non-zero value for
+``CL_KERNEL_SPILL_MEM_SIZE_INTEL``.
```
coldav, line 369: is this a typo? I can't work out how this fits.

coldav: ah I see it's a strange part of the extension. why is it non-zero - the spec seems to allow for zero meaning no spill happened.

coldav: feel free to leave as is, but I think a few extra words to say why it's a non-zero might help here.

frasercrmck: I think it's better for us, without any information, to assume that some spill happened, rather than baselessly guarantee to the user didn't.

PR #38, opened 2023-06-21
Support SPIR-V 'SubgroupSize' and enable cl_intel_required_subgroup_size by default

frasercrmck:
----

[spirv-ll] Support the 'SubgroupSize' ExecutionMode

This commit extends our SPIR-V consumer to understand the `SubgroupSize`
execution mode. It does so to accomodate both higher-level standards
like SYCL 2020 which have a `reqd_sub_group_size` attribute and OpenCL
which has `intel_reqd_sub_group_size` (via an extension). Although
there's no mandate that this execution mode should encode this data,
that's what tools like DPC++ do in practice and so we should handle
that.

This is awkward for us, as `SubgroupSize` comes as part of the
`SubgroupDispatch` capability, which in turn implicitly defines the
`DeviceEnqueue` capability. The OpenCL environment spec explicitly says
we are not required to support `DeviceEnqueue` as we don't support
device-side enqueue, but it also says that environments supporting
SPIR-V v1.1 binaries should support `SubgroupDispatch` for OpenCL 3.0
devices supporting subgroups. This indicates a spec bug, and is
something we'll raise and hopefully fix in a future version of the
specs.

As a result, it's now perfectly possible for us to encounter SPIR-V
binaries that declare `SubgroupDispatch` and thus are implicitly allowed
to use device-side enqueue features. I feel it's better if we're strict
here and, rather than silently ignoring such features, we throw an error
to let users know what's going on. I've done that, and added a bit of
testing to cover it: full lit testing for each individual thing we don't
support would require a lot of noisy tests.

This also interops with the cl_intel_required_subgroup_size OpenCL
extension, meaning the extension should now also work with SPIR-V
programs. With that, the extension can be considered complete and safe
to enable by default. That will be done in a follow-up commit.

----

[CL] Enable cl_intel_required_subgroup_size by default.

Now that SPIR-V support has been plumbed in, this extension should be
good to enable by default.

### DIFF
--- a/modules/compiler/spirv-ll/source/builder_core.cpp
+++ b/modules/compiler/spirv-ll/source/builder_core.cpp
@@ -621,12 +621,19 @@ cargo::optional<Error> Builder::create<OpTypeEvent>(const OpTypeEvent *op) {
   return cargo::nullopt;
 }
 
+static Error errorUnsupportedDeviceEnqueueOp(const std::string &opName) {
+  // Capability DeviceEnqueue isn't supported by CL 1.2, see OpenCL SPIR-V
+  // environment spec section 3.1 for supported capabilities.
+  // It is, however, implicitly declared by SPIR-V 1.1 modules which declare the
+  // SubgroupDispatch capability for CL 3.0 devices supporting subgroups. This
+  // is highly dubious, and appears like a spec bug of some kind.
+  return Error(opName + " is not supported.");
+}
+
 template <>
 cargo::optional<Error> Builder::create<OpTypeDeviceEvent>(
     const OpTypeDeviceEvent *) {
-  // Capability DeviceEnqueue isn't supported by CL 1.2, see OpenCL SPIR-V
-  // environment spec section 6.1 for supported capabilities.
-  return cargo::nullopt;
+  return errorUnsupportedDeviceEnqueueOp("OpTypeDeviceEvent");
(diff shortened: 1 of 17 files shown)


modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -621,12 +621,19 @@ cargo::optional<Error> Builder::create<OpTypeEvent>(const OpTypeEvent *op) {
   return cargo::nullopt;
 }
 
+static Error errorUnsupportedDeviceEnqueueOp(const std::string &opName) {
+  // Capability DeviceEnqueue isn't supported by CL 1.2, see OpenCL SPIR-V
+  // environment spec section 3.1 for supported capabilities.
+  // It is, however, implicitly declared by SPIR-V 1.1 modules which declare the
+  // SubgroupDispatch capability for CL 3.0 devices supporting subgroups. This
+  // is highly dubious, and appears like a spec bug of some kind.
```
coldav, line 629: Should there be a jira to address this?

coldav (approved):

modules/compiler/spirv-ll/source/builder_core.cpp
frasercrmck: I've created an internal ticket to look into this.

PR #44, opened 2023-06-26
Add building of code for PRs to seed sccache and run some PR testing for code

coldav:
# Overview

Added PR testing for host and riscv m1 on ubuntu

# Reason for change

This is required for the improvement of PR testing on github.

# Description of change

There are two main parts to the change:
  * We run a build run on main (with some testing) to seed an sccache
  * We run PR testing on pull requests which will read the sccache only

Both use composite actions to avoid too much repetition for building
of the oneapi-construction-kit.

sccache uses an action which has an underlying usage of github caches.
This is tied to a branch, so seeding it has to be done on main to be
available to pull requests. Every time the build PR worflow is run,
this will introduce a new timestamped sccache, so this will have to be
managed. Currently this is only run either manually or when the workflow
files are updated.

Much of the complexity is in the composite actions which can be called
from the workflows and other composite actions:
  * A common action which is uses to set the cmake and build the targets
  * An action specifically for building riscv targets

Note the common action for building can mean that it overly sets cmake
variables in order to allow lots of different inputs. This may benefit
from reviewing at a future date.

### DIFF
--- a/.github/actions/do_build_ock/action.yml
+++ b/.github/actions/do_build_ock/action.yml
@@ -1,0 +1,112 @@
+name: build-ock
+description: Action to build the oneapi-construction-kit
+
+# Some of these are riscv or host target specific, but it does not harm at present to
+# overset cmake values
+inputs:
+  build_type:
+    description: 'build type e.g Release, ReleaseAssert (default ReleaseAssert)'
+    default: ReleaseAssert
+  mux_targets_enable:
+    description: 'mux targets to enable e.g. riscv, host (default host)'
+    default: "host"
+  mux_compilers_enable:
+    description: 'mux compiler target to enable e.g. riscv, host (default host)'
+    default: "host"
+  usm:
+    description: 'Enable usm (default ON)'
+    default: ON
+  command_buffer:
+    description: 'Enable command buffers (default ON)'
+    default: ON
+  debug_support:
+    description: 'Enable debug support (default OFF)'
+    default: OFF
+  offline_kernel_tests:
+    description: 'Enable offline kernel testing (default ON)'
+    default: ON
+  llvm_install_dir:
+    description: 'Directory for llvm install'
+    default: ${{ github.workspace }}/llvm_install
+  build_targets:
+    description: 'cmake targets to build (default UnitCL muxc veczc clc)'
+    default: UnitCL muxc veczc clc
+  host_fp16:
+    description: 'Enable host fp16 (default OFF)'
+    default: OFF
+  host_image:
+    description: 'Enable host images (default OFF)'
+    default: OFF
+  host_enable_builtins:
+    description: 'Enable host builtins (default ON)'
+    default: ON
+  external_compiler_dirs:
+    description: 'External compiler directories, default is to empty string, which means no external directories'
+    default: ""
+  hal_description:
+    description: 'Description to be used for HAL, typically risc-v ISA (default RV64GCV)'
+    default: "RV64GCV"
+  hal_refsi_soc:
+    description: 'HAL Refsi SOC e.g. M1 or G1 default M1'
+    default: "M1"
+  riscv_enabled:
+    description: "Enable riscv target (default OFF)"
+    default: OFF
+  build_dir:
+    description: "Directory to be used for building"
+    default: build
+  extra_flags:
+    description: "Any additional cmake flags to be passed (default '')"
+    default:
+  assemble_spirv_ll_lit_test_offline:
+    description: "Enable Spir-V LL tests offline (Default OFF)"
+    default: OFF
+  enable_api:
+    description: "APIs to enable (default CL)"
+    default: cl
+  enable_rvv_scalable_vecz_check:
+    description: "Enable RVV scalable vecz check (default OFF)"
+    default: OFF
+  enable_rvv_scalable_VP_vecz_check:
+    description: "Enable RVV scalable vecz VP check (default OFF)"
+    default: OFF
+
+runs:
+  # We don't want a new docker just a list of steps, so mark as composite
+  using: "composite"
+  steps:
+    - name: cmake_ock
+      shell: bash
+      run:
+        cmake -GNinja
+              -B${{ inputs.build_dir }}
+              -DCMAKE_C_COMPILER_LAUNCHER=sccache
+              -DCMAKE_CXX_COMPILER_LAUNCHER=sccache                
+              -DCA_MUX_TARGETS_TO_ENABLE=${{ inputs.mux_targets_enable }}
+              -DCA_ENABLE_API=${{ inputs.enable_api }}
+              -DCA_LLVM_INSTALL_DIR=${{ inputs.llvm_install_dir }}
+              -DCA_ENABLE_DEBUG_SUPPORT=${{ inputs.debug_support }}
+              -DCMAKE_BUILD_TYPE=${{ inputs.build_type }}
+              -DCA_ENABLE_HOST_IMAGE_SUPPORT=${{ inputs.host_image }}
+              -DCA_HOST_ENABLE_BUILTIN_KERNEL=${{ inputs.debug_support }}
+              -DCA_HOST_ENABLE_BUILTINS_EXTENSION=${{ inputs.host_enable_builtins}}
+              -DCA_HOST_ENABLE_FP16=${{ inputs.host_fp16 }}
+              -DCA_CL_ENABLE_OFFLINE_KERNEL_TESTS=${{ inputs.offline_kernel_tests }}
+              -DCA_ASSEMBLE_SPIRV_LL_LIT_TESTS_OFFLINE=${{ inputs.assemble_spirv_ll_lit_test_offline }}
+              -DOCL_EXTENSION_cl_intel_unified_shared_memory=${{ inputs.usm }}
+              -DOCL_EXTENSION_cl_khr_command_buffer=${{ inputs.command_buffer }}
+              -DOCL_EXTENSION_cl_khr_command_buffer_mutable_dispatch=${{ inputs.command_buffer }}
+              -DCA_MUX_COMPILERS_TO_ENABLE=${{ inputs.mux_compilers_enable}}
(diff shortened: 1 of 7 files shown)


.github/actions/setup_ubuntu_build/action.yml
```diff
@@ -0,0 +1,47 @@
+name: setup-ubuntu-build
+description: Setup ubuntu ready for building. Includes installs, ninja, build cache setup and loading llvm cache
+
+inputs:
+  build_type:
+    description: 'build type (Release, RelAssert)'
+    default: RelAssert
```
frasercrmck, line 7: It's a shame that it's `RelAssert` and not `ReleaseAssert` for consistency. Are we at all in control of that?

.github/actions/setup_ubuntu_build/action.yml
```diff
@@ -0,0 +1,47 @@
...
+  llvm_version:
+    description: 'version to use llvm cache fetching'
```
frasercrmck, line 9: this sentence could be better worded, I think

.github/workflows/run_pr_tests.yml
```diff
@@ -0,0 +1,84 @@
...
+      - 'hal/**'
+      - '.github/actions/do_build_ock/**'
+      - '.github/actions/setup_ubuntu_build/**'
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+
+  # build and run host x86_64, execute UnitCL and lit tests amd build and run offline
```
frasercrmck, line 21: typo: amd -> and

.github/actions/do_build_ock/do_build_m1/action.yml
```diff
@@ -0,0 +1,22 @@
+name: build-ock-m1
+description: Action to build the oneapi-construction-kit for default riscv m1 target
+
+inputs:
```
frasercrmck, line 4: So does this action hide the inputs of `do_build_ock`, meaning we can't have a `do_build_m1` that also sets `inputs.offline_kernel_tests`?

.github/actions/do_build_ock/action.yml
```diff
@@ -0,0 +1,112 @@
...
+  riscv_enabled:
+    description: "Enable riscv target (default OFF)"
+    default: OFF
+  build_dir:
+    description: "Directory to be used for building"
+    default: build
+  extra_flags:
+    description: "Any additional cmake flags to be passed (default '')"
+    default:
+  assemble_spirv_ll_lit_test_offline:
```
frasercrmck, line 61: In gitlab we have this OFF on all 'online' builds and ON on 'offline' builds. I fear we'll forget about this option if we don't mimic this behaviour now.

.github/workflows/run_pr_tests.yml
```diff
@@ -0,0 +1,84 @@
...
+      # use the previous build for online to get clc
+      - name: build_host_x86_64-offline
+        uses: ./.github/actions/do_build_ock
+        with:
+          build_type: Release
+          extra_flags: -DCA_RUNTIME_COMPILER_ENABLED=OFF -DCA_EXTERNAL_CLC=${{ github.workspace }}/build/bin/clc
+          build_dir: build_offline
+          build_targets: UnitCL
+
+      - name: run host_x86 offline
```
frasercrmck, line 52: The `name`s are inconsistent: we've got `host`, `host_x86` and `host_x86_64` in the one job. Also the inconsistencies between underscores, hyphens and spaces, and capitalization.

.github/workflows/build_pr_cache.yml
```diff
@@ -0,0 +1,66 @@
...
+    paths:
+      - .github/workflows/build_pr_cache.yml
+      - .github/actions/do_build_ock/**
+      - .github/actions/setup_ubuntu_build/**
+  
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  build_ca_host_x86_64:
```
frasercrmck, line 23: This job is called 'host' but also runs riscv?

.github/workflows/run_pr_tests.yml
```diff
@@ -0,0 +1,84 @@
...
+        run:
+          ninja -C build_offline check-UnitCL
+
+      - name: run just online lit
```
frasercrmck, line 56: Maybe run this before the offline steps?

.github/workflows/run_pr_tests.yml
```diff
@@ -0,0 +1,84 @@
...
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      
+      # installs tools, ninja and installs llvm (default 16, RelAssert)
+      - name: setup-ubuntu
+        uses:  ./.github/actions/setup_ubuntu_build
+  
+      # These need to match the configurations of build_mr_cache to use the cache effectively
```
frasercrmck, line 33: Is this meant to say `build_pr_cache`?

.github/actions/setup_ubuntu_build/action.yml
frasercrmck: Also this should maybe say `llvm_build_type` to be clear.

.github/workflows/run_pr_tests.yml
```diff
@@ -0,0 +1,84 @@
...
+  # build and run riscv m1, execute UnitCL and lit tests
+  run_riscv_m1:  
+
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      
+      # installs tools, ninja and installs llvm (default 16, RelAssert)
```
frasercrmck, line 71: Do we need to say `(default 16, RelAssert)` all the time? Would this be better left as being explicitly set (using `with:`?)

Otherwise I fear these comments will easily become stale.

.github/workflows/build_pr_cache.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: clean_build
+        run:
+          rm -rf ${{ github.workspace }}/build*
+
+      - name: build riscv m1
+        uses: ./.github/actions/do_build_ock/do_build_m1  
+
+      - name: run riscv M1 UnitCL for extra checking
+        run:
+          ninja -C build check-UnitCL
```
frasercrmck, line 62: What does running `check-UnitCL` and `check-all-lit` actually do when building a cache? We've already built UnitCL previously, haven't we? Wouldn't this just pointlessly run tests?

frasercrmck (commented):
Do we get any UnitCL/lit artefacts from all of this? It can be quite hard to see the problems just from the console.

.github/actions/setup_ubuntu_build/action.yml
coldav: I had issues with patterns on the caches

ori-sky (changes requested):

.github/actions/do_build_ock/do_build_m1/action.yml
coldav: right now no, but we can add them as needed.

.github/workflows/build_pr_cache.yml
coldav: bad naming, I'll update

.github/workflows/run_pr_tests.yml
coldav: maybe, I was trying to keep the file short

.github/workflows/build_pr_cache.yml
coldav: we are, it was more giving us confidence it was doing the right thing. I was in two minds as to the benefit, so I'll remove

.github/actions/setup_ubuntu_build/action.yml
coldav: I believe we could possible make it use ReleaseAssert if we changed the ordering in the cache name, but I think that should be left for  a later commit as it would require rebuilding the llvm caches.

coldav: I'll look into this at a later point.

.github/actions/do_build_ock/action.yml
coldav: I've set it to on for offline

.github/workflows/run_pr_tests.yml
coldav: agreed

.github/workflows/run_pr_tests.yml
coldav: yes

.github/actions/do_build_ock/action.yml
frasercrmck: Sounds good, but the documented default value below this line is wrong now.

.github/actions/setup_ubuntu_build/action.yml
frasercrmck: Aye fine

.github/workflows/run_pr_tests.yml
```diff
@@ -0,0 +1,89 @@
...
+  # build and run riscv m1, execute UnitCL and lit tests
+  run_riscv_m1:  
+
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      
+      # installs tools, ninja, installs llvm and sets up scccahe
```
frasercrmck, line 73: typo: scccahe

.github/workflows/run_pr_tests.yml
```diff
@@ -0,0 +1,89 @@
...
+
+  # build and run host x86_64, execute UnitCL and lit tests and build and run offline
+  run_host_x86_64:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+      
+      # installs tools, ninja, installs llvm and sets up scccahe
```
frasercrmck, line 29: typo: scccahe -> sccache

frasercrmck (approved):
Just little bits of documentation, otherwise looks good

.github/workflows/run_pr_tests.yml
coldav: yes fair

PR #53, opened 2023-07-03
Retrieve values from barrier struct for use in new memory barriers

AmyCodeplay:
# Overview

The Handle Barriers Pass creates new Mux Memory Barrier calls in between barrier regions. The arguments of this call come from the arguments of the barrier calls inside the original function. When these arguments are constant integers, all is good. However, when the arguments are expressions in the IR of the original function, the result is using values from the original Function inside the wrapper Function, which is not valid IR.

# Reason for change

We should not be generating invalid IR from any of our passes.

# Description of change

When building the new Mux Memory Barrier, we now check for any non-constant arguments and pull them from the first instance of the barrier struct where necessary. The code used by the Barrier Regions to access data in the barrier struct has been made accessible through the public header (`modules/compiler/utils/include/compiler/utils/barrier_regions.h`) so that the Handle Barriers Pass can use it.

### DIFF
--- a/modules/compiler/utils/include/compiler/utils/barrier_regions.h
+++ b/modules/compiler/utils/include/compiler/utils/barrier_regions.h
@@ -174,6 +174,12 @@ class Barrier {
     llvm::Value *getGEP(const llvm::Value *live);
 
     /// @brief get a value reloaded from the barrier struct.
+    ///
+    /// @param[in] live the live value to retrieve from the barrier
+    /// @param[in] insert where to insert new instructions
+    /// @param[in] name a postfix to append to new value names
+    /// @param[in] reuse whether to generate the load for a given value only
+    /// once, returning the previously cached value on further requests.
     llvm::Value *getReload(llvm::Value *live, llvm::Instruction *insert,
                            const char *name, bool reuse = false);
   };
--- a/modules/compiler/utils/include/compiler/utils/barrier_regions.h
+++ b/modules/compiler/utils/include/compiler/utils/barrier_regions.h
@@ -157,6 +157,27 @@ class Barrier {
   llvm::Function &getFunc() { return func_; }
   const llvm::Function &getFunc() const { return func_; }
 
+  /// @brief struct to help retrieval of values from the barrier struct
+  struct LiveValuesHelper {
+    Barrier &barrier;
+    llvm::DenseMap<const llvm::Value *, llvm::Value *> live_GEPs;
+    llvm::DenseMap<const llvm::Value *, llvm::Value *> reloads;
+    llvm::Instruction *insert_point = nullptr;
+    llvm::Value *barrier_struct = nullptr;
+    llvm::Value *vscale = nullptr;
+
+    LiveValuesHelper(Barrier &b, llvm::Instruction *i, llvm::Value *s)
+        : barrier(b), insert_point(i), barrier_struct(s) {}
+
+    /// @brief get a GEP instruction pointing to the given value in the barrier
+    /// struct.
+    llvm::Value *getGEP(const llvm::Value *live);
+
+    /// @brief get a value reloaded from the barrier struct.
+    llvm::Value *getReload(llvm::Value *live, llvm::Instruction *insert,
+                           const char *name, bool reuse = false);
+  };
+
  private:
   /// @brief The first is set for livein and the second is set for liveout
   using live_in_out_t =
--- a/modules/compiler/utils/source/barrier_regions.cpp
+++ b/modules/compiler/utils/source/barrier_regions.cpp
@@ -264,6 +264,120 @@ void UpdateAndTrimPHINodeEdges(BasicBlock *BB, ValueToValueMapTy &vmap) {
 
 }  // namespace
 
+Value *compiler::utils::Barrier::LiveValuesHelper::getGEP(const Value *live) {
+  auto gep_it = live_GEPs.find(live);
+  if (gep_it != live_GEPs.end()) {
+    return gep_it->second;
+  }
+
+  Value *gep;
+  Type *data_ty = live->getType();
+  if (auto *AI = dyn_cast<AllocaInst>(live)) {
+    data_ty = AI->getAllocatedType();
+  }
+
+  if (!isa<ScalableVectorType>(data_ty)) {
+    auto field_it = barrier.live_variable_index_map_.find(live);
+    if (field_it == barrier.live_variable_index_map_.end()) {
+      return nullptr;
+    }
+
+    LLVMContext &context = barrier.module_.getContext();
+    unsigned field_index = field_it->second;
+    Value *live_variable_info_idxs[2] = {
+        ConstantInt::get(Type::getInt32Ty(context), 0),
+        ConstantInt::get(Type::getInt32Ty(context), field_index)};
+
+    gep = GetElementPtrInst::Create(
+        barrier.live_var_mem_ty_, barrier_struct, live_variable_info_idxs,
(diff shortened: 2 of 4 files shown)


modules/compiler/utils/source/handle_barriers_pass.cpp
```diff
@@ -1143,7 +1143,36 @@ Function *compiler::utils::HandleBarriersPass::makeWrapperFunction(
...
         Value *Ops[2] = {CI->getOperand(1), CI->getOperand(2)};
-        auto *Call = B.CreateCall(MemBarrier, Ops);
+
+        auto *const Call = B.CreateCall(MemBarrier, Ops);
+
+        // Patch up any operands that were non-constants by fetching them from
+        // the barrier struct. We do this after creating the call because we
+        // need an instruction to function as an insert point.
+        if (!isa<Constant>(Ops[0]) || !isa<Constant>(Ops[1])) {
+          dbgs() << *Call << "\n";
```
frasercrmck, line 1153: `dbgs` :)

modules/compiler/utils/source/handle_barriers_pass.cpp
```diff
@@ -1143,7 +1143,36 @@ Function *compiler::utils::HandleBarriersPass::makeWrapperFunction(
...
+          auto *const barrier0 = ir.CreateInBoundsGEP(
+              barrierMain.getLiveVarsType(), barrierMain.getMemSpace(), {zero});
+
+          Barrier::LiveValuesHelper live_values(barrierMain, Call, barrier0);
+
+          size_t op_index = 0;
+          for (auto *const op : Ops) {
+            if (!isa<Constant>(op)) {
+              auto *const new_op =
+                  live_values.getReload(op, Call, "_load", true);
```
frasercrmck, line 1169: `/*reuse*/true` might be clearer

modules/compiler/utils/source/barrier_regions.cpp
```diff
@@ -264,6 +264,120 @@ void UpdateAndTrimPHINodeEdges(BasicBlock *BB, ValueToValueMapTy &vmap) {
...
+
+Value *compiler::utils::Barrier::LiveValuesHelper::getReload(
+    Value *live, Instruction *insert, const char *name, bool reuse) {
+  auto &mapped = reloads[live];
+  if (reuse && mapped) {
+    return mapped;
+  }
+
+  Value *v = getGEP(live);
+  if (v) {
```
frasercrmck, line 347: maybe `if (Value *v =` to avoid polluting the scope

modules/compiler/utils/source/barrier_regions.cpp
```diff
@@ -264,6 +264,120 @@ void UpdateAndTrimPHINodeEdges(BasicBlock *BB, ValueToValueMapTy &vmap) {
...
+    if (!isa<AllocaInst>(live)) {
+      // If live variable is not allocainst, insert load.
+      v = new LoadInst(live->getType(), v, Twine(live->getName(), name),
+                       insert);
+    }
+    mapped = v;
+    return v;
+  } else if (auto *I = dyn_cast<Instruction>(live)) {
```
frasercrmck, line 355: can maybe remove `else` after `return` - your call

modules/compiler/utils/include/compiler/utils/barrier_regions.h
```diff
@@ -157,6 +157,27 @@ class Barrier {
...
+
+    LiveValuesHelper(Barrier &b, llvm::Instruction *i, llvm::Value *s)
+        : barrier(b), insert_point(i), barrier_struct(s) {}
+
+    /// @brief get a GEP instruction pointing to the given value in the barrier
+    /// struct.
+    llvm::Value *getGEP(const llvm::Value *live);
+
+    /// @brief get a value reloaded from the barrier struct.
+    llvm::Value *getReload(llvm::Value *live, llvm::Instruction *insert,
```
frasercrmck, line 177: Documenting what `reuse` does could be nice

modules/compiler/utils/source/handle_barriers_pass.cpp
```diff
@@ -1143,7 +1143,36 @@ Function *compiler::utils::HandleBarriersPass::makeWrapperFunction(
...
+        // Patch up any operands that were non-constants by fetching them from
+        // the barrier struct. We do this after creating the call because we
+        // need an instruction to function as an insert point.
+        if (!isa<Constant>(Ops[0]) || !isa<Constant>(Ops[1])) {
+          dbgs() << *Call << "\n";
+          // We expect these values to be uniform so it should be safe to get
+          // from the barrier struct at index zero. Barriers are convergent, so
+          // there should be no chance that the value does not exist.
+          auto *const zero =
+              Constant::getNullValue(Type::getIntNTy(context, 8 * sizeTyBytes));
```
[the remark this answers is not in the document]
AmyCodeplay: hmm? it's a variable local to this function.

modules/compiler/utils/source/handle_barriers_pass.cpp
AmyCodeplay: oops lol

modules/compiler/utils/source/barrier_regions.cpp
AmyCodeplay: yeah doesn't really make much difference imho other than forcing the `if` onto the next line

AmyCodeplay:
> Could do with some lit tests to cover this.

hmm, tricky.. i don't know if there's a way to Lit test the barrier pass in isolation

AmyCodeplay:
ah yes muxc can do it. but i'll have to have a think about it.

modules/compiler/utils/source/handle_barriers_pass.cpp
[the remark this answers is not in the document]
frasercrmck: Oh sorry I meant we could use `llvm::IntegerType *getSizeType(const llvm::Module &m)`, to avoid the internal details of `Type::getIntNTy(context, 8 * sizeTyBytes)`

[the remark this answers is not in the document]
AmyCodeplay: ah yes.. i don't know why we've always done it this way.. it's all through this entire pass though.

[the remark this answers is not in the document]
AmyCodeplay: decided not to do anything about it now

AmyCodeplay:
Lit test added. Also found and fixed a segfault/UB in the Barrier Regions code that came up when i wrote the test a bit wrong rofl

modules/compiler/utils/source/barrier_regions.cpp
```diff
@@ -317,14 +432,16 @@ void compiler::utils::Barrier::FindBarriers() {
...
       if (CallInst *call_inst = dyn_cast<CallInst>(&bi)) {
         Function *callee = call_inst->getCalledFunction();
-        auto const B = bi_->analyzeBuiltin(*callee);
-        if (callee != nullptr && isMuxWGControlBarrierID(B.ID)) {
-          unsigned id = ~0u;
-          auto *const id_param = call_inst->getOperand(0);
-          if (auto *const id_param_c = dyn_cast<ConstantInt>(id_param)) {
-            id = id_param_c->getZExtValue();
+        if (callee != nullptr) {
+          auto const B = bi_->analyzeBuiltin(*callee);
```
frasercrmck, line 436: whoops

frasercrmck (approved):
I'm happy if CI's happy

PR #62, opened 2023-07-20
[compiler] Make some utilities more flexible

frasercrmck:
This provides some groundwork for an upcoming change to the kernel ABI with regards to samplers. I've split this prep work away from that main change, in case we want to land it later, or revert it later on.

Each change is tested with UnitCompiler, so should be safe to land with a pass to make use of them.

### DIFF
--- a/modules/compiler/utils/include/compiler/utils/metadata.h
+++ b/modules/compiler/utils/include/compiler/utils/metadata.h
@@ -269,6 +269,14 @@ struct KernelInfo {
 void populateKernelList(llvm::Module &m,
                         llvm::SmallVectorImpl<KernelInfo> &results);
 
+/// @brief Replaces instances of kernel fromF with toF in module-level
+/// !opencl.kernels metadata.
+/// @param fromF Function to replace with toF in metadata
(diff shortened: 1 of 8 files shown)


modules/compiler/utils/include/compiler/utils/metadata.h
```diff
@@ -269,6 +269,11 @@ struct KernelInfo {
 void populateKernelList(llvm::Module &m,
                         llvm::SmallVectorImpl<KernelInfo> &results);
 
+/// @brief Replaces instances of kernel fromF with toF in module-level
+/// !opencl.kernels metadata.
```
coldav, line 273: should probably have @param too, although I'm not sure how much that gives you here.

coldav (approved):

modules/compiler/utils/include/compiler/utils/metadata.h
frasercrmck: aye right you are. doesn't hurt though and it's probably not for me to decide whether something's too obvious to document?

PR #63, opened 2023-07-24
[compiler][host] Pass samplers to kernels as i32 parameters

frasercrmck:
This changes the kernel ABI, which in practice only affects the host target since riscv doesn't support images.

The host runtime now passes samplers as their native i32 values rather than as integers pretending to be pointers, which are reinterpreted by the kernel.

The compiler code here is inflexible and target agnostic, but is intended as a short-term solution; it won't be the code path taken by LLVM 17, which will come with its own pass for replacing target extension types with target-defined types. In that sense, in the near future we will pass off the responsibility for deciding how samplers, images and events are passed to the kernel to each target.

The compiler code is simply to replace kernels with wrappers that take i32 parameters, reinterpreting them as pointers and calling the original kernel. These pointers will then be reinterpreted back to i32s when calling the libimg functions, so should be eliminated during the usual course of optimizations (i.e., inlining).


modules/compiler/source/base/source/image_argument_substitution_pass.cpp
```diff
@@ -52,6 +55,109 @@ const std::map<std::string, std::string> funcToFuncMap = {{
...
+                    }
+                  }
+                  return P.value();
+                });
+      // If we've not found any samplers, skip to the next kernel
+      if (equal(KernelF->getFunctionType()->params(), KernelArgTys)) {
+        continue;
+      }
+
+      // As this point in the pipeline, we're not in a position to rename the
```
coldav, line 120: As -> At

coldav (approved):

modules/compiler/source/base/source/image_argument_substitution_pass.cpp
frasercrmck: right you are, thanks

PR #68, opened 2023-07-26
Find clang_tidy at version 9

coldav:
# Overview

Split clang-format and clang-tidy tools, so we can still use clang-tidy-9.

# Reason for change

There were some failures from using the latest clang-tidy which need resolved. 

# Description of change

Split the tool access for clang-format and clang-tidy so we can use separate versions and set clang-tidy to 9 which is known to work.


CMakeLists.txt
```diff
@@ -180,6 +180,7 @@ find_package(PythonInterp 3.6 REQUIRED)
 # When updating the version here, also update that used in the merge request
 # config
 find_package(ClangTools 16 COMPONENTS clang-format clang-tidy)
```
frasercrmck, line 182: is it intentional that we're still trying to find clang-tidy at version 16? Will it print a message that it wasn't found?

coldav: oops I meant to take that bit out

PR #72, opened 2023-07-28
Add basic pass through for OpCapability MemoryAccessAliasingINTEL 

MaryaSharf:
# Overview

To get sycl-nbody runwith oneapi-construction-kit and dpc++, `OpCapability MemoryAccessAliasingINTEL` support is needed.

# Reason for change

While trying to run sycl-nbody with oneapi-cnstruction-kit, we get the error:
`error: OpCapability MemoryAccessAliasingINTEL (#5910) not supported by device`.

# Description of change

As for now just the basic pass through is added. A proper fix may be added later.

# Checklist

### DIFF
--- a/modules/compiler/spirv-ll/test/spvasm/intel_memory_access_aliasing.spvasm
+++ b/modules/compiler/spirv-ll/test/spvasm/intel_memory_access_aliasing.spvasm
@@ -1,0 +1,36 @@
+; Copyright (C) Codeplay Software Limited
+;
+; Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+; Exceptions; you may not use this file except in compliance with the License.
+; You may obtain a copy of the License at
+;
+;     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+;
+; Unless required by applicable law or agreed to in writing, software
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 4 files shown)


frasercrmck:
Even though it's a pass-through support, I think a simple LIT test would be prudent (to double-check we don't crash)

modules/compiler/spirv-ll/test/spvasm/intel_memory_access_aliasing.spvasm
```diff
@@ -0,0 +1,39 @@
...
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; REQUIRES: spirv-as-v2022+
+; RUN: %if online-spirv-as %{ spirv-as --target-env %spv_tgt_env -o %spv_file_s %s %}
+; RUN: spirv-ll-tool -a OpenCL -b 64 -e SPV_INTEL_memory_access_aliasing %spv_file_s | FileCheck %s
```
frasercrmck, line 19: This is just testing that we don't crash, so I think we can get rid of `| FileCheck %s` and remove all the `CHECK`s from the tests.

coldav (approved):

PR #73, opened 2023-08-01
Drop support for SPIR 1.2 IR programs

frasercrmck:
This IR serialization is outdated and hard to maintain as it requires an old forked version of clang to generate, it can't produce OpenCL 3.0 programs, and it's increasingly incompatible with modern LLVM IR (opaque pointers, target extension types, etc).


modules/compiler/source/base/source/image_argument_substitution_pass.inc
```diff
@@ -120,10 +120,8 @@
 // intanium mangling mandates that built-ins are not substituted.
 // However,
 // clang 3.6 got it wrong and does do the substitution. By adding the
-// extra mapping we make sure that if clang ever gets fixed (or we are
-// consuming a binary generated by SPIR 1.2 generator) we can correctly
-// match the function. There is a funny twist to that issue, CTS SPIR
-// test
+// extra mapping we make sure that if clang ever gets fixed we can correctly
+ // match the function. There is a funny twist to that issue, CTS SPIR test
```
coldav, line 124: weird indentation - also should we be talking about SPIR tests if we are skipping them?

modules/compiler/source/base/source/printf_replacement_pass.cpp
```diff
@@ -335,8 +335,8 @@ std::optional<std::string> getPointerToStringAsString(Value *op) {
       if (const_string->getOpcode() == Instruction::GetElementPtr) {
         var = dyn_cast<GlobalVariable>(const_string->getOperand(0));
       } else if (const_string->getOpcode() == Instruction::IntToPtr) {
-        // SPIR generates a PtrToInt expression inside an IntToPtr expression,
-        // so therefore we need to unwrap it twice.
+        // Sometimes we see a PtrToInt expression inside an IntToPtr
```
coldav, line 338: This said SPIR before - is this still an issue?

coldav (approved):

modules/compiler/source/base/source/printf_replacement_pass.cpp
frasercrmck: If it's not incorrect IR I was thinking we should keep the code to handle it. I think the old comment was overly specific to where we've seen the IR coming from.

PR #80, opened 2023-08-03
[compiler] Introduce mux builtins for 'group operations'

frasercrmck:
This introduces a set of builtins to represent operations like OpenCL's
`sub_group_xxx` and `work_group_xxx` and SPIR-V `OpGroupXXX` operations
in a language-agnostic way.

It also introduces 'vector group' operations which represent sub-groups
as the vectorizer currently thinks of them: as groups of work-items
being 'simulated' together via vectorized code, despite being executed
on only one invocation. These have no direct OpenCL equivalent, but can
be used by the compiler to simulate larger sub-groups than the hardware
actually provides.

Until now, the construction kit's handling of these operations has been
centered around analysis of the OpenCL builtins, which is inflexible and
brittle. We may one day want to support IR in a format not compatible
with OpenCL-like IR, for instance. It also means that the compiler is
either constrained by the OpenCL semantics of these operations.

These builtins are not yet generated in the default pipeline, but
there's a new pass which introduces them to the module by asking the
language-level `BuiltinInfo` to replace calls to its concept of group
builtins with mux's concept of these builtins. The builtins should also
be correctly wired up through the identification and get-or-declare
APIs, but are not yet able to be *defined* by mux. This will come in a
later change.

### DIFF
--- a/doc/specifications/mux-compiler-spec.rst
+++ b/doc/specifications/mux-compiler-spec.rst
@@ -1,7 +1,7 @@
 ComputeMux Compiler Specification
 =================================
 
-   This is version 0.77.0 of the specification.
+   This is version 0.78.0 of the specification.
 
 ComputeMux is Codeplay’s proprietary API for executing compute workloads across
 heterogeneous devices. ComputeMux is an extremely lightweight,
@@ -922,6 +922,10 @@ A Mux implementation **shall** provide definitions for these builtin functions.
   the current work-group.
 * ``i32 __mux_get_max_sub_group_size()`` - Returns the maximum subgroup size
   in the current kernel.
+* ``i32 __mux_get_sub_group_size()`` - Returns the number of invocations in the
+  subgroup.
+* ``i32 __mux_get_sub_group_local_id()`` - Returns the unique invocation ID
+  within the current sub-group.
 * ``size_t __mux_get_group_id(i32 %i)`` - Returns the unique work-group
   identifier for the ``%i``'th dimension.
 * ``i32 __mux_get_work_dim()`` - Returns the number of dimensions in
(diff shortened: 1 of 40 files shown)


doc/specifications/mux-compiler-spec.rst
```diff
@@ -997,6 +997,114 @@ A Mux implementation **shall** provide definitions for these builtin functions.
...
+  range. These builtins **must** only take scalar values.
+* ``sub-group`` - a subset of invocations in a work-group which can synchronize
+  and share data efficiently. ComputeMux leaves the choice of sub-group size
+  and implementation to the target; ComputeMux only defines these builtins with
+  a "trivial" sub-group size of 1. These builtins **must** only take scalar
+  values.
+* ``vec-group`` - a software level group of invocations processing data in
+  parallel *on a single invocation*. This allows the compiler to simulate a
+  sub-group without any hardware sub-group support (e.g., through
+  vectorization). These buitins **may** take scalar *or vector* values. The
```
coldav, line 1036: buitins -> builtins

modules/compiler/utils/include/compiler/utils/builtin_info.h
```diff
@@ -518,21 +587,55 @@ class BuiltinInfo {
...
   /// @brief Defines the body of a ComputeMux builtin declaration
   ///
   /// If the Module already has a function definition with the corresponding
   /// function name, it is left alone and returned.
   ///
   /// Will declare any builtins it requires as transitive dependencies.
-  llvm::Function *defineMuxBuiltin(BuiltinID, llvm::Module &M);
+  llvm::Function *defineMuxBuiltin(
+      BuiltinID, llvm::Module &M,
+      llvm::ArrayRef<llvm::Type *> OverloadInfo = {});
```
coldav, line 633: I think it would be useful to have an @param for the overloadInfo to give an idea of what it's for

modules/compiler/utils/source/cl_builtin_info.cpp
```diff
@@ -2982,6 +2989,451 @@ CallInst *CLBuiltinInfo::mapSyncBuiltinToMuxSyncBuiltin(
...
+    while (!IsSignedInt && Qual.getCount()) {
+      IsSignedInt |= Qual.pop_front() == compiler::utils::eTypeQualSignedInt;
+    }
+
+    bool IsFP = ArgumentTypes[0]->isFloatingPointTy();
+    switch (MuxBuiltinID) {
+      default:
+        llvm_unreachable("unknown group operation for which to check the type");
+      case eMuxBuiltinSubgroupReduceAdd:
+        MuxBuiltinID = IsFP ? eMuxBuiltinSubgroupReduceFAdd : MuxBuiltinID;
```
coldav, line 3262: could you explain why it's an ID like eMuxBuiltinSubgroupReduceFAdd  or something generic like muxBuiltinID?

modules/compiler/test/group_ops.cpp
```diff
@@ -0,0 +1,402 @@
...
+          "i32", Collective));
+
+      Collective.Op = GroupCollective::OpKind::All;
+      Collective.Recurrence = RecurKind::And;
+      GroupOpNames.emplace_back(GroupOp(
+          Mangler.mangleName(BaseName + "all", I32Ty, {eTypeQualSignedInt}),
+          "i32", Collective));
+    }
+
+    if (IncludeBroadcasts) {
```
coldav, line 116: might be worth breaking this function up as it's quite long

modules/compiler/utils/include/compiler/utils/builtin_info.h
coldav: In fact, I don't see it really explained anywhere include the documentation (which looking at doesn't go into too much direct detail about builtinInfo - although talks about it quite a bit indirectly).

coldav (approved):

modules/compiler/utils/include/compiler/utils/builtin_info.h
frasercrmck: I'd added a `@param` now, and have also added a static function `BuiltinInfo::isOverloadableMuxBuiltinID` to tie those concepts together

You're right that `BuiltinInfo` isn't very well documented in the docs. That's something we should address

modules/compiler/utils/source/cl_builtin_info.cpp
frasercrmck: It's because `MuxBuiltinID` has already been set to a 'base' ID in a previous step, which in this case is `eMuxBuiltinSubgroupReduceAdd`. This step is just refining that to `FAdd` if it's `IsFP`.

frasercrmck: It's basically to try and do less complicated logic in a previous step, because we don't want to start demangling builtins and checking types unless we have to, which is only in comparitively few situations, like add/mul/min/max operations.

frasercrmck: Oops, just noticed we were doing a redundant demangling in all cases anyway. I've removed that now

coldav: I think just a comment to explain why we set it to the generic version as it's not obvious.

modules/compiler/test/group_ops.cpp
frasercrmck: done

modules/compiler/utils/source/cl_builtin_info.cpp
frasercrmck: done

PR #84, opened 2023-08-09
[compiler] Enable mux subgroup/workgroup builtins

frasercrmck:
This enables the use of the mux subgroup and workgroup 'collective'
builtins in the pipeline, replacing the OpenCL ones early on.

The passes which used to care about the OpenCL builtins have been
updated to care about the mux builtins. There is no real change in
behaviour, except in a few cases where OpenCL any/all worked on i32
types but the mux versions work on i1 types: vecz was already
translating between these two types, so there's less to do now.

This should simplify a lot of the code.

* The mux sub-group builtins are defined by DefineMuxBuiltinsPass; there
  is no 'trivial' implementation in the OpenCL headers.
* As such, the LinkBuiltinsPass no longer needs to know whether it's
  'early' or not, as the sub-group builtins are declarations and are
  thus safe to link at any point.
* We can get away with doing a lot less function name demangling,
  mangling and manipulation, as we have exposed enums for the builtin
  IDs.
  * The DegenerateSubGroupPass still does name shenanigans, but we can
    fix that up separately.


doc/modules/compiler/utils.rst
```diff
@@ -1173,29 +1172,25 @@ such APIs, several of which are given here by way of example:
...
-``modules/compiler/utils/source/define_mux_builtins_pass.cpp``.
+A implementation of OpenCL C sub-group builtins is provided by the default
+compiler pipeline.
+
+The OpenCL C sub-group builtins are first translated into the corresponding
+ComputeMux builtin functions. These functions are understood by the rest of the
+compiler and can be identified and analyzed by the ``BuiltinInfo`` analysis.
+
+A definition of these mux builtins for where the sub-group size is 1 is
+provided of ``BIMuxInfoConcept`` used by the `DefineMuxBuiltinsPass`_.
```
coldav, line 1183: provided by

modules/compiler/utils/source/degenerate_sub_group_pass.cpp
```diff
@@ -50,59 +50,53 @@ namespace {
...
-    for (unsigned i = 0; i < 2; ++i) {
-      ArgumentTypes.push_back(sizeTy);
-      Qualifiers.push_back(compiler::utils::eTypeQualNone);
+  std::string WorkgroupFName = "__mux_work" + SubgroupFName.substr(9).str();
+
+  // Work-group builtins have an extra 'barrier id' as the first parameter.
+  SmallVector<Type *, 4> ArgTys;
+  ArgTys.push_back(IntegerType::getInt32Ty(M.getContext()));
+  ArgTys.push_back(SubgroupF.getArg(0)->getType());
+  // Broadcasts have different IDs
```
coldav, line 89: as an additional ID?

doc/modules/compiler/utils.rst
frasercrmck: ta

modules/compiler/utils/source/degenerate_sub_group_pass.cpp
frasercrmck: I've clarified this comment, thanks. It was a bit confusing with respect to the 'barrier IDs' mentioned above.

coldav (approved):

PR #86, opened 2023-08-14
[compiler] Further integrate mux group builtins

frasercrmck:
This PR just better integrates the new mux group builtins by:

* providing helper methods for common queries
* tidying up some code
* refactoring the degenerate sub-groups pass to use builtin IDs and not function name manipulation.

### DIFF
--- a/modules/compiler/utils/source/cl_builtin_info.cpp
+++ b/modules/compiler/utils/source/cl_builtin_info.cpp
@@ -1006,7 +1006,6 @@ Builtin CLBuiltinInfo::analyzeBuiltin(Function const &Callee) const {
 
   bool IsConvergent = false;
   unsigned Properties = eBuiltinPropertyNone;
-  llvm::SmallVector<llvm::Type *, 2> OverloadInfo;
   switch (ID) {
     default:
       // Assume convergence on unknown builtins.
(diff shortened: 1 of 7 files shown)


modules/compiler/utils/source/cl_builtin_info.cpp
```diff
@@ -1278,10 +1278,7 @@ Builtin CLBuiltinInfo::analyzeBuiltin(Function const &Callee) const {
     case eCLBuiltinWorkgroupScanLogicalXorExclusive:
       IsConvergent = true;
       Properties |= eBuiltinPropertyMapToMuxGroupBuiltin;
-      if (ID != eCLBuiltinWorkgroupAll && ID != eCLBuiltinWorkgroupAny &&
```
coldav, line 1281: why are we able to remove this check?

frasercrmck: Well my thinking was that __mux any/all _are_ overloaded so expect this to be set. But when I went back to look to address this comment, it turns out this function is just pushing stuff into a vector and never returning it, so it's all dead code. It might come from an earlier development version of the original PR.

The overload info that's important is being constructed afresh in `CLBuiltinInfo::mapGroupBuiltinToMuxGroupBuiltin`.

So I reckon we can just get rid of `OverloadInfo` completely from this function.

coldav (approved):

PR #87, opened 2023-08-14
[compiler] Start to combine 'ReplaceXXX' passes

frasercrmck:
We have many passes which replace either generic language builtins
(`ReplaceGroupFuncsPass` & `ReplaceBarriersPass`) or hard-coded OpenCL
builtins (`ReplaceAsyncCopiesPass`, `ReplaceAtomicFuncsPass`,
`ReplaceC11AtomicFunsPass`) with some kind of mux/llvm equivalents.

This is not a scalable approach and is hard to communicate to
users, and so this PR starts to merge these passes into one more
versatile pass: `LowerToMuxBuiltinsPass`. Its job is to ask the language
BuiltinInfo which builtins are to be replaced with mux equivalents, so that
the following compiler passes can correctly understand the program semantics.

This marks the beginning of us moving more things more consistently into
the 'mux' sphere and dropping the support for language builtins at all
stages of compilation.

Now, _all_ work-item builtins are mapped to mux builtins early on in the
`LowerToMuxBuiltinsPass`.

To support this, mux builtins can now report their range metadata (identical
to the current CLBuiltinInfo implementation)


doc/modules/compiler/utils.rst
```diff
@@ -45,6 +45,30 @@ Their job is three-fold:
...
+LowerToMuxBuiltinsPass
+----------------------
+
+Replaces calls to the language implementation's builtin functions with
+alternative sequences of instructions involving ComputeMux builtins.
+
+Targets **must** lower their language builtins to ComputeMux builtins in order
+for other ComputeMux compiler passes to corectly recognise the program
+semantics. This is because the compiler passes provided by the oneAPI
+construction kit only understand the semantics ComputeMux builtins, and thus
```
coldav, line 57: sentence does not scan right

modules/compiler/test/lit/passes/lower-to-mux-builtins.ll
```diff
@@ -0,0 +1,83 @@
...
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; RUN: muxc --passes lower-to-mux-builtins,verify -S %s | FileCheck %s
+
+; Check that the replace-wgc correctly defines the work-group collective functions
```
coldav, line 19: If this is calling lower-to-mux-builtins, why does it mention replace-wgc here?

doc/modules/compiler/utils.rst
frasercrmck: aye, you're right. I've tried to sort it out by rewriting this paragraph.

modules/compiler/test/lit/passes/lower-to-mux-builtins.ll
frasercrmck: copy/paste :)

modules/compiler/utils/source/cl_builtin_info.cpp
```diff
@@ -2820,13 +2840,66 @@ static multi_llvm::Optional<unsigned> parseMemoryOrderParam(Value *const P) {
...
-Instruction *CLBuiltinInfo::mapSyncBuiltinToMuxSyncBuiltin(
+Instruction *CLBuiltinInfo::lowerBuiltinToMuxBuiltin(
     CallInst &CI, BIMuxInfoConcept &BIMuxImpl) {
   auto &M = *CI.getModule();
   auto *const F = CI.getCalledFunction();
   assert(F && "No calling function?");
   auto const ID = identifyBuiltin(*F);
 
+  // Handle straightforward 1:1 mappings.
+  if (auto MuxID = [](BuiltinID ID) -> std::optional<BuiltinID> {
```
coldav, line 2851: I think I'd find this more readable outside the if (). Also may be worth factoring it out as this function is pretty big.

modules/compiler/utils/source/mux_builtin_info.cpp
```diff
@@ -1127,5 +1127,49 @@ Value *BIMuxInfoConcept::initializeSchedulingParamForWrappedKernel(
...
+    case eMuxBuiltinGetGlobalSize: {
+      auto *DimIdx = CI.getOperand(0);
+      if (!isa<ConstantInt>(DimIdx)) {
+        return std::nullopt;
+      }
+      uint64_t DimVal = cast<ConstantInt>(DimIdx)->getZExtValue();
+      if (DimVal >= SizesPtr->size() || !(*SizesPtr)[DimVal]) {
+        return std::nullopt;
+      }
+      // ID builtins range from [0,size) and size builtins from [1,size]. Thus
```
coldav, line 1161: I know these are mathematical descriptions, but not everyone remembers their meaning so I think I'd prefer it to be described e.g. 0 to size-1 and 1 to size inclusive

coldav (approved):

modules/compiler/utils/source/cl_builtin_info.cpp
frasercrmck: Yeah I agree, I've split that into its own function.

modules/compiler/utils/source/mux_builtin_info.cpp
frasercrmck: I think I'll keep the concise notation but add `(inclusive)` and `(exclusive)` after each to be extra clear.

PR #98, opened 2023-08-22
Amy/refactor work group collectives 2

AmyCodeplay:
# Overview

Work group collective reductions and broadcasts to be handled directly in the Handle Barriers Pass, with support in Vecz for efficient implementation, removes the need for these builtins to be processed by the Replace WGC Pass.

# Reason for change

The implementations of the Work Group Collective functions in the Replace WGC Pass creates unsightly global variables as accumulators, which forces us to introduce more barriers than is really necessary. Additionally, the work group collectives are implemented in terms of subgroup builtins, which in the case of degenerate subgroups, no longer actually mean what they should (the work group is the subgroup in this case, but the subgroup builtins are given the temporary meaning of "vecz group"). So we would like to move towards phasing out the Replace WGC Pass altogether.

# Description of change

Vecz will now internally handle a work group collective reduction and broadcast, producing a work group reduction or broadcast over the vectorized reduction or broadcast. The Handle Barriers Pass processes a work group reduction or broadcast like a barrier, but inserts extra code to perform the operation. In the case of a reduction, this is a linear reduction loop over all items in the barrier struct. The broadcast is implemented as a simple fetch from the barrier struct at the given item ID.

The Replace WGC Pass can now run in a "scans only" mode whereby only work group collective scan operations will be replaced. This is necessary because scans are not yet implemented in the Handle Barriers Pass. Any target that does not run the Handle Barriers Pass is still reliant on the functioning implementations provided by the Replace WGC Pass.


modules/compiler/test/lit/passes/replace-wgc-khr-spv-group-uniform-arithmetic.ll
```diff
@@ -21,46 +21,15 @@
...
 
-; CHECK-DAG: @[[MULI_ACCUM:.+]] = internal addrspace(3) global i32 undef
-; CHECK-DAG: @[[MULF_ACCUM:.+]] = internal addrspace(3) global float undef
-; CHECK-DAG: @[[MULH_ACCUM:.+]] = internal addrspace(3) global half undef
 ; CHECK-DAG: @__mux_work_group_scan_inclusive_and_i32.accumulator = internal addrspace(3) global i32 undef
 ; CHECK-DAG: @__mux_work_group_scan_exclusive_and_i32.accumulator = internal addrspace(3) global i32 undef
 ; CHECK-DAG: @__mux_work_group_scan_inclusive_logical_and_i1.accumulator = internal addrspace(3) global i1 undef
 
-; CHECK: define spir_func i32 @__mux_work_group_reduce_mul_i32(i32 %id, i32 [[PARAM:%.*]])
-declare spir_func i32 @__mux_work_group_reduce_mul_i32(i32 %id, i32 %x)
```
frasercrmck, line 32: Why don't we run the `replace-wgc` pass in `!scans_only` mode? Then we can still test that it does the right thing for reductions and broadcasts.

modules/compiler/utils/include/compiler/utils/replace_wgc_pass.h
```diff
@@ -39,7 +39,13 @@ namespace utils {
 /// target making use of that pass.
 class ReplaceWGCPass final : public llvm::PassInfoMixin<ReplaceWGCPass> {
  public:
+  ReplaceWGCPass(bool scansOnly = false) : scans_only(scansOnly) {}
```
frasercrmck, line 42: This property should be settable on the command line(e.g. `--passes=replace-wgc<scans-only>`) so we can test it in both modes.

modules/compiler/utils/include/compiler/utils/replace_wgc_pass.h
```diff
@@ -39,7 +39,13 @@ namespace utils {
...
   llvm::PreservedAnalyses run(llvm::Module &, llvm::ModuleAnalysisManager &);
+
+ private:
+  /// @brief when set to true, only work group collective scans will be
+  /// replaced.
+  bool scans_only = false;
```
frasercrmck, line 48: What about a property in `BIMuxInfoImpl` that requests full software implementations of the WGC functions, which RefSi G1 WIPT sets and the `ReplaceWGCPass` reads? That way we don't need to anticipate the barriers pass running.

Or a bitfield of which ones the target would like to be provided for it: `unsigned please_give_me = reductions | broadcasts | scans`

That way it's not so much as pass pipeline thing as it is a target configuration.

modules/compiler/utils/source/replace_wgc_pass.cpp
```diff
@@ -615,6 +615,11 @@ PreservedAnalyses compiler::utils::ReplaceWGCPass::run(
     auto Builtin = BI.analyzeBuiltin(F);
     if (auto WGC = BI.isMuxGroupCollective(Builtin.ID);
         WGC && WGC->isWorkGroupScope()) {
+      if (scans_only &&
```
frasercrmck, line 618: `scans_only && !WGC->isScan()`  would be simpler

modules/compiler/riscv/source/riscv_pass_machinery.cpp
```diff
@@ -223,6 +223,11 @@ llvm::ModulePassManager RiscvPassMachinery::getLateTargetPasses() {
   // functions will be cloned to give a version using degenerate sub-groups and
   // a version using non-degenerate sub-groups, for selection by the runtime.
   tuner.degenerate_sub_groups = true;
+
+  // We run the handle barriers pass, which deals with work group reductions and
+  // broadcasts itself.
+  tuner.handling_barriers = true;
```
frasercrmck, line 229: I don't see how this all fits together. If `RefSi G1 WIPT` is relying on `handling_barriers = false` as the default value, does that mean that `RefSi M1` is doing that too? Because that would be incorrect.

modules/compiler/utils/source/handle_barriers_pass.cpp
```diff
@@ -347,9 +375,262 @@ struct ScheduleGenerator {
...
+
+    compiler::utils::CreateLoopOpts inner_opts;
+    inner_opts.attemptUnroll = true;
+    inner_opts.disableVectorize = true;
+
+    BasicBlock *preheader = block;
+    BasicBlock *exitBlock = nullptr;
+    PHINode *resultPhi = nullptr;
+
+    auto *const zero = Constant::getNullValue(Type::getIntNTy(
```
frasercrmck, line 399: this is

``` cpp
 auto *const zero = Constant::getNullValue(compiler::utils::getSizeType(module));
```

frasercrmck (commented):
As best as I can tell, only `barriers-cfg-reduce.ll` is actually lit testing the barriers pass behaviour, and only for reductions.

modules/compiler/source/base/include/base/pass_pipelines.h
```diff
@@ -53,6 +53,9 @@ struct BasePassPipelineTuner {
   /// @brief Whether or not to generate code for degenerate sub groups.
   bool degenerate_sub_groups = false;
 
+  /// @brief Whether or not the Handle Barriers Pass is going to be run.
+  bool handling_barriers = false;
```
coldav, line 57: should this really be the default - most targets want this to be true

modules/compiler/utils/source/barrier_regions.cpp
```diff
@@ -45,8 +46,19 @@ using namespace llvm;
 namespace {
 using AlignIntTy = uint64_t;
 
-bool isMuxWGControlBarrierID(compiler::utils::BuiltinID ID) {
-  return ID == compiler::utils::eMuxBuiltinWorkGroupBarrier;
+/// @brief it retuns the instruction as a call instruction if and only if it is
```
coldav, line 49: retuns -> returns

modules/compiler/vecz/source/transform/packetizer.cpp
```diff
@@ -1147,9 +1150,16 @@ Value *Packetizer::Impl::packetizeSubgroupReduction(Instruction *I) {
   Value *v =
       createSimpleTargetReduction(B, &TTI, opPackets.front(), Info->Recurrence);
 
-  IC.deleteInstructionLater(CI);
-
-  CI->replaceAllUsesWith(v);
+  if (isWorkGroup) {
+    // For a work group operation, we leave the origial reduction function and
```
coldav, line 1154: origial -> original

modules/compiler/riscv/source/riscv_pass_machinery.cpp
coldav, line 229: Since this is only used in once place and the flag is `scansOnly` - should this flag be that - and a better description added? The default should be whatever works for the old main wg targets (host/riscv G1 WG, M1), so it's not a major change for any customers who would need to set it.

modules/compiler/source/base/include/base/pass_pipelines.h
coldav, line 57: Make the default true so any customer targets inherit this since it fits most of our targets and in the past most our customer targets.

modules/compiler/utils/include/compiler/utils/replace_wgc_pass.h
AmyCodeplay: cool, i was wondering if we could do that

modules/compiler/utils/include/compiler/utils/replace_wgc_pass.h
AmyCodeplay: i just wanted to do something quickly that worked because (a) i'm about to go away and (b) this whole thing is supposed to be got rid of before too long anyway. it's only hanging around at all because i didn't have time to get Scans working the same way.

modules/compiler/utils/source/replace_wgc_pass.cpp
AmyCodeplay: good call

modules/compiler/riscv/source/riscv_pass_machinery.cpp
AmyCodeplay: it would still be correct. replacing the reductions and broadcasts should still work in any case. but i don't know how those things worked so i made the changes in the targets i understood.

AmyCodeplay: i wanted the flag in the tuner to be as abstract as possible, the "scans only" meaning in the actual implementation is a temporary measure. Well, probably "handling barriers" should be too although perhaps there are other ways the vectorizer could use this information in future, who knows

modules/compiler/source/base/include/base/pass_pipelines.h
AmyCodeplay: if i know how to make only the G1 target set it to false then i could do that.

modules/compiler/utils/include/compiler/utils/replace_wgc_pass.h
coldav: you have to add support for it in the pass machinery / .def file

modules/compiler/riscv/source/riscv_pass_machinery.cpp
coldav: well it's open source so I'd rather it still reflected what it did with a big old comment.

modules/compiler/source/base/include/base/pass_pipelines.h
coldav: under examples/refsi/refsi_g1_wi/compiler/refsi_g1_wi/source/refsi_pass_machinery.cpp -  you can set the flag directly there - although check with @frasercrmck as it's all changed since I last looked.

coldav:
You might also have to update under modules/compiler/cookie which has pipelines too.

modules/compiler/riscv/source/riscv_pass_machinery.cpp
AmyCodeplay: if true is default we don't need this line anymore anyway

AmyCodeplay: well `true` is default now so i can resolve this..

modules/compiler/utils/source/handle_barriers_pass.cpp
```diff
@@ -347,12 +375,264 @@ struct ScheduleGenerator {
...
+    auto *const accTy = accumulator->getType();
+    Function *const func = block->getParent();
+
+    // Subgroup induction variables
+    Value *subgroupIVs[] = {accumulator};
+    auto *const totalSize = barrier.getTotalSize();
+
+    compiler::utils::CreateLoopOpts inner_opts;
+    inner_opts.attemptUnroll = true;
+    inner_opts.disableVectorize = true;
```
coldav, line 393: there seens to be a bit of a mixture of naming styles here (underscores, then camelcase) - I note that above we use undescores

coldav (approved):

coldav (approved):

modules/compiler/utils/source/handle_barriers_pass.cpp
AmyCodeplay: yeah that's true all over the place, there's no consistency at all, and then i copy and paste a bit from somewhere else, maybe some day we should decide on something and then go through and enforce it.

modules/compiler/utils/include/compiler/utils/replace_wgc_pass.h
AmyCodeplay: if you want to do this then by all means, it could be useful for other things who knows, but getting the scans into the barrier pass would be more of a priority and then we won't need to parameterize the replace-wgc pass at all, we would simply not run it when using the barrier pass, in fact it should by that point be possible to run it after vecz rather than before like now so in effect it just becomes a replacement for/alternative to the barrier pass.

Resolving comment now.

PR #99, opened 2023-08-22
[compiler] Remove assume-like intrinsics used by allocas

frasercrmck:
In the barriers pass, when a live var is an alloca instruction, we replace it and all of its uses directly with a GEP into the live variable struct.

This is unsafe to do in the presence of "assume-like" intrinsics, as the new memory region is not the same as the old one. Some interesting examples include `llvm.assume`, whose assumption may no longer hold, `llvm.lifetime.*` intrinsics, as the lifetime of the GEP is not equal to that of the alloca, and `llvm.invariant.*` intrinsics.

This doesn't seem to have shown up as a problem before LLVM 17, where a `llvm.lifetime.start` intrinsic using an alloca with the *base address* of the live variables struct. This lifetime intrinsic was being called after a store to *another* member of the live variables struct that we inserted during the pass. LLVM was then later able to replace the base GEP with the actual pointer (including in the lifetime intrinsic), and thus infer that the store was happening before the lifetime of its base address, and so removed the store.

### DIFF
--- a/modules/compiler/utils/source/barrier_regions.cpp
+++ b/modules/compiler/utils/source/barrier_regions.cpp
@@ -1095,7 +1095,7 @@ Function *compiler::utils::Barrier::GenerateNewKernel(BarrierRegion &region) {
     }
   }
 
-  SmallVector<AllocaInst *, 8> allocas_to_remove;
+  SmallVector<Instruction *, 8> allocas_and_intrinsics_to_remove;
 
   // Store only live variables that are defined in this kernel.
   //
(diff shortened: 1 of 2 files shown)


modules/compiler/utils/source/barrier_regions.cpp
```diff
@@ -1120,8 +1120,22 @@ Function *compiler::utils::Barrier::GenerateNewKernel(BarrierRegion &region) {
...
+        // Also remove any assume-like intrinsics that are users of this
+        // alloca. These assumptions may not hold. For example, lifetime
+        // intrinsics are definitely dangerous, as by directly replacing their
+        // alloca operands with the address of the live variable struct, we are
+        // telling LLVM that *all* accesses of the live variable struct also
+        // start/end at that point, which is not true.
+        // Similarly, llvm.assume and llvm.experimental.noalias.scope.decl may
+        // hold for the alloca but not the live variables struct.
+        for (auto *const user : alloca_inst->users()) {
+          if (auto *const lifetime = dyn_cast<IntrinsicInst>(user);
```
coldav, line 1134: presumably this could catch no-lifetime intrinsics? If so, the name is misleading

frasercrmck: whoops yeah, I originally started out handling only lifetime intrinsics but saw that we were buggy in other ways. I've updated to simply `intrinsic`

coldav (approved):

PR #110, opened 2023-08-29
Account for loop unrolling in compiler::utils::createLoop()

AmyCodeplay:
# Overview

The Handle Barriers Pass needs to account for the case that `compiler::utils::createLoop()` can do its own loop unrolling.

# Reason for change

UnitCL tests were failing on Arm because of a vectorization factor that enabled the vector reduction loops to be unrolled, resulting in bad code.

# Description of change

The IR generation for reduction loops no longer assumes that `createLoop()` has to return an empty exit block. When it unrolls, it will populate the block with the actual iterations, and then return that. In this case, we don't need to create an LCSSA Phi node and can return the accumulator directly.

A lit test is also included.

### DIFF
--- a/modules/compiler/test/lit/passes/subgroup-loop-unroll.ll
+++ b/modules/compiler/test/lit/passes/subgroup-loop-unroll.ll
@@ -1,0 +1,81 @@
+; Copyright (C) Codeplay Software Limited
+;
+; Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+; Exceptions; you may not use this file except in compliance with the License.
+; You may obtain a copy of the License at
+;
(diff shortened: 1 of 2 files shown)


modules/compiler/test/lit/passes/subgroup-loop-unroll.ll
```diff
@@ -0,0 +1,65 @@
+; RUN: muxc --passes barriers-pass,verify < %s | FileCheck %s
```
frasercrmck, line 1: Needs a copyright header

AmyCodeplay: ah right you are

frasercrmck (approved):

PR #113, opened 2023-08-30
[vecz] Vectorize sub-groups on top of mux sub-groups

frasercrmck:
This PR finally pulls apart the conflation between sub-groups and vectorized group operations into two distinct concepts.

A sub-group operation (also called a mux sub-group) operates on a set of individual invocations of the kernel function. How this is done is ultimately up to the target (we don't have any examples in tree so this is largely academic at this point).

These operations can be vectorized, increasing the effective apparent sub-group size from the user's point of view (but crucially not from the target's). For example, two invocations of the kernel may be considered a mux sub-group. If the kernel is vectorized by 4, now each invocation does the work of 4 work-items and, as such, 8 work-items are being operated on by the mux sub-group. This is the actual sub-group size from the semantics of OpenCL.

We'd previously discarded mux sub-groups when vectorizing, replacing them entirely with vectorized group operations.

Now, with this PR, each mux sub-group operation is widened to suit the vectorization factor. The vectorized work-items form a "vector group", which operate *on top of* the mux sub-groups.

1. The vectorized sub-group size is the sub-group reduction sum of all vector group sizes
2. The vectorized sub-group local ID is now a step vector on top of the mux sub-group ID
3. Vectorized reductions (including any/all) reduce first over the vector group then over the sub-group
4. Vectorized broadcasts pull out the work-item from the vector group and broadcast that
5. Vectorized scans are more complicated:

Each vector group performs its own scan as before:

`<a,b,c,d> -> <a,a+b,a+b+c,a+b+c+d>` (mux sub-group item 0)
`<x,y,z,w> -> <x,x+y,x+y+z,x+y+z+w>` (mux sub-group item 1)

Each consecutive sub-group invocation needs access to the previous's scan results.

We reduce the operation over the vector:

`<a,b,c,d> -> <a+b+c+d>` (mux sub-group item 0)
`<x,y,z,w> -> <x+y+z+w>` (mux sub-group item 1)

Then perform the *exclusive scan of this*, essentially to shift these results over by one:

`<a+b+c+d> -> I` (mux sub-group item 0)
`<x,y,z,w> -> <a+b+c+d>` (mux sub-group item 1)

These results are then splatted and combined with the initial vector scan result:

`<a,a+b,a+b+c,a+b+c+d> -> <I+a,I+a+b,I+a+b+c,I+a+b+c+d>` (mux sub-group item 0)
`<x,x+y,x+y+z,x+y+z+w> -> <a+b+c+d+x,a+b+c+d+x+y,a+b+c+d+x+y+z,a+b+c+d+x+y+z+w>` (mux sub-group item 1)

When these two vector groups are viewed contiguously along the whole mux sub-group, this is the final scan.

### DIFF
--- a/modules/compiler/vecz/source/transform/packetizer.cpp
+++ b/modules/compiler/vecz/source/transform/packetizer.cpp
@@ -694,23 +694,38 @@ bool Packetizer::Impl::packetize() {
         continue;
       }
 
-      auto *const Callee = CI->getCalledFunction();
-      if (Callee && Ctx.builtins().analyzeBuiltin(*Callee).ID ==
+      if (auto *const Callee = CI->getCalledFunction();
+          Callee && Ctx.builtins().analyzeBuiltin(*Callee).ID ==
                         compiler::utils::eMuxBuiltinGetSubGroupSize) {
         auto *const replacement = [this](CallInst *CI) -> Value * {
+          // The vectorized sub-group size is the mux sub-group reduction sum
+          // of all of the vectorized sub-group sizes:
+          // |   mux 0     |      mux 1       |
+          // | < a,b,c,d > | < e,f,g > (vl=3) |
+          // The total sub-group size above is 4 + 3 => 7.
+          // Note that this expects that the mux sub-group consists entirely of
+          // equivalently vectorized kernels.
+          Value *VecgroupSize;
+          IRBuilder<> B(CI);
+          auto *const I32Ty = B.getInt32Ty();
           if (VL) {
-            return VL;
-          }
-
-          auto *const I32Ty = Type::getInt32Ty(F.getContext());
-          auto *const VFVal =
-              ConstantInt::get(I32Ty, SimdWidth.getKnownMinValue());
-          if (!SimdWidth.isScalable()) {
-            return VFVal;
+            VecgroupSize = VL;
           } else {
-            IRBuilder<> B(CI);
-            return B.CreateVScale(VFVal);
+            auto *const VFVal = B.getInt32(SimdWidth.getKnownMinValue());
+            if (!SimdWidth.isScalable()) {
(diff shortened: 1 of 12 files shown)


frasercrmck:
I should also point out, that mux sub-groups are currently always sized at 1, and this PR has no effect on the final generated code in tests I performed using the RISC-V target.

modules/compiler/vecz/source/transform/packetizer.cpp
```diff
@@ -1700,9 +1702,62 @@ ValuePacket Packetizer::Impl::packetizeSubgroupScan(
...
+  // Thus we essentially keep the original mux sub-group scan, but change it to
+  // be an exclusive one.
+  auto *Reduction = Ops.front();
+  if (VL) {
+    Reduction = sanitizeVPReductionInput(B, Reduction, VL, Scan.Recurrence);
+    if (!Reduction) {
+      return results;
+    }
+  }
+  Reduction = createSimpleTargetReduction(B, &TTI, Reduction, Scan.Recurrence);
```
AmyCodeplay, line 1740: this surely is just the same as the last element of the vectorized scan (at least, for inclusive scans) so why compute it again?

frasercrmck: Yes, good question - I thought I'd documented that, sorry.

I thought I'd do it this way because it means we treat inclusive scans, exclusive scans, and vector-predicated code the same way. In the absence of any performance analysis or sub-groups of any size other than 1, I like keeping the code as simple as possible.

frasercrmck: Ah yeah that would be an interesting avenue for optimization, I hadn't thought of that.

AmyCodeplay (approved):

PR #114, opened 2023-08-30
[compiler] Reduce uses of multi_llvm

frasercrmck:
This removes `multi_llvm/opaque_pointers.h` entirely as our pointers are always opaque.

This also reduces our uses of `multi_llvm::Optional` to just *one*, and hopefully we can remove that one soon, as well as the header, when we move to LLVM 16 minimum.


modules/compiler/utils/include/compiler/utils/metadata.h
```diff
@@ -249,15 +249,15 @@ multi_llvm::Optional<std::array<uint64_t, 3>> parseRequiredWGSMetadata(
 ///
 /// @return The maximum work dimension or llvm::None if there is no such
```
coldav, line 250: are all these comments now wrong - they don't return llvm::None an more but std::nullopt?

frasercrmck: oh, yes, good spot. I was only grepping for `multi_llvm::`

frasercrmck: done

coldav (approved):

PR #118, opened 2023-08-31
[compiler] Rename HandleBarriersPass -> WorkItemLoopsPass

frasercrmck:
The historical naming has long been confusing to newcomers (so much so that explicitly documented this fact in
doc/modules/compiler/overview.rst)

Its new name reflects the fact that it *primarily* is there to satisfy work-group scheduling requirements by adding loops over all work-items in the work-group. The fact that is "handles barriers" is a side-effect of doing this correctly.


doc/modules/compiler/utils.rst
```diff
@@ -181,16 +181,86 @@ functions with :ref:`mux-kernel <specifications/mux-compiler-spec:Function
...
+WorkItemLoopsPass
+-----------------
+
+The ``WorkItemLoopsPass`` is responsible for adding explicit parallelism to
+implicitly parallel SIMT kernels. It does so by wrapping each kernel up in a
+triple-nested loop over all work-items in the work-group. Thus, kernels
+scheduled by this pass can be invoked once per work-group.
+
+The order in which work-items are executed is fairly flexible as per the
+programming models the oneAPI construction kit supports, but generally in
```
coldav, line 193: we've tended to capitalize the `construction kit` elsewhere

doc/modules/compiler/utils.rst
```diff
@@ -181,16 +181,86 @@ functions with :ref:`mux-kernel <specifications/mux-compiler-spec:Function
...
+     size_t id = get_global_id(0);
+     out[id] = in[id] * 4;
+   }
+
+   void new_kernel(int *in, int *out) {
+     for (size_t z = 0; z != get_local_size(2); z++) {
+       for (size_t y = 0, y != get_local_size(1); y++) {
+         for (size_t x = 0, x != get_local_size(0); x++) {
+           size_t id = get_global_id(0);
+           out[id] = in[id] * 4;
```
coldav, line 212: it may be better to have an example that actually uses the z and y

doc/modules/compiler/utils.rst
```diff
@@ -181,16 +181,86 @@ functions with :ref:`mux-kernel <specifications/mux-compiler-spec:Function
...
-between these functions an analysis is done to create a struct of live variables
-which is passed as an argument to each kernel. Generated kernels then reference
-this struct rather than the original values. Work-item scheduling is also redone
-so that every work-item in a work-group invokes each kernel function before
-advancing.
+To propagate data dependencies between these functions an analysis
+is done to create a struct of live variables which is passed as an argument to
+each kernel. Generated kernels then reference this struct rather than the
+original values. Work-item scheduling is also redone so that every work-item in
+a work-group invokes each kernel function before advancing.
```
coldav, line 263: I wonder if it's worth mention that this is a vector across the workgroup for non-uniform values, since this is a source of confusion (or is it a vector of structs I can't remember)

I also seem to recall a different live struct for vector and scalar parts

coldav (approved):
This looks great, and makes things a lot clearer.

doc/modules/compiler/utils.rst
frasercrmck: Yeah. `get_local_linear_id`  implicitly uses all three dimensions - would that work? Maybe then we could replace it with

``` c
(z * get_local_size(1) * get_local_size(0)) + (y * get_local_size(0)) + x.
```

frasercrmck: Done just that

PR #120, opened 2023-09-05
[compiler] Ensure we link against static LLVM libraries

### LINKED ISSUE #109: CL_INVALID_PROGRAM (-44) on kernel enqueue
### Version

git commit 22eff6c

### What is your question or problem?

I'm trying to test hipSYCL's OpenCL backend with the oneAPI construction kit, and get some SYCL kernels to run on the CPU device that the construction kit exposes. It discovers the device just fine, however, when I call `clEnqueueNDRangeKernel`, I get error -44 (CL_INVALID_PROGRAM). The same happens with DPC++, so I assume that my SPIR-V is fine and the issue is probably somewhere in how the construction kit is built or set up.

I've built it like so:
```
cmake -DCA_LLVM_INSTALL_DIR=/usr/lib/llvm-15 -DCA_ENABLE_API=cl -DCMAKE_INSTALL_PREFIX=/path/to/install/dir /path/to/source
make install
```
Is there something obvious I am missing for this simple test case?

frasercrmck:
If we are building with a clang/LLVM built with
`LLVM_LINK_LLVM_DYLIB=ON`, then the static clang libraries we want to
link `compiler-base` against pull in the dynamic "LLVM" library. This
isn't the case for any of our locally built LLVMs, but is the case in
some distributed versions of clang/LLVM, e.g., Ubuntu's.

We are unable to safely link against LLVM's dynamic library for various
reasons ultimately stemming from LLVM's reliance on global state. As an
LLVM library ourselves, we may end up with some LLVM symbols in our
libCL library with others in libLLVM. In the more concrete case of
`llvm::Any` this leads to subtle breakages, as seen in issue #109.

I can't find a nice way around this other than by explicitly linking
against the full paths to the filenames of the static libraries.

Also removes some unused LLVM CMake code in muxc.

### DIFF
--- a/modules/compiler/source/base/CMakeLists.txt
+++ b/modules/compiler/source/base/CMakeLists.txt
@@ -70,9 +70,27 @@ target_include_directories(compiler-base PUBLIC
 target_include_directories(compiler-base SYSTEM PUBLIC
   ${spirv-headers_SOURCE_DIR}/include)
 
+# Explicitly link against the individual static versions of the clang
+# libraries, as if clang was built with LLVM_LINK_LLVM_DYLIB=ON, then the
+# 'clangCodeGen' library target brings in the 'LLVM' dynamic library. It is
+# not safe for us to link with LLVM's dynamic library, due to LLVM's use of
+# global/static state.
+foreach(CLANG_LIB "clangCodeGen" "clangFrontend" "clangDriver"
+                  "clangParse" "clangSerialization" "clangSema"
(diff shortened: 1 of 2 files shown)


coldav:
I think it would be useful to say in the commit message/description why we don't want to link with the shared library.

frasercrmck:
> I think it would be useful to say in the commit message/description why we don't want to link with the shared library.

Yes, totally. I thought I'd done that but apparently not. Thanks

PR #121, opened 2023-09-05
[host] Ensure we always check llvm::Error objects

frasercrmck:
In the case that no callback was provided, we weren't checking the `Error` objects before returning a generic error, which we must do or else LLVM will crash.

Since we can't do anything with the message, we just use `llvm::consumeError` which consumes it and marks it as checked without actually doing anything.


frasercrmck:
I don't think we can regression test this easily, since it both relies on an error occurring, while at the same time, `UnitCL` and our other test programs always set the callback that hides the issue.

coldav (approved):

PR #122, opened 2023-09-06
[compiler] Fix segfaults in work-group broadcasts

frasercrmck:
The code was assuming the specific value being broadcast was always available in the barrier/live-variables struct.

This is not the case when the value is previously analyzed as 'trivial' and is removed for space-saving reasons.

Note that in this case it's incorrect to simply rematerialize the value as, if it's based on the local ID, the value being broadcast will differ for each work-item receiving the broadcast, whereas we need to retrieve a specific (uniform) value computed by the work-item doing the broadcasting.

Doing this 'properly' whilst maintaining the struct trimming optimization would require either some compiler analysis to only remove values which are not (transitively) used by work-group broadcasts, or it would require special rematerialization logic to handle these values which are currently unsafe to rematerialize.

Since either of these approaches would be error-prone, a quick safe fix should be to keep the full barrier struct when any of the kernel's barriers is a work-group broadcast.

### DIFF
--- a/modules/compiler/utils/source/barrier_regions.cpp
+++ b/modules/compiler/utils/source/barrier_regions.cpp
@@ -38,6 +38,8 @@
 #include <multi_llvm/creation_apis_helper.h>
 #include <multi_llvm/multi_llvm.h>
 
+#include <optional>
+
 using namespace llvm;
 
 #define NDEBUG_BARRIER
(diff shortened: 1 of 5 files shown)


modules/compiler/utils/source/barrier_regions.cpp
```diff
@@ -406,26 +411,50 @@ void compiler::utils::Barrier::Run(llvm::ModuleAnalysisManager &mam) {
...
-      mam.invalidate(module_, PreservedAnalyses::allInSet<CFGAnalyses>());
-    }
-    // do the splitting first in case a value is used on both sides of a
-    // barrier within the same basic block.
-    SplitBlockwithBarrier();
-    FindLiveVariables();
+    return;
+  }
+
+  // If we found some barriers, we need to split up our kernel across them!
```
coldav, line 417: is the else no longer relevant here? the else seems to reflect finding barriers which this comment suggests it is assuming.

coldav (approved):

PR #125, opened 2023-09-12
[compiler] Vectorize to any required sub-group size

frasercrmck:
This updates users of the vectorizer to set the desired vectorization factor according to the kernel's required sub-group size, if present. It does so using a common method, to ensure consistent behaviour.This helper method assumes that the current underlying 'mux' sub-group size is 1, as we haven't got any in-tree targets that make use of anything larger than that.

All targets consider the required sub-group size less than they do `optnone` or `-cl-wfv=never` options. This can be revisited in the future if users desire different behaviour.

### DIFF
--- a/modules/compiler/vecz/source/pass.cpp
+++ b/modules/compiler/vecz/source/pass.cpp
@@ -30,6 +30,7 @@
 
 #include <cstdlib>
 #include <functional>
+#include <optional>
 #include <tuple>
 
 #include "vectorization_context.h"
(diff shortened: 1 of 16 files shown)


coldav (approved):

modules/compiler/vecz/source/pass.cpp
```diff
@@ -30,6 +30,7 @@
 
 #include <cstdlib>
 #include <functional>
+#include <optional>
```
ori-sky, line 33: Is this needed? We're adding `#include <optional>` to vecz/pass.h which is being included by this file.

ori-sky (changes requested):

modules/compiler/vecz/source/pass.cpp
frasercrmck: whoops sorry, already merged. It's not technically required but clang-tidy has started auto-including headers to avoid relying transitive dependencies. I guess making things more stable at the expense of compile time? I'm not sure about it, personally.

frasercrmck: Oh sorry it might be clangd: https://clangd.llvm.org/guides/include-cleaner#scenarios-and-solutions

PR #129, opened 2023-09-13
Added a sycl cts profile to city runner

coldav:
# Overview

Add a sycl cts profile to city runner to better catch the output

# Reason for change

The standard cts / default profile of city runner has a lot of code to handle OpenCL CTS and fails to catch all cases in the SYCL CTS.

# Description of change

This is a simplified version of the parsing to reflect the output. We try to catch the following cases in the output:

    test cases: with number of pass and fails
    All tests passed : with no information on how many
    No tests ran (sometimes used for skipping)

Due to the way it reports we don't try to captures numbers but just check for some form of all passed or failed


scripts/testing/city_runner/profiles/sycl_cts.py
```diff
@@ -0,0 +1,101 @@
...
+
+        # There are 3 main possible outputs:
+        #   test cases: with number of pass and fails
+        #   All tests passed : with no information on how many
+        #   No tests ran (sometimes used for skipping)
+        # Due to the way it reports we don't try to captures numbers but just check
+        # for some form of all passed or failed
+        test_cases_pattern = re.compile(b"^test cases: (.*$)")
+        all_test_passed_pattern = re.compile(b"^(All tests passed)")
+        no_test_run_pattern = re.compile(b"^(No tests ran$)")
```
frasercrmck, line 55: Do we need the `()` in this regex?

scripts/testing/city_runner/profiles/sycl_cts.py
```diff
@@ -0,0 +1,101 @@
...
+        stream = io.BytesIO(self.output)
+
+        # There are 3 main possible outputs:
+        #   test cases: with number of pass and fails
+        #   All tests passed : with no information on how many
+        #   No tests ran (sometimes used for skipping)
+        # Due to the way it reports we don't try to captures numbers but just check
+        # for some form of all passed or failed
+        test_cases_pattern = re.compile(b"^test cases: (.*$)")
+        all_test_passed_pattern = re.compile(b"^(All tests passed)")
```
frasercrmck, line 54: Do we need the `()` in this regex? We're not using the resulting match group.

frasercrmck (commented):
There are other changes we made to the 'CTS' profile in 66523c15a6eda93f8ae1bf852db6c90713c90af5 - do we want to undo those as well?

It might be nice for the PR description and commit message if there was an example of what it parses and what it reports in each case. It's not immediately obvious to me how this works. We only report 1 or 0 pass per test invocation?

coldav:
> There are other changes we made to the 'CTS' profile in [66523c1](https://github.com/codeplaysoftware/oneapi-construction-kit/commit/66523c15a6eda93f8ae1bf852db6c90713c90af5) - do we want to undo those as well?
> 
> It might be nice for the PR description and commit message if there was an example of what it parses and what it reports in each case. It's not immediately obvious to me how this works. We only report 1 or 0 pass per test invocation?

I'll try and clarify it. Based on the output it seems a single pass, fail or skip was best for each individual run of  a test.
I don't think there is any other change directly to the CTS profile.

frasercrmck:
> 
> I don't think there is any other change directly to the CTS profile.

The commit I linked added a regex for `All tests passed` which we probably don't need anymore. I guess the `shlex.split` is good to keep? Dunno.

scripts/testing/city_runner/profiles/sycl_cts.py
```diff
@@ -0,0 +1,127 @@
...
+        elif skipped:
+            self.status = "SKIP"
+        else:
+            stdout.write(
+                "WARNING: Unable to tell if test pass or fail")
+            self.status = "FAIL"
+
+        # Special cases e.g SIGTERM
+        if self.return_code:
+            TestRunBase.analyze_process_output(self)
```
frasercrmck, line 119: Should this `return` or will this fall through? Is that okay?

frasercrmck (approved):
Nicely documented, thanks!

PR #134, opened 2023-09-19
[riscv][vecz] Auto-vectorize sub-group kernels to a device-specific sub-group size

frasercrmck:
This PR updates the riscv/refsi targets to automatically vectorize kernels/functions containing explicit use of sub-groups to one of the device's advertised sub-group widths.

The rationale for this is as follows: The SYCL 2020 spec has `info::device::sub_group_sizes` which returns "**_the_**" list of sub-group sizes reported by the device. The CTS expects that if it runs a kernel which (for example) returns the kernel's sub-group size to the host, then that size is found in that same device list.

This rules out the use of degenerate sub-groups. We'd have to list all the sub-group (read: work-group) sizes from 1 to 1024, which is useless for users. All of those sub-group sizes could then, in turn, be _required_ by a user, and the compiler would effectively be unable to satisfy. A user requiring a sub-group size of 244 would lead to a very strange situation where the compiler must create a degenerate sub-groups kernel and would then just _hope_ that the user uses a work-group size of 244. Anything else would be invalid. This is another bad user experience.

The vectorizer can be asked to return a suitable vectorization factor for a kernel, given its use of sub-groups and the device's list of sub-group sizes. It is quite rudimentary in its heuristic and doesn't choose between legal powers of two: the first one is taken as the preferred list.

I've filed a ticket for clarification/questions about sub-groups and degenerate sub-groups against the SYCL spec (see https://github.com/KhronosGroup/SYCL-Docs/issues/463), but I think this is a suitable "workaround" for now, or even for the long term. The behaviour is arguably more intuitive for users.

This hasn't been done on `host` due to issues with its compiler framework and its current inability to handle multiple online JIT kernels. That will come later.

### DIFF
--- a/modules/compiler/vecz/include/vecz/pass.h
+++ b/modules/compiler/vecz/include/vecz/pass.h
@@ -66,7 +66,23 @@ struct VeczPassOptions {
   uint64_t local_size;
 };
 
+/// @brief Returns the vectorization options that would vectorize the provided
+/// function to its required sub-group size.
 std::optional<VeczPassOptions> getReqdSubgroupSizeOpts(llvm::Function &);
+
+/// @brief Returns the vectorization options that would vectorize the provided
+/// function to its required sub-group size (if set) or one of the device's
+/// sub-group sizes.
+///
+/// Only returns options if the function uses sub-group operations, as
+/// determined by the SubGroupAnalysis pass.
+///
+/// Tries to find a good fit that produces one of the device's sub-group sizes,
+/// preferring ones which fit the known local work-group size and powers of
+/// two. The device's sub-group sizes can be sorted such that preferable sizes
+/// are placed towards the front.
+std::optional<VeczPassOptions> getAutoSubgroupSizeOpts(
+    llvm::Function &, llvm::ModuleAnalysisManager &);
 
 /// @brief Analysis pass which determines on which functions @ref RunVeczPass
 /// should operate.
--- a/modules/compiler/vecz/source/pass.cpp
+++ b/modules/compiler/vecz/source/pass.cpp
@@ -20,6 +20,7 @@
 #include <compiler/utils/builtin_info.h>
 #include <compiler/utils/device_info.h>
 #include <compiler/utils/metadata.h>
+#include <compiler/utils/sub_group_analysis.h>
 #include <compiler/utils/vectorization_factor.h>
 #include <llvm/IR/DebugInfoMetadata.h>
 #include <llvm/IR/LLVMContext.h>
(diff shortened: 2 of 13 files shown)


modules/compiler/vecz/include/vecz/pass.h
```diff
@@ -66,8 +66,24 @@ struct VeczPassOptions {
   uint64_t local_size;
 };
 
+/// @brief Returns the vectorization options that would vectorize the provided
+/// function to its required sub-group size.
 std::optional<VeczPassOptions> getReqdSubgroupSizeOpts(llvm::Function &);
 
+/// @brief Returns the vectorization options that would vectorize the provided
+/// function to its required sub-group size (if se) or one of the device's
```
coldav, line 74: if se?

modules/compiler/vecz/source/pass.cpp
```diff
@@ -262,4 +271,96 @@ std::optional<VeczPassOptions> getReqdSubgroupSizeOpts(Function &F) {
...
+
+  // Now try and choose the best width.
+  std::optional<unsigned> best_width;
+  auto const mux_sub_group_size = compiler::utils::getMuxSubgroupSize(F);
+
+  auto can_produce_legal_width = [&mux_sub_group_size](unsigned size) {
+    // We only support vectorization widths where there's a clean multiple, and
+    // we can vectorize *up* to the desired size - we can't shrink the
+    // sub-group size by vectorizing.
+    return size >= mux_sub_group_size && size % mux_sub_group_size == 0;
```
coldav, line 314: I'd prefer a few brackets here for readability

modules/compiler/vecz/include/vecz/pass.h
frasercrmck: `if set`. fixed that typo now

modules/compiler/vecz/source/pass.cpp
frasercrmck: Around `size % mux_sub_group_size`?

modules/compiler/vecz/source/pass.cpp
```diff
@@ -262,4 +271,96 @@ std::optional<VeczPassOptions> getReqdSubgroupSizeOpts(Function &F) {
...
+  };
+
+  for (auto size : DI.reqd_sub_group_sizes) {
+    if (!can_produce_legal_width(size)) {
```
coldav, line 318: if i understand this correctly it will only replace the first one if it is a:
*non-scalar
* required WGS related
* power of 2

Does this mean we have a preferred ordering in terms of priority for  DI.reqd_sub_group_sizes

frasercrmck: Aye that's right. Well, plus "not 1". I didn't want to encode device-specific logic at this stage.

I think I documented that it's preferred ordering, but documenting it in the mux side of things seems wrong to me as I don't think we're ready to make it part of the mux "spec". This is more of a compiler implementation detail.

coldav (approved):

modules/compiler/vecz/tools/source/veczc.cpp
```diff
@@ -372,10 +378,15 @@ int main(const int argc, const char *const argv[]) {
...
   passMach.getMAM().registerPass(
       [&] { return compiler::utils::BuiltinInfoAnalysis(); });
+  passMach.getMAM().registerPass(
+      [&] { return compiler::utils::SubgroupAnalysis(); });
   passMach.getFAM().registerPass([] { return llvm::TargetIRAnalysis(); });
   passMach.getMAM().registerPass([] {
     compiler::utils::DeviceInfo Info{/*half*/ 0, /*float*/ 0, DoubleSupport,
                                      /*MaxWorthWidth*/ 64};
+    for (const auto S : SGSizes) {
+      Info.reqd_sub_group_sizes.push_back(S);
```
coldav, line 388: I'm wondering whether this is a good name, given there is the ability to suggest a sub group size in the language? Maybe supported_sub_group_sizes?

PR #136, opened 2023-09-19
[ci] Test OpenCL CTS subgroups during MR testing

frasercrmck:
These are relatively quick (taking less than a minute total) and might help catch regressions in sub-group conformance before merging.

Note that we currently only run the OpenCL CTS for riscv during MRs.


coldav (approved):

frasercrmck:
This is in limbo because the RefSI WIPT target doesn't correctly implement sub-group barriers (if I recall correctly). So we'd either have to fix that or add a WIPT-specific CTS ignore file. The latter is probably okay, but I haven't got round to it yet.

frasercrmck:
I've disabled these tests on that target (in our internal CI) so this is good to go.

PR #137, opened 2023-09-20
[NATIVECPU] Enabling using the Construction Kit as an LLVM subproject

PietroGhg:
# Overview

This PR is a first attempt at enabling using the OneAPI Construction Kit as an LLVM subproject. It contains only changes to the CMake files, and overall using the OCK as subproject is still W.I.P.

# Reason for change

These changes are needed because we aim to integrate the OCK in DPC++ (https://github.com/intel/llvm), and use it to implement the SYCL Native CPU target.

# Description of change

Changes in the CMake that allow to build only a few components (what we need for the DPC++ integration) as LLVM libraries.


modules/compiler/CMakeLists.txt
```diff
@@ -14,8 +14,11 @@
 #
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+if(CA_NATIVE_CPU)
+  add_subdirectory(multi_llvm)
+  add_subdirectory(utils)
```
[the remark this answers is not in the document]
PietroGhg: Good call, thank you

modules/CMakeLists.txt
```diff
@@ -44,6 +39,17 @@ FetchContent_Declare(
 )
 FetchContent_MakeAvailable(SPIRV-Headers)
 
+if(CA_NATIVE_CPU)
```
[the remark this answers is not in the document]
PietroGhg: `FetchContent` needs to be called before adding `mux` and `compiler`, but I'd like to avoid adding `debug`, `loader` and `utils` for native cpu. I've refactored my changes so that the intent is more clear (hopefully)

CMakeLists.txt
```diff
@@ -155,6 +159,10 @@ foreach(OPTIONS_MODULE ${OPTIONS_MODULES})
   endif()
 endforeach()
 
+if(CA_NATIVE_CPU)
+  set(CA_RUNTIME_COMPILER_ENABLED Off)
```
coldav, line 163: it would be good to have a comment explaining why you needed to do this (it's slightly counter-intuitive, but ultimately I think it relates to what needs to be built).

modules/metadata/CMakeLists.txt
```diff
@@ -17,16 +17,26 @@
...
-    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/include/metadata/detail)
-
 add_ca_library(md_handler STATIC
-    ${CMAKE_CURRENT_SOURCE_DIR}/source/metadata/handler/generic_metadata.cpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/source/metadata/handler/vectorize_info_metadata.cpp)
-target_link_libraries(md_handler PUBLIC cargo md_api)
-target_include_directories(md_handler PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
+  ${CMAKE_CURRENT_SOURCE_DIR}/source/metadata/handler/generic_metadata.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/source/metadata/handler/vectorize_info_metadata.cpp)
+if(CA_NATIVE_CPU)
```
coldav, line 23: most of the if/else is the same, maybe refactor to just show the difference?
Also add a comment to explain.

CMakeLists.txt
PietroGhg: I've added the comment, thank you

modules/metadata/CMakeLists.txt
PietroGhg: Good point, looks better now

coldav (approved):

coldav (approved):

PR #139, opened 2023-09-21
Fix stdout usage in city runner python script

coldav:
# Overview

Fixed sycl_cts profile when no output

# Reason for change

The sycl_cts profile was using stdout without importing it.

# Description of change

Added import for stdout


frasercrmck (approved):
LGTM

PR #142, opened 2023-09-25
[refsi] Fix AUBSAN error on zero-sized argument packs

frasercrmck:
When there are no arguments to pack, the argument 'data' we want to store to the dummy address is null. `MemoryController::store` doesn't expect to be able to store null bytes, as it passes it off to memcpy, which results in undefined behaviour.


hvdijk:
The commit message, there is no problem with storing null bytes, is there? The change itself looks good though, and I don't necessarily mind a not 100% accurate commit message.

frasercrmck:
> The commit message, there is no problem with storing null bytes, is there? The change itself looks good though, and I don't necessarily mind a not 100% accurate commit message.

I think my wording is unclear. There's no problem with storing the value `null` to a memory address, but there's a problem with passing `null` as either `src` or `dest` to `std::memcpy` (undefined behaviour). Maybe that doesn't apply to C  `memcpy`? I can't find a definitive source.

Since `MemoryController::store` uses `memcpy` under the hood, it also can't accept `nullptr`. We could maybe also more clearly document this, and/or add an assertion, and/or add a "if size is zero, return" case too?

hvdijk:
> I think my wording is unclear. There's no problem with storing the value `null` to a memory address,

You write in the description and commit message: "MemoryController::store doesn't expect to be able to store null bytes". It's the "store null bytes" that doesn't sound right to me.

> but there's a problem with passing `null` as either `src` or `dest` to `std::memcpy` (undefined behaviour). Maybe that doesn't apply to C `memcpy`? I can't find a definitive source.

That applies to C's `memcpy` too. C99 7.21.1p2  "Unless explicitly stated otherwise in the description of a particular function in this subclause, pointer arguments on such a call shall still have valid values, as described in 7.1.4." And 7.1.4p1 "Each of the following statements applies unless explicitly stated otherwise in the detailed descriptions that follow: If an argument to a function has an invalid value (such as [...] a null pointer [...]) [...], the behavior is undefined." A null pointer is considered an "invalid value" and has undefined behavior for any library string function that does not specifically say it is supported.

> Since `MemoryController::store` uses `memcpy` under the hood, it also can't accept `nullptr`. We could maybe also more clearly document this, and/or add an assertion, and/or add a "if size is zero, return" case too?

Documentation and assertions are always good. A "if size is zero, do nothing, and do not require pointers to be valid" is something that I would be in favor of for `memcpy`, but there is no hope of that happening, and since that has no hope of happening, I personally think it is more valuable for us to be consistent with the standard library function than it is to provide a better version of it.

hvdijk (approved):

PR #143, opened 2023-09-26
Add a script to test vgg16 and resnet50 with oneapi-construction-kit, oneapi, portBLAS and SYCL-DNN

MaryaSharf:
# Overview

Adding a script to build portBLAS and SYCL-DNN, and download and build vgg16 and resnet50

# Reason for change

The purpose of this script is to add testing of VGG16 and RESNET50 with oneAPI.

# Description of change

This script assumes that you have already cloned the portBLAS and SYCL-DNN repositories. Additionally, you should have cloned and built the oneAPI Construction Kit. Ensure that oneAPI itself has been built in advance as well. The build dirs and cloned repositories need to be set as environment variables; 

- portBLAS_source
- SYCL_DNN_source
- path_to_oneapi 
- OCK_libCL_path
- input_image 

# Checklist

### DIFF
--- a/examples/technical_blogs/tartan_blog/build_and_run_networks.py
+++ b/examples/technical_blogs/tartan_blog/build_and_run_networks.py
@@ -1,0 +1,244 @@
+#!/usr/bin/env python
+# Copyright (C) Codeplay Software Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+# Exceptions; you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+import os
+import sys
+import shutil
+import requests
+import argparse
+import subprocess
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(description='The purpose of this script is to build and execute VGG16 and '
+                                     'Resnet50 networks using the oneAPI framework and the oneapi-construction-kit.'
+                                     ' It requires specifying paths to the oneapi_construction_kit build directory '
+                                     'and oneapi_base_toolkit. Additionally, portBLAS and portDNN are necessary to '
+                                     'run and test the networks, but they will be cloned if the paths are not provided.'
+                                     ' An input image path should be provided to test using the networks, in the absence'
+                                     ' of image, the networks will only be downloaded and run.')
+
+    # Prereqs: oneapi oneapi_construction_kit should be cloned and installed.
+    # portDNN and portBLAS will be cloned if paths are not provided.
+    parser.add_argument('--portBLAS_source', help='Provide path to portBLAS source.')
+    parser.add_argument('--clone_portBLAS', action='store_true', help='Include this flag to clone repo if path to portBLAS_source is not provided.')
+    parser.add_argument('--portDNN_source', help='Provide path to portDNN source.')
+    parser.add_argument('--clone_portDNN',action='store_true', help='Include this flag to clone repo if path to portDNN is not provided')
+    parser.add_argument('--oneapi_construction_kit_build_dir', required=True, help='Path to oneAPI construction kit')
+    parser.add_argument('--oneapi_base_toolkit', required=True, help='Path to oneAPI base toolkit')
+    parser.add_argument('--input_image', help='Path to the input image to be processed using the VGG16 and Resnet50.')
+    return parser.parse_args()
+
+# Set environment variables
+def set_env_vars(args):
+    os.environ["CMAKE_CXX_COMPILER"] = os.path.join(args.oneapi_base_toolkit, "compiler", "linux", "bin-llvm", "clang++")
+    os.environ["CMAKE_C_COMPILER"] = os.path.join(args.oneapi_base_toolkit, "compiler", "linux", "bin-llvm", "clang")
+    os.environ["LD_LIBRARY_PATH"] = f"{os.path.join(args.oneapi_base_toolkit, 'compiler', 'linux', 'lib')}:{os.path.join(args.oneapi_base_toolkit, 'compiler', 'linux', 'compiler', 'lib', 'intel64_lin')}:{os.environ.get('LD_LIBRARY_PATH', '')}"
+    os.environ["OCL_ICD_FILENAMES"] = os.path.join(args.oneapi_construction_kit_build_dir, 'lib', 'libCL.so')
+
+
+def run_cmake_and_ninja(cmake_command, build_dir):
+    if os.path.exists(build_dir):
+        print(f"{build_dir} exists,skipping building.")
+    else:
+        try:
+            subprocess.run(cmake_command, check=True)
+            subprocess.run(["ninja", "-C", build_dir], check=True)
+        except subprocess.CalledProcessError as e:
+            print(f"Error executing command: {e}", file=sys.stderr)
+            sys.exit(1)
+
+# Clone repository if not present
+def clone_git_repository(repo_url, destination_directory):
+    try:
+        subprocess.run(["git", "clone", "--recursive", repo_url, destination_directory], check=True)
+        print(f"Repository cloned successfully to {destination_directory}")
+    except subprocess.CalledProcessError as e:
+        print(f"Error cloning repository: {e}", file=sys.stderr)
+        sys.exit(1)
+
+# Build portBLAS
+def build_portBLAS(args):
+    script_directory = os.getcwd()
+    portBlas_build_dir = os.path.join(script_directory, "build_portBlas")
+
+    # Check if path to portBLAS is present, else clone portBLAS
+    if args.portBLAS_source is not None:
+        portBLAS_source = args.portBLAS_source
+    elif args.clone_portBLAS:
+        portBLAS_repo_url = "https://github.com/codeplaysoftware/portBLAS.git"
+        portBLAS_source = os.path.join(script_directory, "portBLAS")
+        if os.path.exists(portBLAS_source):
+            print(f"portBLAS exists. Continueing with the cmake.")
+        else:
+            clone_git_repository(portBLAS_repo_url, portBLAS_source)
+    else:
+        print(f'Error: portBLAS not found. Pass portBLAS_source or enable clone_portBLAS.', file=sys.stderr)
+        sys.exit(1)
+
+    cmake_portBlas_command = [
+        "cmake",
+        f"-B {portBlas_build_dir}",
+        f"{portBLAS_source}",
+        "-GNinja",
+        "-DSYCL_COMPILER=dpcpp"
+    ]
+
+    run_cmake_and_ninja(cmake_portBlas_command, portBlas_build_dir)
+
+# Build portDNN
+def build_portDNN(args):
+    script_directory = os.getcwd()
+    portdnn_build_dir = os.path.join(script_directory, "build_portdnn")
+    
+    # Check if path to portDNN is present, else clone portDNN
+    if args.portDNN_source is not None:
+        portDNN_source = args.portDNN_source
+    elif args.clone_portDNN:
+        portDNN_repo_url = "https://github.com/codeplaysoftware/portDNN.git"
+        portDNN_source = os.path.join(script_directory, "portDNN")
+        if os.path.exists(portDNN_source):
+            print(f"portDNN exists. Continueing with the cmake.")
+        else:
+            clone_git_repository(portDNN_repo_url, portDNN_source)
+    else:
+        print(f'Error: portDNN not found. Pass portDNN_source or enable clone_portDNN.', file=sys.stderr)
+        sys.exit(1)
+
+    cmake_portdnn_command = [
+        "cmake",
+        f"-B {portdnn_build_dir}",
+        f"{portDNN_source}",
+        "-GNinja",
+        f"-DCMAKE_CXX_COMPILER={os.environ['CMAKE_CXX_COMPILER']}",
+        "-DSNN_BUILD_BENCHMARKS=OFF",
+        "-DSNN_BENCH_SYCLBLAS=OFF",
+        "-DSNN_BUILD_DOCUMENTATION=OFF",
+        f"-DSyclBLAS_DIR='{os.path.join(script_directory, 'build_portBlas')}'"
+    ]
+
+    run_cmake_and_ninja(cmake_portdnn_command, portdnn_build_dir)
+    return portDNN_source
+
+
+def download_and_convert_model(dir, file_url, destination_file, h5_to_bin_script):
+    # Check if the model dir already exists
+    path = os.path.join(os.getcwd(), dir)
+    if os.path.exists(path):
+        print(f"'{path}' already exists. Skipping creating '{path}'.")
+    else:
+        os.mkdir(path)
+        print(f"Directory '{path}' created successfully.")
+
+    # Check if the file already exists
+    if os.path.exists(destination_file):
+        print(f"'{destination_file}' already exists. Skipping download.")
+    else:
+        try:
+            response = requests.get(file_url)
+            response.raise_for_status()
+            with open(destination_file, "wb") as file:
+                file.write(response.content)
+            print(f"Downloaded '{destination_file}' successfully.")
+        except requests.exceptions.RequestException as e:
+            print(f"Error downloading file: {e}", file=sys.stderr)
+            sys.exit(1)
+        try:
+            subprocess.run(["python3", h5_to_bin_script, destination_file], check=True, cwd=path)
+            print(f"Converted '{destination_file}' to binary successfully.")
+        except subprocess.CalledProcessError as e:
+            print(f"Error running h5toBin.py: {e}", file=sys.stderr)
+            sys.exit(1)
+
+
+def prepare_image(img_script_path, input_image):
+    try:
+        subprocess.run(["python3", img_script_path, input_image], check=True)
+        print(f"Image prepared successfully.")
+    except subprocess.CalledProcessError as e:
+        print(f"Error executing the script: {e}", file=sys.stderr)
+        sys.exit(1)
+
+# Test on images
+def test_image(model_script, data_dir, output_file):
+    command_args = [
+        model_script,
+        data_dir + "/",
+        output_file
+    ]
+    try:
+        subprocess.run(command_args, check=True)
+        print(f"Image tested successfully with {os.path.basename(model_script)}.")
+    except subprocess.CalledProcessError as e:
+        print(f"Error executing the command: {e}", file=sys.stderr)
+        sys.exit(1)
+
+
+def main():
+    args = parse_arguments()
+    set_env_vars(args)
+    build_portBLAS(args)
+    portDNN_source = build_portDNN(args)
+    portdnn_build_dir = os.path.join(os.getcwd(), "build_portdnn")
+    vgg_dir = os.path.join(os.getcwd(), "vdata")
+    resnet_dir = os.path.join(os.getcwd(), "rdata")
+
+    # Set the environment variables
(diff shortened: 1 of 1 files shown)


examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
+import subprocess
```
frasercrmck, line 1: Do we need a copyright header on this file?

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
...
+        print(f"Directory '{path}' removed successfully.")
+    os.mkdir(path)
+    print(f"Directory '{path}' created successfully.")
+
+def run_cmake_and_ninja(cmake_command):
+    try:
+        subprocess.run(cmake_command, check=True)
+        subprocess.run("ninja", check=True)
+    except subprocess.CalledProcessError as e:
+        print(f"Error executing command: {e}")
```
frasercrmck, line 31: Errors would be best printed to `stderr`

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
...
+portBLAS_source = os.environ.get("portBLAS_source")
+SYCL_DNN_source = os.environ.get("SYCL-DNN_source")
+path_to_oneapi = os.environ.get("path_to_oneapi")
+OCK_libCL_path = os.environ.get("OCK_libCL_path")
+input_image = os.environ.get("input_image")
+
+# Set environment variables
+os.environ["CMAKE_CXX_COMPILER"] = os.path.join(path_to_oneapi, "compiler/linux/bin-llvm/clang++")
+os.environ["LD_LIBRARY_PATH"] = f"{os.path.join(path_to_oneapi, 'compiler/linux/lib/')}:{os.path.join(path_to_oneapi, 'compiler/linux/compiler/lib/intel64_lin')}:{os.environ.get('LD_LIBRARY_PATH', '')}"
+os.environ["CXX"] = os.environ["CMAKE_CXX_COMPILER"]
```
frasercrmck, line 16: Who's using the `CXX` and `CC` environment variables? If it's CMake, would using `CMAKE_CXX_COMPILER` and `CMAKE_C_COMPILER` work? Just seeing a potential bit of redundancy between these two sets of variables.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
...
+
+# Prereqs: portBLAS, SYCL-DNN, oneapi oneapi_construction_kit should be cloned and installed.
+portBLAS_source = os.environ.get("portBLAS_source")
+SYCL_DNN_source = os.environ.get("SYCL-DNN_source")
+path_to_oneapi = os.environ.get("path_to_oneapi")
+OCK_libCL_path = os.environ.get("OCK_libCL_path")
+input_image = os.environ.get("input_image")
+
+# Set environment variables
+os.environ["CMAKE_CXX_COMPILER"] = os.path.join(path_to_oneapi, "compiler/linux/bin-llvm/clang++")
```
frasercrmck, line 14: I dunno if we want to be more generic and avoid path separators everywhere, e.g., `os.path.join(path_to_oneapi, "compiler", "linux", "bin-llvm", "clang++")`? I notice you do that sort of thing further down in this script.

I don't mind, as long as we're consistent. It doesn't look like we're going to be running this on Windows anyway.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
...
+os.environ["CC"] = os.path.join(path_to_oneapi, "compiler/linux/bin-llvm/clang")
+
+def create_directory(path):
+    if os.path.exists(path):
+        shutil.rmtree(path)
+        print(f"Directory '{path}' removed successfully.")
+    os.mkdir(path)
+    print(f"Directory '{path}' created successfully.")
+
+def run_cmake_and_ninja(cmake_command):
```
frasercrmck, line 26: I think this could take the build directory as a parameter, and we invoke `cmake -B f{build_dir}` && `ninja -C f{build_dir}`.

Then people can use this function "anywhere" and don't have to `cd` into the build directory and back out again afterwards.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
...
+    os.environ["CA_HAL_DEBUG"] = "1"
+    os.environ["CA_PROFILE_LEVEL"] = "3"
+    os.environ["OCL_ICD_FILENAMES"] = f"{OCK_libCL_path}"
+    os.environ["ONEAPI_DEVICE_SELECTOR"] = "opencl::acc"
+    os.environ["SYCL_CONFIG_FILE_NAME"] = "null.cfg"
+
+    # VGG16
+    vgg_dir = "vdata"
+    create_directory(vgg_dir)
+    os.chdir(vgg_dir)
```
frasercrmck, line 126: I generally don't like scripts changing directories as it's harder for users to recreate what's happening inside the script (e.g., by copying commands that are `echo`d to the terminal). Or maybe even the script will error, not change from `foo/bar` into `foo/bar/baz`, but will then change into `..` and end up in `foo`, not back in `foo/bar`. Then the script's in a weird state.

Here, can we set `destination_file` to to `vgg_dir/{destination_file}` and avoid changing directories? It feels as though everything will "just" work, but I don't know the nitty-gritty.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
...
+        response = requests.get(file_url)
+        response.raise_for_status()
+        with open(destination_file, "wb") as file:
+            file.write(response.content)
+        print(f"Downloaded '{destination_file}' successfully.")
+        
+        subprocess.run(["python", h5_to_bin_script, destination_file], check=True)
+        print(f"Converted '{destination_file}' to binary successfully.")
+    except requests.exceptions.RequestException as e:
+        print(f"Error downloading file: {e}")
```
frasercrmck, line 85: Should the script terminate if it catches an exception in any of these files? Or should it try and continue on? Maybe we should re-throw an exception and catch it at the top-level? `try: main() except ...`?

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
...
+    ]
+    try:
+        subprocess.run(command_args, check=True)
+        print(f"Image tested successfully with {os.path.basename(model_script)}.")
+    except subprocess.CalledProcessError as e:
+        print(f"Error executing the command: {e}")
+
+
+if __name__ == "__main__":
+    build_portBLAS()
```
frasercrmck, line 112: Might be good to chuck this all in a `main` function and call main from here.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
+import subprocess
+import os
+import shutil
+import requests
+
+# Prereqs: portBLAS, SYCL-DNN, oneapi oneapi_construction_kit should be cloned and installed.
+portBLAS_source = os.environ.get("portBLAS_source")
```
frasercrmck, line 7: Having these as arguments would be more user-friendly. I don't know how much of a priority that is for this script.

Using `argparse`, for example, would allow users to call `--help` and see what arguments they're meant to be supplying. Especially if these environment variables are only used as script inputs, which they appear to be.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
...
+        print(f"Error executing the command: {e}")
+
+
+if __name__ == "__main__":
+    build_portBLAS()
+    build_sycl_dnn()
+    sycl_dnn_build_dir = os.path.join(os.getcwd(), "build_sycl_dnn")
+
+    # Set the environment variables
+    os.environ["CA_HAL_DEBUG"] = "1"
```
frasercrmck, line 117: I wonder if we want to set some of these based on the existing environment, to allow users to override:

``` py
 os.environ["CA_HAL_DEBUG"] = os.environ.get("CA_HAL_DEBUG", "1")
```

Then users can set their own `CA_HAL_DEBUG`, but `1` will be used as a default.

Just a minor idea.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # VGG16
+    vgg_dir = "vdata"
+    create_directory(vgg_dir)
+    os.chdir(vgg_dir)
+    file_url = "https://storage.googleapis.com/tensorflow/keras-applications/vgg16/vgg16_weights_tf_dim_ordering_tf_kernels.h5"
+    destination_file = "vgg16_weights_tf_dim_ordering_tf_kernels.h5"
+    h5_to_bin_script = os.path.join(SYCL_DNN_source, "samples", "networks", "vgg", "h5toBin.py")
+    
+    download_and_convert_model(file_url, destination_file, h5_to_bin_script)
```
frasercrmck, line 131: I wonder if google's APIs would block a user after requesting to download this file over and over in a short period of time? E.g., if a developer repeatedly runs this scripts but keeps hitting errors during development.

The only way around that would be to check if the destination file already exists and either a) trust that it's correct or b) use some kind of hash to check whether the contents are identical?

examples/technical_blogs/tartan_blog/build_and_run_networks.py
hvdijk: CMake can either take `CC`/`CXX` from the environment, or `CMAKE_C_COMPILER`/`CMAKE_CXX_COMPILER` from the command line. A `CMAKE_CXX_COMPILER` environment variable, as far as I am aware, would not be used by anything, so the simplest way to remove redundancy would be to remove that one.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
MaryaSharf: I can look into it, but for now I am trusting that if the file exisits already then its the correct one.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,187 @@
...
+        "-DSNN_BENCH_SYCLBLAS=OFF",
+        "-DSNN_BUILD_DOCUMENTATION=OFF",
+        f"-DSyclBLAS_DIR='{os.path.join(script_directory, 'build_portBlas')}'"
+    ]
+
+    run_cmake_and_ninja(cmake_sycl_dnn_command, sycl_dnn_build_dir)
+
+
+def download_and_convert_model(dir, file_url, destination_file, h5_to_bin_script):
+    #Check if the model dir already exists
```
coldav, line 88:
```suggestion
    # Check if the model dir already exists
```

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,187 @@
...
+        try:
+            response = requests.get(file_url)
+            response.raise_for_status()
+            with open(destination_file, "wb") as file:
+                file.write(response.content)
+            print(f"Downloaded '{destination_file}' successfully.")
+        
+            subprocess.run(["python", h5_to_bin_script, destination_file], check=True)
+            print(f"Converted '{destination_file}' to binary successfully.")
+            os.chdir("..")  # Move up one level to the parent directory
```
coldav, line 109: could we avoid the chdir by just making the destination_file path include the sub directory?

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,187 @@
...
+parser = argparse.ArgumentParser(description='The purpose of this script is to build and run '
+                                 'VGG16 and Resnet50 networks with oneAPI and oneapi-construction-kit.')
+
+# Prereqs: portBLAS, SYCL-DNN, oneapi oneapi_construction_kit should be cloned and installed.
+parser.add_argument('--portBLAS_source', required=True, help='Path to portBLAS source')
+parser.add_argument('--SYCL_DNN_source', required=True, help='Path to SYCL-DNN source')
+parser.add_argument('--oneapi_construction_kit_build_dir', required=True, help='Path to oneAPI construction kit')
+parser.add_argument('--oneapi_base_toolkit', required=True, help='Path to oneAPI base toolkit')
+parser.add_argument('--input_image', required=True, help='Path to the input image to be processed using the VGG16 and Resnet50.')
+
```
coldav, line 34: would be nice to be able to skip some parts e.g. just run the network.

coldav, line 34: maybe add a verbose option - it looks like it might just go away for a long time?

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,187 @@
...
+os.environ["LD_LIBRARY_PATH"] = f"{os.path.join(args.oneapi_base_toolkit, 'compiler', 'linux', 'lib')}:{os.path.join(args.oneapi_base_toolkit, 'compiler', 'linux', 'compiler', 'lib', 'intel64_lin')}:{os.environ.get('LD_LIBRARY_PATH', '')}"
+os.environ["OCL_ICD_FILENAMES"] = f"{os.path.join(args.oneapi_construction_kit_build_dir, 'lib', 'libCL.so')}"
+
+
+def run_cmake_and_ninja(cmake_command, build_dir):
+    try:
+        subprocess.run(cmake_command, check=True)
+        subprocess.run(["ninja", "-C", build_dir], check=True)
+    except subprocess.CalledProcessError as e:
+        print(f"Error executing command: {e}", file=sys.stderr)
```
coldav, line 49: will this show the build output if it fails?

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,187 @@
...
+    build_sycl_dnn()
+    sycl_dnn_build_dir = os.path.join(os.getcwd(), "build_sycl_dnn")
+    vgg_dir = os.path.join(os.getcwd(), "vdata")
+    resnet_dir = os.path.join(os.getcwd(), "rdata")
+
+    # Set the environment variables
+    os.environ["CA_HAL_DEBUG"] =  os.environ.get("CA_HAL_DEBUG", "1")
+    os.environ["CA_PROFILE_LEVEL"] = os.environ.get("CA_PROFILE_LEVEL", "3")
+    os.environ["ONEAPI_DEVICE_SELECTOR"] = "opencl::acc"
+    os.environ["SYCL_CONFIG_FILE_NAME"] = "null.cfg"
```
coldav, line 152: maybe comment on some of this

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,187 @@
...
+    build_portBLAS()
+    build_sycl_dnn()
+    sycl_dnn_build_dir = os.path.join(os.getcwd(), "build_sycl_dnn")
+    vgg_dir = os.path.join(os.getcwd(), "vdata")
+    resnet_dir = os.path.join(os.getcwd(), "rdata")
+
+    # Set the environment variables
+    os.environ["CA_HAL_DEBUG"] =  os.environ.get("CA_HAL_DEBUG", "1")
+    os.environ["CA_PROFILE_LEVEL"] = os.environ.get("CA_PROFILE_LEVEL", "3")
+    os.environ["ONEAPI_DEVICE_SELECTOR"] = "opencl::acc"
```
coldav, line 151: might be worth considering allowing the user to add what they want, we may be one of many accelerators

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,187 @@
...
+def main():
+    build_portBLAS()
+    build_sycl_dnn()
+    sycl_dnn_build_dir = os.path.join(os.getcwd(), "build_sycl_dnn")
+    vgg_dir = os.path.join(os.getcwd(), "vdata")
+    resnet_dir = os.path.join(os.getcwd(), "rdata")
+
+    # Set the environment variables
+    os.environ["CA_HAL_DEBUG"] =  os.environ.get("CA_HAL_DEBUG", "1")
+    os.environ["CA_PROFILE_LEVEL"] = os.environ.get("CA_PROFILE_LEVEL", "3")
```
coldav, line 150: These two are not necessary to run the tests - maybe support a `-e` option?

examples/technical_blogs/tartan_blog/build_and_run_networks.py
hvdijk: Do we need to handle the error and call `sys.exit` here? If we don't catch it, if we let the error go up to `main`, it will be handled there and exit the process the same way, but it also allows callers to possibly handle the error.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,187 @@
...
+import subprocess
+
+parser = argparse.ArgumentParser(description='The purpose of this script is to build and run '
+                                 'VGG16 and Resnet50 networks with oneAPI and oneapi-construction-kit.')
+
+# Prereqs: portBLAS, SYCL-DNN, oneapi oneapi_construction_kit should be cloned and installed.
+parser.add_argument('--portBLAS_source', required=True, help='Path to portBLAS source')
+parser.add_argument('--SYCL_DNN_source', required=True, help='Path to SYCL-DNN source')
+parser.add_argument('--oneapi_construction_kit_build_dir', required=True, help='Path to oneAPI construction kit')
+parser.add_argument('--oneapi_base_toolkit', required=True, help='Path to oneAPI base toolkit')
```
coldav, line 32: I wonder if we should just pass a "path to clang++" so that it can be used with a vanilla clang++ we've built or downloaded.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,187 @@
...
+import shutil
+import requests
+import argparse
+import subprocess
+
+parser = argparse.ArgumentParser(description='The purpose of this script is to build and run '
+                                 'VGG16 and Resnet50 networks with oneAPI and oneapi-construction-kit.')
+
+# Prereqs: portBLAS, SYCL-DNN, oneapi oneapi_construction_kit should be cloned and installed.
+parser.add_argument('--portBLAS_source', required=True, help='Path to portBLAS source')
```
coldav, line 29: A flag to do the cloning for you could be useful?

examples/technical_blogs/tartan_blog/build_and_run_networks.py
```diff
@@ -0,0 +1,234 @@
...
+        print(f"{build_dir} exists,skipping building.")
+    else:
+        try:
+            subprocess.run(cmake_command, check=True)
+            subprocess.run(["ninja", "-C", build_dir], check=True)
+        except subprocess.CalledProcessError as e:
+            print(f"Error executing command: {e}", file=sys.stderr)
+            sys.exit(1)
+
+#Clone repository if not present
```
coldav, line 60:
```suggestion
# Clone repository if not present
```

coldav (approved):

examples/technical_blogs/tartan_blog/build_and_run_networks.py
MaryaSharf: It shows the output already.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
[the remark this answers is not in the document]
MaryaSharf: I can look into in a separate commit.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
MaryaSharf: I can look into in a separate commit.

examples/technical_blogs/tartan_blog/build_and_run_networks.py
MaryaSharf: I can look into in a separate commit.

MaryaSharf:
I'll add the binary support in a later commit as well.

PR #147, opened 2023-10-02
Release any queues left when atexit is called.

coldav:
# Overview

Release any queues left when atexit is called.

# Reason for change

This is due to an issue with dpc++ where it can leave temporary queues unreleased. This can then have the effect on the closing down of the device.

# Description of change

This is resolved by registering and deregistering the queues on creation and deletion and then releasing any queues which still have an external reference when `atexit` is called.

This should be reviewed when https://github.com/intel/llvm/issues/11156 is fixed.

### DIFF
--- a/source/cl/source/command_queue.cpp
+++ b/source/cl/source/command_queue.cpp
@@ -52,6 +52,7 @@ _cl_command_queue::_cl_command_queue(cl_context context, cl_device_id device,
       in_flush(false) {
   cl::retainInternal(context);
   cl::retainInternal(device);
+  { device->RegisterCommandQueue(this); }
 }
 
 _cl_command_queue::~_cl_command_queue() {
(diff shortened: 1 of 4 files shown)


source/cl/source/device.cpp
```diff
@@ -330,6 +331,25 @@ _cl_device_id::_cl_device_id(cl_platform_id platform,
 #endif
 }
 
+void _cl_device_id::RegisterCommandQueue(cl_command_queue queue) {
+  registered_queues.push_back(queue);
+}
+void _cl_device_id::DeregisterCommandQueue(cl_command_queue queue) {
+  registered_queues.remove_if(
+      [queue](cl_command_queue q) { return q == queue; });
```
hvdijk, line 339: `remove_if` is O(n), is it possible to use a different data structure instead? Maybe `std::set`?

hvdijk (approved):

source/cl/source/command_queue.cpp
```diff
@@ -52,6 +52,7 @@ _cl_command_queue::_cl_command_queue(cl_context context, cl_device_id device,
       in_flush(false) {
   cl::retainInternal(context);
   cl::retainInternal(device);
+  { device->RegisterCommandQueue(this); }
```
hvdijk, line 55: The extra braces here are surprising to me, do they have a non-obvious function or are they accidental?

PR #148, opened 2023-10-04
Add instructions to use OCK with official intel base toolkit

MaryaSharf:
# Overview
Adding the documentation to  get the official Intel base toolkit to run and test with OCK.

# Reason for change

Intel official base toolkit can be acquired from the Intel website, and can be used to build with OCK. The instructions are slightyly different from using the pre-release or nightly builds of Intel oneAPI, hence adding the documentation to download, install and test using the official toolkit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,18 +75,89 @@ ninja -C build-riscv install
 ### Cross-compiling oneAPI Construction Kit
 When cross-compiling with CMake, you need to set the `CMAKE_TOOLCHAIN_FILE` variable to tell CMake how to compile for the target architecture. This file sets up various CMake variables, such as the locations of the C and C++ compilers, the assembler, the linker, and the target file system root. By setting these variables correctly, CMake can generate the appropriate build system files for the target platform. More information regarding cross compiling of the oneAPI Construction Kit, can be found [here](doc/developer-guide.md#cross-platform-building-llvm-and-oneapi-construction-kit-for-linux)
 
-### Compiling oneAPI-samples vector-add using DPC++ pre-released compiler
-To obtain the pre-released DPC++ compiler, please visit the following link: https://github.com/intel/llvm/releases. It is recommended to download the DPC++ daily version released on May 18, 2023, which can be found at https://github.com/intel/llvm/releases/tag/sycl-nightly%2F20230518.
+### Compiling oneAPI-samples vector-add using official Intel oneAPI Base Toolkit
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -163,6 +163,74 @@ python buildbot/compile.py -o build
...
+
+Now to compile the vector add from oneAPI samples mentioned in previous section, using the official Intel base toolkit, set the environment variables and follow the stepsgiven below:
+
+```sh
+export OCL_ICD_FILENAMES=$ONEAPI_CON_KIT_INSTALL_DIR/lib/libCL.so
+export LD_LIBRARY_PATH=/path/to/intel/oneapi/compiler/2023.2.0/linux/lib:/path/to/intel/oneapi/compiler/2023.2.0/linux/compiler/lib/intel64_lin:/path/to/intel/oneapi/compiler/2023.2.0/linux/compiler/lib/:$LD_LIBRARY_PATH
+export ONEAPI_DEVICE_SELECTOR="*:acc"
+
+/path/to/intel/oneapi/compiler/2023.2.0/linux/bin-llvm/clang++ -fsycl /path/to/oneAPI-samples/DirectProgramming/C++SYCL/DenseLinearAlgebra/vector-add/src/vector-add-buffers.cpp -o vector-add-buffers
+CA_HAL_DEBUG=1 SYCL_CONFIG_FILE_NAME=
```
coldav, line 197: the does not need null.cfg then?

README.md
```diff
@@ -163,6 +163,74 @@ python buildbot/compile.py -o build
...
+refsi_hal_device::mem_free(address=0x97ffc700)
+refsi_hal_device::mem_free(address=0x98006380)
+[0]: 0 + 0 = 0
+[1]: 1 + 1 = 2
+[2]: 2 + 2 = 4
+...
+[9999]: 9999 + 9999 = 19998
+Vector add successfully completed on device.
+```
+
```
coldav, line 233: I wonder if it is also worth updating the part about the binary version of clang++ which is tied to quite an old version - maybe update to a recent one and say this updates daily.

coldav (approved):

README.md
MaryaSharf: No. I'll update the comment, if mentioned elsewhere. I think someone mentioned that if we add null.cfg here that would mean we need to have a file named null.cgf. Just leaving it empty works same way.

coldav (approved):

PR #153, opened 2023-10-09
[compiler] Flag uses of half types.

hvdijk:
# Overview

When targeting a device that does not support half types, if we nonetheless received input that used half types, we would rely on LLVM's software implementation using library functions from libgcc/compiler-rt which we do not provide, and receive confusing linker errors as a result. For double types, we already flagged any use of such type as an error. This change extends that check to also flag half types, so that we issue an easier-to-understand error message.

# Reason for change

When running the OpenCL CTS proposed change to test half support (https://github.com/KhronosGroup/OpenCL-CTS/pull/1681), we get confusing linker errors when running without half support, complaining about undefined references to __extendhfsf2 and __truncsfhf2.

# Description of change

This change extends the existing check for double support which flags uses of double early when that type is not supported, and makes it so that half is also flagged early when that type is not supported. The existing tests for double are adapted to test half.

N/A

# Checklist

### DIFF
--- a/modules/compiler/source/base/source/check_for_unsupported_types_pass.cpp
+++ b/modules/compiler/source/base/source/check_for_unsupported_types_pass.cpp
@@ -14,7 +14,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-#include <base/check_for_doubles_pass.h>
+#include <base/check_for_unsupported_types_pass.h>
 #include <compiler/utils/device_info.h>
 #include <llvm/IR/DiagnosticPrinter.h>
 #include <llvm/IR/Instructions.h>
@@ -33,41 +33,53 @@ void compiler::DiagnosticInfoDoubleNoDouble::print(DiagnosticPrinter &P) const {
   P << formatMessage();
 }
 
-namespace {
-// Tracks presence of double types in a BasicBlock.
-inline bool findDoubleTypes(const BasicBlock &BB) {
-  // Each instruction in the BB
-  for (const auto &I : BB) {
-    // The instruction's return type
-    if (I.getType()->isDoubleTy()) {
-      return true;
-    }
-    // Else check each operand of the instruction
-    for (const auto &Op : I.operands()) {
-      if (Op->getType()->isDoubleTy()) {
-        return true;
-      }
-    }
-  }
-  return false;
-}
-}  // namespace
+int compiler::DiagnosticInfoHalfNoHalf::DK_HalfNoHalf =
+    getNextAvailablePluginDiagnosticKind();
 
-PreservedAnalyses compiler::CheckForDoublesPass::run(
+StringRef compiler::DiagnosticInfoHalfNoHalf::formatMessage() const {
+  return "A half precision floating point number was generated, "
+         "but cl_khr_fp16 is not supported on this target.";
+}
+
+void compiler::DiagnosticInfoHalfNoHalf::print(DiagnosticPrinter &P) const {
+  P << formatMessage();
+}
+
+PreservedAnalyses compiler::CheckForUnsupportedTypesPass::run(
     Function &F, FunctionAnalysisManager &AM) {
   const auto &MAMProxy = AM.getResult<ModuleAnalysisManagerFunctionProxy>(F);
   const auto *DI =
       MAMProxy.getCachedResult<compiler::utils::DeviceInfoAnalysis>(
           *F.getParent());
-
-  // Check if doubles are supported, in which case there's nothing to do.
-  if (DI && DI->double_capabilities != 0) {
+  // Check if double and half are supported
+  bool CheckDouble = !DI || !DI->double_capabilities;
+  bool CheckHalf = !DI || !DI->half_capabilities;
+  // If we do not have to check for either type, exit right away.
+  if (!CheckDouble && !CheckHalf) {
     return PreservedAnalyses::all();
   }
-  for (const auto &BB : F) {
-    if (findDoubleTypes(BB)) {
+  auto CheckType = [&](llvm::Type *T) {
+    if (T->isDoubleTy() && CheckDouble) {
       F.getContext().diagnose(DiagnosticInfoDoubleNoDouble());
-      return PreservedAnalyses::all();
+      CheckDouble = false;
+    }
+    if (T->isHalfTy() && CheckHalf) {
+      F.getContext().diagnose(DiagnosticInfoHalfNoHalf());
+      CheckHalf = false;
+    }
+  };
+  for (const auto &BB : F) {
+    for (const auto &I : BB) {
+      CheckType(I.getType());
+      if (!CheckDouble && !CheckHalf) {
+        return PreservedAnalyses::all();
+      }
+      for (const auto &Op : I.operands()) {
+        CheckType(Op->getType());
+        if (!CheckDouble && !CheckHalf) {
+          return PreservedAnalyses::all();
+        }
+      }
     }
   }
   return PreservedAnalyses::all();
--- a/modules/compiler/test/lit/passes/check-for-halfs-optnone.ll
+++ b/modules/compiler/test/lit/passes/check-for-halfs-optnone.ll
@@ -17,9 +17,10 @@
 ; RUN: muxc --passes "require<device-info>,check-unsupported-types" -S %s 2>&1
 ; RUN: not muxc --passes check-unsupported-types -S %s 2>&1 | FileCheck %s --check-prefix ERROR
 
-; ERROR: error: A double precision floating point number was generated, but cl_khr_fp64 is not supported on this target.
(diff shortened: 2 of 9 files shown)


frasercrmck:
Hmm yes good idea. I have one question, though; how did this work when `half` is enabled as a storage-only type? Might we not see conversion operations in the IR, which are indeed legal for that configuration?

hvdijk:
> Hmm yes good idea. I have one question, though; how did this work when `half` is enabled as a storage-only type? Might we not see conversion operations in the IR, which are indeed legal for that configuration?

They are legal in LLVM but need runtime support which we do not provide: these `__extendhfsf2` and `__truncsfhf2` functions are the functions used to handle the conversion between half and float. We do not provide them, so we cannot even support `half` as a storage-only type.

frasercrmck:
> > Hmm yes good idea. I have one question, though; how did this work when `half` is enabled as a storage-only type? Might we not see conversion operations in the IR, which are indeed legal for that configuration?
> 
> They are legal in LLVM but need runtime support which we do not provide: these `__extendhfsf2` and `__truncsfhf2` functions are the functions used to handle the conversion between half and float. We do not provide them, so we cannot even support `half` as a storage-only type.

Ah right yes I now see what you're going for, thanks.

I think the diagnostic/commit message does appear as if it's some kind of programmer error or standard violation, rather than a limitation of our current implementation. I assume this is a temporary (or per-target) diagnostic, as we would eventually want to provide these compiler-rt functions via some kind of runtime library.

What would happen if one of our targets didn't have issues with `__extendhfsf2`/`__truncsfhf2`, but another did? Is this change going to limit that first target's use of this pass to check for doubles?

hvdijk:
> I think the diagnostic/commit message does appear as if it's some kind of programmer error or standard violation, rather than a limitation of our current implementation. I assume this is a temporary (or per-target) diagnostic, as we would eventually want to provide these compiler-rt functions via some kind of runtime library.

We might possibly want that in the future, yes. I can imagine targeting RV64GCV but explicitly setting some CMake option to indicate that FP16 should be supported via library calls. And that would indeed involve including these functions from compiler-rt.

> What would happen if one of our targets didn't have issues with __extendhfsf2/__truncsfhf2, but another did? Is this change going to limit that first target's use of this pass to check for doubles?

Yes, if I am understanding you correctly, it does. The pass finds a `DeviceInfo` for a single device, and checks for that device. If that device supports FP16 but not FP64, the pass will still flag any FP64 values in the module.

frasercrmck:
> > I think the diagnostic/commit message does appear as if it's some kind of programmer error or standard violation, rather than a limitation of our current implementation. I assume this is a temporary (or per-target) diagnostic, as we would eventually want to provide these compiler-rt functions via some kind of runtime library.
> 
> We might possibly want that in the future, yes. I can imagine targeting RV64GCV but explicitly setting some CMake option to indicate that FP16 should be supported via library calls. And that would indeed involve including these functions from compiler-rt.
> 
> > What would happen if one of our targets didn't have issues with __extendhfsf2/__truncsfhf2, but another did? Is this change going to limit that first target's use of this pass to check for doubles?
> 
> Yes, if I am understanding you correctly, it does. The pass finds a `DeviceInfo` for a single device, and checks for that device. If that device supports FP16 but not FP64, the pass will still flag any FP64 values in the module.

Ah sorry I meant "limit" as in "restrict", but I understand what you mean. So, I had it in my head that we could have a device that supports storage-only half (e.g., through compiler-rt) but the `DeviceInfo` says it doesn't support fp16. In that case, the target would be restricted from using the pass to check for doubles in a module that has storage-only half types.

I suppose it depends on what we mean by the device "supporting" fp16. I think I've been assuming it means "full hardware support" but I can see your interpretation, in which case your change makes sense to me. With fp64 I think it's pretty clear-cut, but the concept of storage-only fp16 makes the `bool` a bit ambiguous to me. These capabilities were taken into the compiler's `DeviceInfo` using the `mux` equivalents. I think we've got an internal ticket somewhere about the ambiguity/correctness of using capabilities bitfields as a bool.

hvdijk:
Ah, that makes sense. Either way, though, the checks in this pass need to be aligned with when the extension is reported as supported, which looks to be done based on [`!device_info->half_capabilities`](https://github.com/codeplaysoftware/oneapi-construction-kit/blob/7372f796dfe3b279278a47ad1f7f17ded870af38/modules/compiler/source/base/source/module.cpp#L971), so the extension is counted as supported if any bits are set. This pass is doing the same, which I think is right?

frasercrmck:
> Ah, that makes sense. Either way, though, the checks in this pass need to be aligned with when the extension is reported as supported, which looks to be done based on [`!device_info->half_capabilities`](https://github.com/codeplaysoftware/oneapi-construction-kit/blob/7372f796dfe3b279278a47ad1f7f17ded870af38/modules/compiler/source/base/source/module.cpp#L971), so the extension is counted as supported if any bits are set. This pass is doing the same, which I think is right?

We chatted about this offline. When `device_info->half_capabilities != 0` the `cl_khr_fp16` extension is enabled and the compiler treats that as `half` types being "legal".

When `device_info->half_capabilities == 0`, we still allow `half`s in storage-only mode via buffers (much like SPIR-V's `Float16Buffer` capability) but those buffers cannot be dereferenced except by special instructions/builtins, which do not take/produce `half` values.

So I'm convinced this change is sound as it diagnoses only `half` value types. Thanks!

modules/compiler/test/lit/passes/check-for-halfs-optnone.ll
```diff
@@ -0,0 +1,27 @@
...
+; RUN: muxc --passes "require<device-info>,check-unsupported-types" -S %s 2>&1
+; RUN: not muxc --passes check-unsupported-types -S %s 2>&1 | FileCheck %s --check-prefix ERROR
+
+; ERROR: error: A half precision floating point number was generated, but cl_khr_fp16 is not supported on this target.
+
+; Check that optnone functions aren't skipped
+define void @foo() optnone noinline {
+  %a = fadd half 0.0, 1.0
+  ret void
+
```
frasercrmck, line 26: Unnecessary blank line. I see this was copied from the original, so feel free to delete it there too if you like - I'm easy.

modules/compiler/test/lit/passes/check-for-halfs.ll
```diff
@@ -0,0 +1,26 @@
...
+
+; RUN: muxc --passes "require<device-info>,check-unsupported-types" -S %s 2>&1
+; RUN: not muxc --passes check-unsupported-types -S %s 2>&1 | FileCheck %s --check-prefix ERROR
+
+; ERROR: error: A half precision floating point number was generated, but cl_khr_fp16 is not supported on this target.
+
+define void @foo() {
+  %a = fadd half 0.0, 1.0
+  ret void
+
```
frasercrmck, line 25: Unnecessary blank line

frasercrmck (commented):
Just nits, looks good otherwise.

modules/compiler/test/lit/passes/check-for-halfs-optnone.ll
hvdijk: It was indeed copied, removed here and in the original.

modules/compiler/source/base/source/check_for_unsupported_types_pass.cpp
```diff
@@ -33,41 +33,53 @@ void compiler::DiagnosticInfoDoubleNoDouble::print(DiagnosticPrinter &P) const {
...
+  for (const auto &BB : F) {
+    for (const auto &I : BB) {
+      CheckType(I.getType());
+      if (!CheckDouble && !CheckHalf) {
+        return PreservedAnalyses::all();
+      }
+      for (const auto &Op : I.operands()) {
+        CheckType(Op->getType());
+        if (!CheckDouble && !CheckHalf) {
+          return PreservedAnalyses::all();
```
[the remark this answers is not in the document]
hvdijk: `CheckType` is a lambda which can clear `CheckDouble` and `CheckHalf`. The idea is that once the pass has warned about `double` once, and about `half` once, there is no need to check the rest of the module.

modules/compiler/source/base/source/check_for_unsupported_types_pass.cpp
```diff
@@ -33,41 +33,53 @@ void compiler::DiagnosticInfoDoubleNoDouble::print(DiagnosticPrinter &P) const {
...
+    if (T->isHalfTy() && CheckHalf) {
+      F.getContext().diagnose(DiagnosticInfoHalfNoHalf());
+      CheckHalf = false;
+    }
+  };
+  for (const auto &BB : F) {
+    for (const auto &I : BB) {
+      CheckType(I.getType());
+      if (!CheckDouble && !CheckHalf) {
+        return PreservedAnalyses::all();
```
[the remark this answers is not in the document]
hvdijk: Same as the other one.

PR #158, opened 2023-10-11
[compiler] Ensure createLoop always creates a loop

frasercrmck:
This simplifies the `createLoop` API so that users can rely on it having
generated a loop CFG structure. It correspondingly removes the
`allowUnroll` option.

Now, even when the loop bounds are constant and wouldn't generate a loop
(i.e., 0 or 1 iterations), `createLoop` will generate a loop. Users can
rely on standard LLVM passes to eliminate/simplify the loop.


frasercrmck:
There are some performance problems with this one, which I need to look into.

coldav (approved):

frasercrmck:
> There are some performance problems with this one, which I need to look into.

These problems seem to largely disappear on x86-64-v3+, and there are benefits too. So I may merge this one #291 is merged.

PR #163, opened 2023-10-17
[host] Fix move-after-move errors

frasercrmck:
The use of `llvm::consumeError` is just to mark the error as 'handled' - it doesn't actually report the error in any way. If there's a callback we have reported *and* handled the error, so using `llvm::consumeError` is more correctly placed in an 'else' block.

This fixes potential move-after-move errors when a compilation error happens and the callback doesn't halt execution at that point.

### DIFF
--- a/modules/compiler/targets/host/source/kernel.cpp
+++ b/modules/compiler/targets/host/source/kernel.cpp
@@ -334,8 +334,9 @@ HostKernel::lookupOrCreateOptimizedKernel(std::array<size_t, 3> local_size) {
       if (auto callback = target.getNotifyCallbackFn()) {
         callback(llvm::toString(std::move(err)).c_str(), /*data*/ nullptr,
                  /*data_size*/ 0);
+      } else {
+        llvm::consumeError(std::move(err));
       }
-      llvm::consumeError(std::move(err));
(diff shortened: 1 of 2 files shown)


coldav (approved):

coldav (approved):

hvdijk:
> If there's a callback we have reported and handled the error

It looks like we only pass a string to the callback. How do we handle the error?

frasercrmck:
> > If there's a callback we have reported and handled the error
> 
> It looks like we only pass a string to the callback. How do we handle the error?

Ah yeah, you're right - here I was meaning we 'handle' the error in the sense that by moving/converting it to a string, LLVM considers the error checked and won't complain at us. We don't 'handle' it in the other sense that the callback does anything other than report it; we 'handle' the error in that sense by returning `compiler::Result::FAILURE` and bubbling the failure upwards.

I can try and clarify this language if you like.

hvdijk:
> Ah yeah, you're right - here I was meaning we 'handle' the error in the sense that by moving/converting it to a string, LLVM considers the error checked and won't complain at us.

Ah, I didn't realise this is sufficient to have LLVM consider the error handled. In that case the PR looks fine to me.

hvdijk (approved):

PR #169, opened 2023-10-18
[host] Initialize unique_identifier.

hvdijk:
This has no directly visible impact since we do not care what unique_identifier's initial value is, we just use it to generate unique names, but valgrind correctly picks up on the fact that we use it without initializing it.

# Overview

This change provides an initial value to `unique_identifier`.

# Reason for change

valgrind reports that we use `unique_identifier` uninitialized.

# Description of change

We use `unique_identifier` only in `HostKernel::lookupOrCreateOptimizedKernel` as follows:

```c++
      if (snprintf(unique_name_data, unique_name_data_length,
                   "__mux_host_%" PRIu64, target.unique_identifier++) < 0) {
```

We do not care what exactly we get as the ID, we just need it to be something we have not seen before. Any initial value will do, so the uninitialized use is mostly harmless and this change should have no visible effect other than to silence valgrind.

N/A

# Checklist

### DIFF
--- a/modules/compiler/targets/host/include/host/target.h
+++ b/modules/compiler/targets/host/include/host/target.h
@@ -73,7 +73,7 @@ class HostTarget : public compiler::BaseTarget {
   /// execution engine has a unique name. This identifier will be suffixed onto
   /// the kernel names, and incremented under the context's mutex lock such that
   /// no conflict should occur.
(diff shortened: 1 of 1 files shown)


frasercrmck (approved):
LGTM

coldav (approved):

PR #170, opened 2023-10-19
[passes] Handle alignments of structs correctly with local memory

RossBrunton:
# Overview

This pass changes accesses to local memory (in the form of globals in address space 3) to offsets into a single struct passed into the function.

However, when calculating the layout of this struct, it incorrectly assumed that the size of a type was equal to its alignment (with some special handling for arrays). This patch changes it so that the datalayout is queried directly for the size of the type.

# Reason for change

Was causing a SYCL-CTS test to segfault when running hierarchical code.

# Description of change

Fixed a bug in the struct layout calculation.

Nothing.

# Checklist

### DIFF
--- a/modules/compiler/utils/source/replace_local_module_scope_variables_pass.cpp
+++ b/modules/compiler/utils/source/replace_local_module_scope_variables_pass.cpp
@@ -194,6 +194,7 @@ PreservedAnalyses compiler::utils::ReplaceLocalModuleScopeVariablesPass::run(
   unsigned int maxAlignment = 0;
   // byte offset in struct of current member
   unsigned int offset = 0;
+  const auto &dl = M.getDataLayout();
   for (auto &global : globals) {
     auto memberType = global->getValueType();
 
@@ -202,8 +203,7 @@ PreservedAnalyses compiler::utils::ReplaceLocalModuleScopeVariablesPass::run(
     // global. This is also needed if '__attribute__(aligned)' was used to
     // set a specific alignment.
     const unsigned int alignment =
-        std::max(global->getAlignment(),
-                 calculateTypeAlign(memberType, M.getDataLayout()));
+        std::max(global->getAlignment(), calculateTypeAlign(memberType, dl));
     assert(alignment > 0 && "'0' is an impossible alignment");
 
     // check if this is the largest alignment seen so far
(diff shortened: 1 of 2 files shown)


modules/compiler/utils/source/replace_local_module_scope_variables_pass.cpp
```diff
@@ -236,12 +236,9 @@ PreservedAnalyses compiler::utils::ReplaceLocalModuleScopeVariablesPass::run(
...
 
-    // update the offset, taking into account any array length
-    unsigned int totalSize = alignment;
-    while (memberType->isArrayTy()) {
-      totalSize *= memberType->getArrayNumElements();
-      memberType = memberType->getArrayElementType();
-    }
+    // update the offset based on the type's size
+    unsigned int totalSize =
+        M.getDataLayout().getTypeAllocSize(memberType).getFixedValue();
```
frasercrmck, line 241: At this point, it might be worth capturing `M.getDataLayout()` somewhere before the loop. We also use it at line 206.

Also while we're here, we could maybe also add a more human-friendly assertion that the type alloc size is indeed not a scalable amount before grabbing `getFixedValue`?

hvdijk, line 241: This should possibly be [`getTypeStoreSize`](https://llvm.org/doxygen/classllvm_1_1DataLayout.html#acfcd22eb38dbfe1acbf138754297437a) instead? `getTypeAllocSize` includes the padding that would be required if we were to add another instance of this type right after.

RossBrunton: I think getTypeAllocSize is correct here.

Doing some testing, a struct `{i20, i20}` is 8 bytes wide, so an individual `i20` entry in that struct must be 4 bytes, which includes the 12 bits of padding. That is the value that getTypeAllocSize returns (getTypeStoreSize only returns 3 bytes).

modules/compiler/utils/source/replace_local_module_scope_variables_pass.cpp
```diff
@@ -236,12 +236,13 @@ PreservedAnalyses compiler::utils::ReplaceLocalModuleScopeVariablesPass::run(
...
 
-    // update the offset, taking into account any array length
-    unsigned int totalSize = alignment;
-    while (memberType->isArrayTy()) {
-      totalSize *= memberType->getArrayNumElements();
-      memberType = memberType->getArrayElementType();
+    // update the offset based on the type's size
+    if (dl.getTypeAllocSize(memberType).isScalable()) {
+      // Not an assert because this can happen in user-supplied IR
+      llvm::report_fatal_error(
```
frasercrmck, line 242: nit: you don't need `llvm::`

frasercrmck (approved):
LGTM with a nit

modules/compiler/utils/source/replace_local_module_scope_variables_pass.cpp
```diff
@@ -236,12 +236,13 @@ PreservedAnalyses compiler::utils::ReplaceLocalModuleScopeVariablesPass::run(
...
+          "Scalable types in local memory are not yet supported");
     }
+    unsigned int totalSize = dl.getTypeAllocSize(memberType).getFixedValue();
```
hvdijk, line 245: I don't know if `dl.getTypeAllocSize()` caches its results so that for consecutive calls, it does not do the same work of going over struct members repeatedly. If it doesn't do that, then this looks good to me. If it does, it might be better to store the result in a local variable. Will tentatively approve and leave it to you to determine whether this is an issue to change; looks good to me otherwise.

hvdijk (approved):

hvdijk (approved):

PR #171, opened 2023-10-20
[test] Review use of kts::N and kts::localN.

hvdijk:
# Overview

* C11Atomics.cpp: fix local buffers. We called AddLocalBuffer(kts::N) where kts::N is 256, where we need local_work_size * sizeof(T) which is 64 * 4, which is also 256. This works out but is fragile and breaks if we get larger T, and disregards kts::localN.
* ktst_compiler_barrier.cpp, ktst_regression_02.cpp: add explicit assertions so we fail hard rather than corrupting memory if we change kts::N or kts::localN.
* ktst_regression_01.cpp: change ASSERT_TRUE to ASSERT_EQ to get a better error message if the assertion fails.

# Reason for change

Planned work regarding atomics resulted in test failures due to incorrect use of kts::N. This is split out into a separate MR for easier review.

# Description of change

This change only affects our own tests. Other people see no changes.

N/A

# Checklist

### DIFF
--- a/source/cl/test/UnitCL/source/C11Atomics.cpp
+++ b/source/cl/test/UnitCL/source/C11Atomics.cpp
@@ -55,15 +55,15 @@ class InitTest : public C11AtomicTestBase {
       return input_data[index];
     };
 
-    // Set up the buffers.
+    // Set up the buffers and run the test.
     AddInputBuffer(kts::N, random_reference);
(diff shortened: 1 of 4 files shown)


frasercrmck (approved):
LGTM, thanks. This is definitely an overly fiddly part of UnitCL

PR #172, opened 2023-10-23
[compiler] Implement 64-bit atomics.

hvdijk:
# Overview

This enables 64-bit atomic operations on devices that support it.

# Reason for change

The primary reason for the change was not to get 64-bit atomics working in OpenCL, it was to get it working in SYCL. But DPC++'s SYCL implementation queries the available OpenCL extensions to determine whether to enable 64-bit atomic support.

# Description of change

* The mux device capabilities provide a way to indicate which widths are supported for atomic operations. Currently, atomic operations of 8, 16, or 32 bits are always presumed supported. Atomic operations of 64 bits are presumed supported on targets with 64-bit pointers.
* The OpenCL extensions cl_khr_int64_base_atomics and cl_khr_int64_extended_atomics are implemented on targets that support 64-bit atomic operations. Note that despite the name, in accordance with the OpenCL specification, these extensions also cover atomic operations on type 'double' when that type is supported.

The UnitCL tests are a copy of the 32-bit tests, adapted for 64-bits.

This causes test failures in SYCL-CTS that are unrelated to handling of atomics, because the tests that check atomic operations on pointers were previously skipped, and now fail, but the failure is in passing pointers from host to device, not the atomic operations. These are planned to be addressed in a future change.

# Checklist


modules/mux/targets/riscv/source/device_info.cpp
```diff
@@ -70,6 +70,9 @@ device_info_s::device_info_s()
 
   // Override this default from hal updates
   this->address_capabilities = mux_address_capabilities_bits64;
+  this->atomic_capabilities =
+      mux_atomic_capabilities_8bit | mux_atomic_capabilities_16bit |
+      mux_atomic_capabilities_32bit | mux_atomic_capabilities_64bit;
```
coldav, line 75: I think we only support 64 bit atomics if it's a a 64 bit risc-v? (Technically also if we support atomics)

source/cl/test/UnitCL/include/Common.h
```diff
@@ -311,6 +311,11 @@ bool hasHalfSupport(cl_device_id device);
 /// @return True if the device has double support, false otherwise
 bool hasDoubleSupport(cl_device_id device);
 
+/// @brief Check if the device at index 'device' has double support
```
coldav, line 314:
```suggestion
/// @brief Check if the device at index 'device' has Atomic64 support
```

modules/mux/targets/riscv/source/device_info.cpp
hvdijk: This is an existing problem in `device_info_s` that I just left as is: we also set `integer_capabilities` to include `mux_integer_capabilities_64bit` even on 32-bit RISC-V where it doesn't work. It should be aligned with that, I think: on RISC-V, we can support 64-bit atomics in exactly the same cases that we can 64-bit integers in general.

source/cl/test/UnitCL/include/Common.h
```diff
@@ -311,6 +311,11 @@ bool hasHalfSupport(cl_device_id device);
...
+/// @param[in] device - The index of the device in getDevices()
+/// @return True if the device has double support, false otherwise
```
coldav, line 316:
```suggestion
/// @return True if the device has double support, false otherwise
```
```suggestion
/// @return True if the device has Atomic64 support, false otherwise
```

source/cl/test/UnitCL/include/Common.h
hvdijk: Thanks, will update and go over the other comments too.

modules/mux/targets/host/source/device.cpp
```diff
@@ -360,10 +360,16 @@ device_info_s::device_info_s(host::arch arch, host::os os, bool native,
     case host::arch::ARM:
     case host::arch::X86:
       this->address_capabilities |= mux_address_capabilities_bits32;
+      this->atomic_capabilities |= mux_atomic_capabilities_8bit |
```
coldav, line 363: I can't see this initialized anywhere, but you are using |=

modules/mux/targets/riscv/source/device_info.cpp
coldav: I think atomics is a bit different though, we'd generally expect 64 bit atomics on a 64 bit (address space) risc-v. If you tried this code on a 32 bit version, we'll get additional atomic link issues most likely.

modules/mux/targets/host/source/device.cpp
hvdijk: You're right, this should have been `=`, fixed and likewise for below.

modules/mux/targets/riscv/source/device_info.cpp
hvdijk: Per offline discussion, even though 32-bit RISC-V support is currently broken when it comes to 64-bit types, we do want to support that and this PR should be so that it does not make things worse there. Will update.

modules/compiler/utils/source/replace_atomic_funcs_pass.cpp
```diff
@@ -225,20 +363,21 @@ bool RunOnInstruction(CallInst &call) {
       auto op1 =
           (1 == call.arg_size()) ? builder.getInt32(1) : call.getArgOperand(1);
 
-      if (call.getType()->isFloatTy()) {
-        auto ptr = builder.getInt32Ty()->getPointerTo(
-            op0->getType()->getPointerAddressSpace());
+      if (call.getType()->isFloatingPointTy()) {
+        auto ptr = builder.getIntNTy(call.getType()->getPrimitiveSizeInBits())
```
frasercrmck, line 367: This isn't strictly necessary as we don't support typed pointers, but I don't mind it either way. This is arguably more versatile.

frasercrmck (approved):
LGTM

modules/compiler/utils/source/replace_atomic_funcs_pass.cpp
hvdijk: True, in opaque pointer mode (the only mode still in use) this just creates a bitcast from `ptr` to `ptr` (except in the right address space). We already did so for `float`, this change makes it so that `double` behaves the same as `float`, but we could take this out for both `float` and `double` now that it is no longer needed. We have tickets to remove specific bits of leftover support for older LLVM versions, I'll create a similar ticket to go over the remaining typed pointer support and take it out.

coldav (approved):

PR #173, opened 2023-10-23
Fix for spir-v translation to IR for OpPtrAccessChain

coldav:
# Overview

Fix for spir-v translation to IR for OpPtrAccessChain

# Reason for change

In some cases the indexes field for OpPtrAccessChain can be empty. The translation code currently assumes this not to be the case and checks for element decorations, resulting in a segmentation fault during SYCL CTS testing.

# Description of change

This fixes the empty index case by not checking the decorations in this case. There are some issues with OpMemberDecorate, so testing of the path with additional arguments requiring decorations will not be tested here.

Also some previous clang-format issues have been resolved.

Additional lit tests have been added.

### DIFF
--- a/modules/compiler/spirv-ll/test/spvasm/op_ptr_access_chain.spvasm
+++ b/modules/compiler/spirv-ll/test/spvasm/op_ptr_access_chain.spvasm
@@ -1,0 +1,67 @@
+; Copyright (C) Codeplay Software Limited
+;
+; Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+; Exceptions; you may not use this file except in compliance with the License.
+; You may obtain a copy of the License at
+;
+;     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+;
+; Unless required by applicable law or agreed to in writing, software
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; RUN: %if online-spirv-as %{ spirv-as --target-env %spv_tgt_env -o %spv_file_s %s %}
+; RUN: spirv-ll-tool -a OpenCL -b 64 %spv_file_s | FileCheck %s
+
+    OpCapability Addresses
+    OpCapability Kernel
+    OpCapability Int64
+    OpMemoryModel Physical64 OpenCL
+    OpEntryPoint Kernel %20 "entry_pt"
+
+    %void = OpTypeVoid
+    %ulong = OpTypeInt 64 0
+    %_ptr_Generic_ulong = OpTypePointer Generic %ulong
+
+    %ulong_0 = OpConstant %ulong 0
+    %ulong_3 = OpConstant %ulong 3
+    %float = OpTypeFloat 32
+    %uchar = OpTypeInt 8 0
+    %uint = OpTypeInt 32 0
+    %_ptr_uint = OpTypePointer CrossWorkgroup %uint
+    %ulong_3 = OpConstant %ulong 3
+
+    %struct_user_struct = OpTypeStruct %float %uint %uchar %_ptr_uint
+    %_ptr_CrossWorkgroup_struct_user_struct = OpTypePointer CrossWorkgroup %struct_user_struct
+
+    %19 = OpTypeFunction %void %_ptr_CrossWorkgroup_struct_user_struct %ulong  %ulong
+    %20 = OpFunction %void None %19
+    %_arg_user_struct = OpFunctionParameter %_ptr_CrossWorkgroup_struct_user_struct
+    %_ulong_input = OpFunctionParameter %ulong
+    %_ulong_input2 = OpFunctionParameter %ulong
+   
+; CHECK: define spir_kernel void @entry_pt(ptr addrspace(1) [[PTR:%.*]], i64 {{%[a-zA-Z0-9]+}}, i64 {{%[a-zA-Z0-9]+}}
+    %entry = OpLabel
+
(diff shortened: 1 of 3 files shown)


modules/compiler/spirv-ll/test/spvasm/op_ptr_access_chain.spvasm
```diff
@@ -0,0 +1,67 @@
...
+    %struct_user_struct = OpTypeStruct %float %uint %uchar %_ptr_uint
+    %_ptr_CrossWorkgroup_struct_user_struct = OpTypePointer CrossWorkgroup %struct_user_struct
+
+    %19 = OpTypeFunction %void %_ptr_CrossWorkgroup_struct_user_struct %ulong  %ulong
+    %20 = OpFunction %void None %19
+    %_arg_user_struct = OpFunctionParameter %_ptr_CrossWorkgroup_struct_user_struct
+    %_ulong_input = OpFunctionParameter %ulong
+    %_ulong_input2 = OpFunctionParameter %ulong
+   
+; CHECK: define spir_kernel void @entry_pt(ptr addrspace(1) %[[PTR:.*]], i64 {{%[a-zA-Z0-9]+}}, i64 {{%[a-zA-Z0-9]+}}
```
frasercrmck, line 47: I don't think mixing regex styles between `%.*` and `%[a-zA-Z0-9]+` makes sense. Best just stick to `%.*` across the board, I think. LLVM's own scripts just use `%.*` so that's what I've been trying to slowly transition us to.

frasercrmck: These LLVM scripts also capture the `%` in the variable itself (e.g., `[[PTR:%.*]]`) and so that's the style I've been favouring.

modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -3010,7 +3010,9 @@ cargo::optional<Error> Builder::create<OpPtrAccessChain>(
       basePointeeTy, base, indexes, module.getName(op->IdResult()),
       IRBuilder.GetInsertBlock());
 
-  if (basePointeeTy->isStructTy()) {
+  // Check member decorations if the original indexes field is not empty
+  // as in this case we are accessing an element
+  if (basePointeeTy->isStructTy() && indexes.size() > 1) {
```
frasercrmck, line 3015: Maybe I'm missing something to do with the meaning of "original", but this condition isn't "not empty" - it's "has more than one index".

I'd maybe recommend `!indexes.empty()`. `empty` is guaranteed to be constant time operation, but `size` does not have that requirement.

frasercrmck:
I think styling `spir-v` as `SPIR-V` in the commit message/description would be more correct

modules/compiler/spirv-ll/source/builder_core.cpp
frasercrmck: Oh sorry I believe as of C++11, `size` has that constant-time requirement too.

modules/compiler/spirv-ll/test/spvasm/op_ptr_access_chain.spvasm
coldav: there was a bit of a mix in the ones I looked at, I meant to be consistent, I can fix it up

modules/compiler/spirv-ll/source/builder_core.cpp
coldav: I can't use empty() as the above code adds the element in, so it will always be at least one. I'll rephrase to make it clearer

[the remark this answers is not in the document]
coldav: I've tried to clarify, there is confusion between the indexes vector and the input.

modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -3010,7 +3010,11 @@ cargo::optional<Error> Builder::create<OpPtrAccessChain>(
       basePointeeTy, base, indexes, module.getName(op->IdResult()),
       IRBuilder.GetInsertBlock());
 
-  if (basePointeeTy->isStructTy()) {
+  // Check member decorations if the input indexes field is not empty
+  // as in this case we are accessing an element.
+  // We do this by checking the size of indexes vector is greater than one as
+  // the element field is always added at the start.
+  if (basePointeeTy->isStructTy() && indexes.size() > 1) {
     checkMemberDecorations(basePointeeTy, indexes, op->IdResult());
```
frasercrmck, line 3018: If we're going to enforce that `checkMemberDecorations` should only be called when `indexes.size() > 1` then it makes sense to enforce that at all call sites, and `assert` in `checkMemberDecorations` itself. Because this can happen also in `OpInBoundsAccessChain` and others.

So maybe just make `checkMemberDecorations` early exit if there aren't enough indices, and then we fix all 4 calls to it at once?

coldav: That's a fair point, I've done that.

modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -3010,9 +3010,7 @@ cargo::optional<Error> Builder::create<OpPtrAccessChain>(
       basePointeeTy, base, indexes, module.getName(op->IdResult()),
       IRBuilder.GetInsertBlock());
 
-  if (basePointeeTy->isStructTy()) {
-    checkMemberDecorations(basePointeeTy, indexes, op->IdResult());
-  }
+  checkMemberDecorations(basePointeeTy, indexes, op->IdResult());
```
frasercrmck, line 3013: Is it valid to call `checkMemberDecorations` when the type isn't a struct type? This seems like a change in behaviour.

coldav: I've reverted now

frasercrmck (approved):
Looks good. It might be worth adding LIT tests for the other nodes with optional index lists that call `checkMemberDecorations`, but that could also come later.

PR #174, opened 2023-10-24
[compiler] Represent samplers/events internally as size_t

frasercrmck:
This fixes a bug in LLVM 17 where, on 64-bit targets, the `target` types used to represent samplers and events have an implicit size/alignment of 8 bytes (as their underlying "layout type" is `ptr`). This means that it is generally invalid to replace them with `i32`, as this may break existing alignment/size assumptions that the rest of the module has made on the access of the 'old' types.

Instead, we choose to replace `target` types with `size_t` (in practice `i32` or `i64` depending on the SPIR triple). This works more or less the same way on 32-bit targets.

On 64-bit targets, however, the runtime passes samplers to kernels as 8-byte values. These values follow the existing munging via pointers on LLVM 16, but are eventually zero-extended or truncated to `i32` when they interface with existing APIs that expect `i32` - such as `libimg`.

### DIFF
--- a/modules/mux/targets/host/source/command_buffer.cpp
+++ b/modules/mux/targets/host/source/command_buffer.cpp
@@ -129,10 +129,10 @@ void populatePackedArgs(
             break;
         }
 
-        uint32_t sampler_ptr = libimg::HostCreateSampler(
+        uint64_t sampler_ptr = libimg::HostCreateSampler(
             info.sampler.normalize_coords, addressing_mode, filter_mode);
-        std::memcpy(packed_args_alloc + offset, &sampler_ptr, sizeof(uint32_t));
-        offset += sizeof(uint32_t);
+        std::memcpy(packed_args_alloc + offset, &sampler_ptr, sizeof(size_t));
+        offset += sizeof(size_t);
 #endif
       } break;
       case mux_descriptor_info_type_plain_old_data: {
--- a/examples/refsi/refsi_m1/compiler/refsi_m1/source/refsi_mux_builtin_info.cpp
+++ b/examples/refsi/refsi_m1/compiler/refsi_m1/source/refsi_mux_builtin_info.cpp
@@ -15,6 +15,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include <compiler/utils/dma.h>
+#include <compiler/utils/pass_functions.h>
 #include <llvm/ADT/StringSwitch.h>
 #include <llvm/IR/Operator.h>
 #include <multi_llvm/llvm_version.h>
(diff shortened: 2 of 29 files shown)


examples/refsi/refsi_m1/compiler/refsi_m1/test/passes/refsi_define_mux_dma_pass_read_1d_tgt_tys.ll
```diff
@@ -17,8 +17,8 @@
 ; REQUIRES: llvm-17+
 ; RUN: muxc --device "%riscv_device" %s --passes replace-target-ext-tys,define-mux-dma,verify -S | FileCheck %s
 
-target datalayout = "e-m:e-p:64:64-i64:64-i128:128-n64-S128"
-target triple = "riscv64-unknown-unknown-elf"
+target datalayout = "e-m:e-p:32:32-i64:64-i128:128-n64-S128"
```
coldav, line 20: This seems to be perturbing M1 into a 32 bit architecture - when it's not really run like that. I get what you are doing here, but maybe some comments to explain that we are tweaking it subtly for the lit test (and the title should reflect 32)

modules/mux/targets/host/source/command_buffer.cpp
```diff
@@ -129,10 +129,10 @@ void populatePackedArgs(
             break;
         }
 
-        uint32_t sampler_ptr = libimg::HostCreateSampler(
+        uint64_t sampler_ptr = libimg::HostCreateSampler(
             info.sampler.normalize_coords, addressing_mode, filter_mode);
-        std::memcpy(packed_args_alloc + offset, &sampler_ptr, sizeof(uint32_t));
-        offset += sizeof(uint32_t);
+        std::memcpy(packed_args_alloc + offset, &sampler_ptr, sizeof(void *));
+        offset += sizeof(void *);
```
coldav, line 135: since we tend towards size_t elsewhere - maybe we should use that here?

coldav (approved):

coldav (approved):

modules/mux/targets/host/source/command_buffer.cpp
frasercrmck: Yeah, I was copying some lines from further up which were `void *`. But `void *` for pointers and `size_t` for values makes sense to me.

examples/refsi/refsi_m1/compiler/refsi_m1/test/passes/refsi_define_mux_dma_pass_read_1d_tgt_tys.ll
frasercrmck: Good point. Done

PR #176, opened 2023-10-25
Handle ConstantArray in WorkItemLoopPass

PietroGhg:
# Overview

Handle case where the kernel function is used by an LLVM ConstantArray.

# Reason for change

An llvm_unreachable was being hit.

# Description of change

Handle similarly to `llvm::ConstantExpression`.

Added test `modules/compiler/test/lit/passes/work-item-loop-array.ll `

# Checklist

### DIFF
--- a/modules/compiler/utils/include/compiler/utils/pass_functions.h
+++ b/modules/compiler/utils/include/compiler/utils/pass_functions.h
@@ -23,6 +23,7 @@
 
 #include <cargo/optional.h>
 #include <llvm/ADT/Twine.h>
+#include <llvm/IR/Constants.h>
 #include <llvm/IR/Function.h>
 #include <llvm/Transforms/Utils/ValueMapper.h>
 
@@ -74,6 +75,18 @@ void replaceConstantExpressionWithInstruction(llvm::Constant *const constant);
 /// @param[in] to Constant which will replace any operands which are `from`
 void remapConstantExpr(llvm::ConstantExpr *expr, llvm::Constant *from,
                        llvm::Constant *to);
+
+/// @brief remap operands of a constant array
+///
+/// @note This will create a new constant array and replace references to
+/// the original constant with the new one
+///
+/// @param[in] arr Constant array to be remapped
+/// @param[in] from Constant which if found in array will be
+/// replaced
+/// @param[in] to Constant which will replace any operands which are `from`
+void remapConstantArray(llvm::ConstantArray *arr, llvm::Constant *from,
+                        llvm::Constant *to);
 
 /// @brief Discover if input function references debug info metadata nodes
 ///
--- a/modules/compiler/utils/source/pass_functions.cpp
+++ b/modules/compiler/utils/source/pass_functions.cpp
@@ -73,20 +73,35 @@ uint64_t computeApproximatePrivateMemoryUsage(const llvm::Function &fn) {
   return bytes;
 }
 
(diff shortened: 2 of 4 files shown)


modules/compiler/utils/source/pass_functions.cpp
```diff
@@ -73,20 +73,36 @@ uint64_t computeApproximatePrivateMemoryUsage(const llvm::Function &fn) {
   return bytes;
 }
 
-void remapConstantExpr(llvm::ConstantExpr *expr, llvm::Constant *from,
-                       llvm::Constant *to) {
+template <typename ConstantT>
+static llvm::SmallVector<llvm::Constant *, 8> getNewOps(ConstantT *constant,
```
frasercrmck, line 77: Perhaps this would be better as returning `llvm::SmallVectorImpl<llvm::Constant *>` to avoid the API and consumers assuming a specific size?

modules/compiler/utils/include/compiler/utils/pass_functions.h
```diff
@@ -75,6 +76,8 @@ void replaceConstantExpressionWithInstruction(llvm::Constant *const constant);
 void remapConstantExpr(llvm::ConstantExpr *expr, llvm::Constant *from,
                        llvm::Constant *to);
 
+void remapConstantArray(llvm::ConstantArray *expr, llvm::Constant *from,
```
frasercrmck, line 79: We should probably copy/paste the doxygen from `remapConstantExpr`

frasercrmck (commented):
Now I'm a bit worried about `ConstantStruct`, `ConstantVector`, etc :D

Not that I think you need to fix that in this PR - I might create a ticket to investigate myself

frasercrmck (approved):
Looks good, thanks!

modules/compiler/utils/source/pass_functions.cpp
```diff
@@ -73,20 +73,36 @@ uint64_t computeApproximatePrivateMemoryUsage(const llvm::Function &fn) {
   return bytes;
 }
 
-void remapConstantExpr(llvm::ConstantExpr *expr, llvm::Constant *from,
-                       llvm::Constant *to) {
-  llvm::SmallVector<llvm::Constant *, 8> newOps;
-  // iterate through the constant expression and create a vector of old and new
+template <typename ConstantT>
+static llvm::SmallVector<llvm::Constant *> getNewOps(ConstantT *constant,
```
coldav, line 77: Does this really need to be a template - would it not just work with Constant *constant as the first arg?

PietroGhg: `Constant::getOperand` returns a `Value`, `ConstantExpression::getOperand` and `ConstantArray::getOperand` return a `Constant`, so I made this a template to make it work with `SmallVector<Constant*>`

coldav (approved):

modules/compiler/utils/source/pass_functions.cpp
hvdijk: But there's two possibilities here, right? Either there's other types of `Constant` that can have non-constant operands and we won't be able to maintain `SmallVector<Constant*>` anyway, or there isn't and we can just cast `Constant::getOperand`'s result to `Constant*`?

hvdijk (approved):

modules/compiler/utils/source/pass_functions.cpp
PietroGhg: Yes you are right, I removed the template and added a `cast`.

PR #177, opened 2023-10-26
[vecz] Packetize sub-group shuffles with uniform indices

frasercrmck:
This extends fixed-width vectorization capabilities to `__mux_sub_group_shuffle` builtins, but only those with uniform indices (where the shuffle index is the same for all invocations in the sub-group). This accounts for the majority of those tested by the SYCL-CTS. Support for varying indices will come down the line, once the other shuffles are covered under similar conditions.

The existing sub-group LIT tests have been split by operation, as they are expected to grow significantly to cover all of the different conditions we can vectorize under.


modules/compiler/vecz/source/transform/packetizer.cpp
```diff
@@ -202,13 +204,24 @@ class Packetizer::Impl : public Packetizer {
...
   /// @return Packetized instruction.
   Value *packetizeGroupBroadcast(Instruction *I);
-  /// @brief Returns true if the instruction is a subgroup shuffle.
+  /// @brief Returns true if the instruction is any subgroup shuffle.
   ///
   /// @param[in] I Instruction to query.
   ///
-  /// @return True if the instruction is a call to a mux subgroup shuffle
+  /// @return The group collective data if the instruction is a call to a mux
+  /// subgroup shuffle-like builtin - std::nullopt otherwise.
```
coldav, line 212: might be useful to give some indication of what a shuffle like instruction is (although is perhaps defined somewhere else).

frasercrmck: I don't want to be overly specific as it's an implementation detail and ultimately depends on the mux spec.

coldav: you mean the mux spec specifies what a shuffle-like builtin is?

frasercrmck: It specifies a bunch of `shuffle*` builtins, yeah

PR #179, opened 2023-10-26
[compiler] Fix up 64-bit atomics.

hvdijk:
# Overview

Fix up 64-bit atomics.

# Reason for change

The initial support for 64-bit atomics in #172 turned out to be incomplete and causes errors in the OpenCL CTS.

# Description of change

* builtins-3.0.h was updated for 64-bit atomics, but builtins.h wasn't. It is now updated too. Note that in accordance with the OpenCL specification, atom_* functions are declared for long* and ulong*, but atomic_* functions are not.
* ReplaceAtomicFuncsPass when replacing inc/dec was still creating constant 1 with a fixed type of i32, rather than the type of the inc/dec. It now uses the right type.
* UnitCL AtomicStreamer is templated to allow testing atomic operations for long. An existing test for atom_inc is split into a version that tests int, and a version that tests long.
* replace-atomics.ll is added to test atom_inc for both int and long.
* replace-c11-atomics.ll is updated to test double in addition to float.

The atom_* and atomic_* functions are a bit of a mess.

In OpenCL 1.2, atomic_* functions operate on int, uint, and float. atom_* functions operate on long, and ulong.

In OpenCL 3.0, atomic_* functions operate on int, uint, and float. atom_* functions operate on int, uint, float, long, and ulong.

Therefore, I made sure to add only 64-bit overloads for atom_* functions, not the atomic_* functions, to builtins.h.

Although no version of OpenCL specifies this, our translation from SPIR-V to LLVM does assume that we are able to accept other overloads as well.

I have manually run both OpenCL CTS's `test_conformance/atomics/test_atomics` and its `test_conformance/c11_atomics/test_c11_atomics` with these changes and have seen both pass.

# Checklist


frasercrmck (approved):
LGTM

PR #180, opened 2023-10-27
[test] More reviews, rework AddLocalBuffer.

hvdijk:
# Overview

More reviews, rework AddLocalBuffer.

# Reason for change

We have UnitCL tests where we under-allocate buffers and access local memory out of bounds.

# Description of change

AddLocalBuffer was error-prone as it was used in conjunction with AddInputBuffer and AddOutputBuffer, which take a number of elements, whereas AddLocalBuffer takes a number of bytes. This commit changes AddLocalBuffer so that it takes a number of elements, and either an element type as a template parameter, or an element size as a function parameter. Updating the tests to use this new API revealed a number of cases where we were still under-allocating buffers.

N/A

# Checklist


frasercrmck (commented):
Looks good, I think this is a more consistent API

frasercrmck (approved):

PR #183, opened 2023-11-01
[vecz] Packetize sub-group shuffle_(up|down) builtins

frasercrmck:
This extends fixed-width vectorization capabilities to `__mux_sub_group_shuffle_up` and `__mux_sub_group_shuffle_down` builtins. Again, these aren't very efficiently vectorized as we have to perform the shuffle for each work-item in a pseudo-scalarized fashion.

### DIFF
--- a/modules/compiler/vecz/source/transform/packetizer.cpp
+++ b/modules/compiler/vecz/source/transform/packetizer.cpp
@@ -152,6 +152,11 @@ class Packetizer::Impl : public Packetizer {
   /// @return Packetized values.
   ValuePacket packetizeAndGet(Value *V, unsigned Width);
 
+  /// @brief Helper to produce a Result from a Packet
+  Packetizer::Result getPacketizationResult(
+      Instruction *I, const SmallVectorImpl<Value *> &Packet,
+      bool UpdateStats = false);
+
   /// @brief Packetize the given value from the function, only if it is a
   /// varying value. Ensures Mask Varying values are handled correctly.
   ///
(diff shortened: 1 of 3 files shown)


modules/compiler/vecz/source/transform/packetizer.cpp
```diff
@@ -803,6 +819,38 @@ PacketRange Packetizer::createPacket(Value *V, unsigned width) {
...
 }
 
+Packetizer::Result Packetizer::Impl::getPacketizationResult(
+    Instruction *I, const SmallVectorImpl<Value *> &Packet, bool UpdateStats) {
+  if (Packet.empty()) {
+    return Result(*this);
+  }
+  auto PacketWidth = Packet.size();
+
+  if (PacketWidth == 1) {
```
coldav, line 829: could add a comment here to explain why PacketWidth==1 is different?

modules/compiler/vecz/source/transform/packetizer.cpp
```diff
@@ -1589,6 +1622,244 @@ Packetizer::Result Packetizer::Impl::packetizeSubgroupShuffleXor(
...
+                        compiler::utils::GroupCollective::OpKind::ShuffleUp) &&
+         "Invalid shuffle kind");
+
+  auto *const CI = cast<CallInst>(I);
+
+  // We don't support scalable vectorization of sub-group shuffles.
+  if (SimdWidth.isScalable()) {
+    return Packetizer::Result(*this);
+  }
+  unsigned const VF = SimdWidth.getFixedValue();
```
coldav, line 1639: I'd personally prefer something more decriptive - I was looking at code below and thinking what's VF?

coldav (approved):

modules/compiler/vecz/source/transform/packetizer.cpp
frasercrmck: Ah, I thought it was well-established by this point that it means "vectorization factor". That's why we have `CA_RISCV_VF`. We use that terminology in `WorkItemLoopsPass`

It's admittedly a bit inconsistent, but I personally have a half-plan to rename `SimdWidth` to something closer to `VF` across the entire vectorizer.

modules/compiler/vecz/source/transform/packetizer.cpp
frasercrmck: Done, thanks

modules/compiler/vecz/source/transform/packetizer.cpp
coldav: It is subtly different to what we call the vectorization factor, as it's only SimdWidth, but feel free to ignore the suggestion.

PR #186, opened 2023-11-02
[cl] Fix typo in comment

frasercrmck:


hvdijk:
"statrt" is a typo, sure, but "Like above" isn't and seems to be relevant to the comment?

frasercrmck:
> "statrt" is a typo, sure, but "Like above" isn't and seems to be relevant to the comment?

It was more that I felt that part didn't real well and didn't contribute anything to the meaning of the comment, so I removed it.

Perhaps it would just be better if I copy/paste the comment I'm assuming it's referring to:

``` cpp
        // We start by setting all dimensions to 1, then copy over the input
        // local size, that way we always have a valid local size.
```

frasercrmck:
> "statrt" is a typo, sure, but "Like above" isn't and seems to be relevant to the comment?

Should be better now.

coldav (approved):

PR #189, opened 2023-11-03
[USM] Accept arbitrary pointers.

hvdijk:
# Overview

Remove much special casing for USM pointers

# Reason for change

The SYCL CTS atomic pointer tests are currently failing, and pass with this change.

# Description of change

The SYCL CTS reveals that the logic we had in clSetKernelArgMemPointerINTEL to detect whether the pointer arguments we passed in were valid, while supported by the cl_intel_unified_shared_memory specification, was causing valid SYCL programs to be rejected, and the specification had an open question on whether this is actually desired. This commit implements a change to just accept arbitrary pointer values, which allows us to also remove a bunch of USM special casing.

# Checklist

### DIFF
--- a/source/cl/include/cl/kernel.h
+++ b/source/cl/include/cl/kernel.h
@@ -262,17 +262,6 @@ struct _cl_kernel final : public cl::base<_cl_kernel> {
     /// @param value_size Size of the data.
     argument(compiler::ArgumentType type, const void *value, size_t value_size);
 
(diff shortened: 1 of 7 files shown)


frasercrmck (approved):
Already pre-reviewed offline, but LGTM.

I don't know enough about this code to know if there are any further clean-ups made possible by this change, and that would be my only suggestion anyway.

PR #193, opened 2023-11-06
Fix typo in test_info.py.

hvdijk:
# Overview

Fix variable name.

# Reason for change

One more error in #190 was the spelling of filter_tests as filtered_tests, which went unnoticed during testing as testing had been done without a filter file.

# Description of change

This change refers to the filter_tests variable using its correct name.

Tested by running the OpenCL CTS with both `-i source/cl/scripts/cts-3.0-online-ignore-linux-host.csv` and `-i source/cl/scripts/cts-3.0-online-ignore-linux-riscv.csv` and making sure that tests run.

# Checklist


frasercrmck (approved):
LGTM

PR #194, opened 2023-11-06
Support SYCL printf.

hvdijk:
# Overview

Support SYCL printf.

# Reason for change

The way DPC++ translates `printf` is invalid and results in assertion errors when we encounter it. We could diagnose that more clearly, but more useful is to be tolerant and translate it to valid LLVM IR.

# Description of change

DPC++ generates invalid SPIR-V for SYCL code containing calls to __builtin_printf. Be tolerant of this.

* Builder::create<OpFunction>: treat "printf" as a special exception and make it variadic.
* getPointerToStringAsString: recognize address space casts of strings.

Tested with a local workaround for the out-of-order command queues problem, with the following SYCL program compiled and linked with DPC++ commit 34a06351da0a7581dea472a199f9803be8433868:

```c++
#include <sycl/sycl.hpp>
int main() {
  sycl::queue queue;
  queue.submit([&](sycl::handler &cgh) {
    cgh.single_task([] {
      __builtin_printf("%s, %s!\n", "Hello", "world");
    });
  });
}
```

# Checklist

### DIFF
--- a/modules/compiler/source/base/source/printf_replacement_pass.cpp
+++ b/modules/compiler/source/base/source/printf_replacement_pass.cpp
@@ -332,16 +332,20 @@ std::optional<std::string> getPointerToStringAsString(Value *op) {
   GlobalVariable *var = dyn_cast<GlobalVariable>(op);
   if (!var) {
     if (auto const_string = dyn_cast<ConstantExpr>(op)) {
-      if (const_string->getOpcode() == Instruction::GetElementPtr) {
-        var = dyn_cast<GlobalVariable>(const_string->getOperand(0));
-      } else if (const_string->getOpcode() == Instruction::IntToPtr) {
-        // Sometimes we see a PtrToInt expression inside an IntToPtr
-        // expression, so therefore we need to unwrap it twice.
-        const_string = dyn_cast<ConstantExpr>(const_string->getOperand(0));
-        if (const_string &&
-            const_string->getOpcode() == Instruction::PtrToInt) {
+      switch (const_string->getOpcode()) {
+        case Instruction::GetElementPtr:
+        case Instruction::AddrSpaceCast:
           var = dyn_cast<GlobalVariable>(const_string->getOperand(0));
-        }
+          break;
+        case Instruction::IntToPtr:
+          // Sometimes we see a PtrToInt expression inside an IntToPtr
+          // expression, so therefore we need to unwrap it twice.
+          const_string = dyn_cast<ConstantExpr>(const_string->getOperand(0));
+          if (const_string &&
+              const_string->getOpcode() == Instruction::PtrToInt) {
+            var = dyn_cast<GlobalVariable>(const_string->getOperand(0));
+          }
+          break;
       }
     } else if (auto gep_string = dyn_cast<GetElementPtrInst>(op)) {
       var = dyn_cast<GlobalVariable>(gep_string->getPointerOperand());
--- a/modules/compiler/spirv-ll/source/builder_core.cpp
+++ b/modules/compiler/spirv-ll/source/builder_core.cpp
@@ -1999,6 +1999,22 @@ cargo::optional<Error> Builder::create<OpFunction>(const OpFunction *op) {
       module.addName(op->IdResult(), name);
     }
 
+    // DPC++ rejects variadic functions in SYCL code, with the exception of
+    // __builtin_printf which it accepts and generates invalid SPIR-V for that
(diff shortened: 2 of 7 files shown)


modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -1999,6 +1999,16 @@ cargo::optional<Error> Builder::create<OpFunction>(const OpFunction *op) {
       module.addName(op->IdResult(), name);
     }
 
+    // DPC++ rejects variadic functions in SYCL code, with the exception of
+    // __builtin_printf which is accepts and generates invalid SPIR-V for. Patch
+    // this up.
+    if (name == "printf" && function_type->getNumParams() == 1 &&
```
coldav, line 2005: This says printf, but the comment says __builtin_printf

hvdijk: The Clang built-in function `__builtin_printf` is translated into a call to the standard library function `printf`, and it's the latter we can detect. Suggestions welcome on how to make it clearer.

coldav: Maybe just say that the __builtin_printf maps to printf.

coldav (approved):

modules/compiler/spirv-ll/source/builder_core.cpp
hvdijk: Thanks, added that with a bit more detail both here and in the commit message.

modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -1999,6 +1999,18 @@ cargo::optional<Error> Builder::create<OpFunction>(const OpFunction *op) {
       module.addName(op->IdResult(), name);
     }
 
+    // DPC++ rejects variadic functions in SYCL code, with the exception of
+    // __builtin_printf which it accepts and generates invalid SPIR-V for that
+    // calls printf, but declares printf as a non-variadic function (because
+    // SPIR-V has no variadic functions) yet calls it with the normal arguments.
+    // Patch this up.
+    if (name == "printf" && function_type->getNumParams() == 1 &&
```
frasercrmck, line 2007: Given our discussion offline I wonder if it's worth adding a bit extra to this comment, saying that this isn't strictly legal to assume in SPIR-V, but we're making use of the fact we only really support translating C/C++/OpenCL-like languages where `printf` is a reserved symbol and we can infer its behaviour.

frasercrmck (approved):
LGTM, with a few minor suggestions

modules/compiler/source/base/source/printf_replacement_pass.cpp
```diff
@@ -332,16 +332,20 @@ std::optional<std::string> getPointerToStringAsString(Value *op) {
...
-        var = dyn_cast<GlobalVariable>(const_string->getOperand(0));
-      } else if (const_string->getOpcode() == Instruction::IntToPtr) {
-        // Sometimes we see a PtrToInt expression inside an IntToPtr
-        // expression, so therefore we need to unwrap it twice.
-        const_string = dyn_cast<ConstantExpr>(const_string->getOperand(0));
-        if (const_string &&
-            const_string->getOpcode() == Instruction::PtrToInt) {
+      switch (const_string->getOpcode()) {
+        case Instruction::GetElementPtr:
+        case Instruction::AddrSpaceCast: {
```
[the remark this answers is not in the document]
hvdijk: Sure, done.

modules/compiler/spirv-ll/source/builder_core.cpp
hvdijk: Thanks, I added a comment about that. Does that look good to you or am I leaving out cases that actually we do care about?

frasercrmck: Thanks! Hmm actually, now you mention it, we technically support Vulkan too. I've no idea what it thinks about `printf`.

frasercrmck: Sorry, maybe it's more correct to say we support SPIR-V generated from GLSL, targeting the vulkan environment?

hvdijk: It seems like the GLSL to SPIR-V translation changes function names which avoids the problem. Updated the comment to mention that.

frasercrmck: Looks good, thanks

PR #196, opened 2023-11-08
Fix one more error in test_info.py.

hvdijk:
# Overview

Fix one more error in test_info.py.

# Reason for change

Currently, while the filter options to run_cities.py no longer cause an error, they do not have the desired effect, filters are ignored.

# Description of change

In #193 I made sure that we would not error out on trying to load filter lists, but did not check that the filter lists worked as intended, and they did not, because list1.append(list2) adds list2 as a single element to list1, whereas the intent was to add each element of list2 to list1, which is what list1.extend(list2) does.

# Checklist


frasercrmck (approved):
LGTM

PR #197, opened 2023-11-09
[spirv-ll] Fix signed ints and duplicate types in 'Kernel' tests

frasercrmck:
This goes a long to way towards cleaning up our LIT tests. It was found that many of them were breaking validation rule 2.16.3:

    Validation Rules for Kernel Capabilities

    * The Signedness in OpTypeInt must always be 0.

It's unclear how this made it into the tests. Presumably they were auto-generated at one point, but I couldn't find the script.

I've gone through all such tests and corrected them to use integers without signedness information. This includes 8-, 16-, and 64-bit integers, which `spirv-val` incorrectly lets slip.

One bug fix is notable:

* When translating the `s_mad24` extended operation, we were incorrectly generating calls to unsigned versions of the `mad24` OpenCL builtin when we didn't have signed integer types.

I've also managed to delete a bunch of redundant tests. For instance, there's no such operation as `s_bitselect` so the tests are now functionally equivalent to the `u_bitselect` tests. This goes also for the tests for "signed" `clz`, `popcount`, `prefetch` `rotate`, `select`, `shuffle` and `shuffle2` operations.

There are some additional redundant test *functions* in cases like the
tests for `OpGroup` where there were separate tests for integers with
signedness and for those without. However, since the operations
themselves don't care for signedness, the two tests became
functionally identical.

Note are tests are still not 'clean'; more (hopefully less impactful) fixes will follow.


modules/compiler/spirv-ll/test/spvasm/op_spec_constant_composite_uint_struct.spvasm
```diff
@@ -39,12 +39,11 @@
        %void = OpTypeVoid
           %3 = OpTypeFunction %void
         %uint = OpTypeInt 32 0
-   %idx_type = OpTypeInt 32 0
         %Foo = OpTypeStruct %uint
  %test_block = OpTypeStruct %Foo
 %_ptr_Uniform_test_block = OpTypePointer Uniform %test_block
           %_ = OpVariable %_ptr_Uniform_test_block Uniform
-         %11 = OpConstant %idx_type 0
+         %11 = OpConstant %uint 0
```
RossBrunton, line 46: Any reason for this change?

modules/compiler/spirv-ll/test/spvasm/op_group_i_add.spvasm
```diff
@@ -101,23 +98,6 @@
...
- ; CHECK: [[RES:%[a-zA-Z0-9]+]] = call spir_func i32 @_Z20sub_group_reduce_addi(i32 [[VAL]])
- %int_sub_group_reduction_res = OpGroupIAdd %int_ty %sub_group_scope Reduce %int_sub_group_reduction_val
- ; CHECK: store i32 [[RES]], ptr addrspace(1) [[IN]]
-                                OpStore %int_sub_group_reduction_in %int_sub_group_reduction_res Aligned 4
- ; CHECK: ret void
-                                OpReturn
-                                OpFunctionEnd
-
- ; CHECK: declare spir_func i32 @_Z20sub_group_reduce_addi(i32)
-
```
hvdijk, line 120: There are a bunch of `CHECK` removals from tests that remove more than I would have expected, is this intentional? Do we no longer generate this at all? Or is this not worth checking, maybe?

modules/compiler/spirv-ll/test/spvasm/op_spec_constant_composite_uint_struct.spvasm
frasercrmck: `%idx_type` is just `%uint` (they're both `OpTypeInt 32 0`) and it's invalid to define any type more than once.

modules/compiler/spirv-ll/test/spvasm/op_group_i_add.spvasm
frasercrmck: Ah right I forgot to mention those but yeah, after removing the signedness, test functions like `int_sub_group_reduction` here are the same as `uint_sub_group_reduction` in the same file.

hvdijk: In that case maybe `int_sub_group_reduction` can be removed entirely, rather than just taking out the checks?

frasercrmck: Ah yes you're right I didn't fully complete the job. One sec.

frasercrmck: That should be that sorted now. Updated the commit message/description too.

hvdijk (approved):

PR #199, opened 2023-11-09
Avoid memory leak in _cl_command_buffer_khr::updateCommandBuffer.

hvdijk:
# Overview

Avoid memory leak in _cl_command_buffer_khr::updateCommandBuffer.

# Reason for change

ASAN builds were showing a memory leak when running e.g. MutableDispatchUSMTest_InvalidArgValue_Test.

# Description of change

In _cl_command_buffer_khr::updateCommandBuffer we need to create a plain old data descriptor, which previously required storing the data externally. We have no pre-existing memory which is guaranteed to persist for long enough that we can use it, but the other descriptor storage types meant that mux_descriptor_info_s itself was already more than large enough to hold this data itself, so this change adds a plain_old_embedded_data descriptor storage type and uses it in _cl_command_buffer_khr::updateCommandBuffer.

This change only adds one trivial use of this new storage type. A future change may add more, but as that is not required to fix the leak that is kept out of this PR.

# Checklist

### DIFF
--- a/doc/modules/mux/changes.rst
+++ b/doc/modules/mux/changes.rst
@@ -11,6 +11,11 @@ version increases mean backward compatible bug fixes have been applied.
    Versions prior to 1.0.0 may contain breaking changes in minor
    versions as the API is still under development.
 
+0.81.0
+------
+
+* Added ``plain_old_embedded_data`` storage type.
+
 0.80.0
 ------
 
--- a/modules/mux/include/mux/mux.h
+++ b/modules/mux/include/mux/mux.h
@@ -37,7 +37,7 @@ extern "C" {
 /// @brief Mux major version number.
 #define MUX_MAJOR_VERSION 0
 /// @brief Mux minor version number.
-#define MUX_MINOR_VERSION 80
+#define MUX_MINOR_VERSION 81
 /// @brief Mux patch version number.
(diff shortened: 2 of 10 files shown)


doc/modules/mux/changes.rst
```diff
@@ -11,6 +11,11 @@ version increases mean backward compatible bug fixes have been applied.
    Versions prior to 1.0.0 may contain breaking changes in minor
    versions as the API is still under development.
 
+0.81.0
+------
+
+* Added ``plain_old_embedded_data`` storage type.
```
coldav, line 17: maybe mention it is part of the mux_descriptor_info_s

modules/mux/include/mux/mux.h
```diff
@@ -2709,6 +2711,14 @@ struct mux_descriptor_info_plain_old_data_s {
   size_t length;
 };
 
+/// @brief A Mux descriptor for plain old data.
+struct mux_descriptor_info_plain_old_embedded_data_s {
+  /// @brief Size in bytes of @p data storage.
+  size_t length;
```
coldav, line 2717: Given we know it can't be longer that size 16, it could be smaller than size_t.

hvdijk:
Closing in favour of #203.

PR #200, opened 2023-11-13
Fix exact multiplication of doubles, fix undefined behavior in is_odd.

hvdijk:
# Overview

Fix exact multiplication of doubles, fix undefined behavior in is_odd.

# Reason for change

We were seeing maths failures in the SYCL CTS.

# Description of change

Fix exact multiplication of doubles: The `multiply_exact_helper<abacus_half>` method has a comment explaining that the non-specialized version is wrong for floating point types with a mantissa with an odd number of bits. Like `half`, `double` also has an odd number of bits in the mantissa, and we were generating incorrect multiplication results as revealed by the SYCL CTS. So remove the `multiply_exact_helper<abacus_half>` specialization and fix the primary template definition instead.

Fix undefined behavior in is_odd: We would previously evaluate `mantissa << unbiasedExp` and then try to mask out the result based on `validExp` which tells us whether this shift is safe to perform. For scalars, `<<` goes by ordinary C++ rules which say the behavior is undefined. For vectors, the specs are less clear and this may be intended to be valid, but regardless, LLVM optimizes it the same as scalar `<<`. The undefined behavior results in LLVM `poison` which cannot be masked away. Mask the exponent as well so that we always have a valid shift count.

This fixes `math_builtin_float_base` and `math_builtin_float_double` in the SYCL CTS. Some of the tests (at least `pow`) require both commits in this PR in order to pass, which is why I have left them as a single PR.

# Checklist

### DIFF
--- a/modules/compiler/builtins/abacus/include/abacus/internal/is_odd.h
+++ b/modules/compiler/builtins/abacus/include/abacus/internal/is_odd.h
@@ -45,13 +45,19 @@ inline typename TypeTraits<T>::SignedType is_odd(const T &x) {
   const SignedType unbiasedExp = abacus::internal::logb_unsafe(x);
 
   // We're using the exponent to shift, we can only shift by [0, <bits in type>)
-  const SignedType validExp = (unbiasedExp >= 0) & (unbiasedExp < numBits);
+  // Scalar relational comparisons return +1 for true, vector relational
+  // comparisons return -1. -1 has all bits set so is useful as a mask, so make
+  // sure we get that even for scalars.
+  const SignedType validExp =
+      TypeTraits<T>::num_elements == 1
+          ? -((unbiasedExp >= 0) & (unbiasedExp < numBits))
+          : (unbiasedExp >= 0) & (unbiasedExp < numBits);
 
   // Shifting our mantissa by the exponent means that the hidden bit now holds
   // the least significant bit of the integer component of the float. If this
   // is set it means that the number is odd.
   const SignedType hiddenBitMasked =
-      (mantissa << unbiasedExp) & mantissaHiddenBit;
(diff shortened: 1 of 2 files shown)


modules/compiler/builtins/abacus/include/abacus/internal/is_odd.h
```diff
@@ -45,13 +45,17 @@ inline typename TypeTraits<T>::SignedType is_odd(const T &x) {
   const SignedType unbiasedExp = abacus::internal::logb_unsafe(x);
 
   // We're using the exponent to shift, we can only shift by [0, <bits in type>)
-  const SignedType validExp = (unbiasedExp >= 0) & (unbiasedExp < numBits);
+  // We will use validExp to mask, so need -1 for true even for scalars.
```
coldav, line 48: Could we add a little more comments here - I'm not quite sure i follow the logic. Also scalar and vector look very different - which confuses me based on your description, which suggested they followed similar shift logic.

hvdijk: Relational operators are different for scalar and vector: for scalar, they return 0 for false, 1 for true, for vectors they return 0 for false, -1 for true. This change makes sure they return 0 for false, -1 for true, both for scalar and for vector, so that the rest of the code works the same for scalar and vector alike. We need -1 (all bits 1) for true because otherwise when we do `validExp & unbiasedExp`, we get the wrong value. Will have a think about how to phrase this better in the comment.

hvdijk: Should hopefully be clearer now, and also expanded the comment on `multiply_exact`.

coldav (approved):

PR #201, opened 2023-11-13
[spirv-ll] Fix up all LIT tests to be valid SPIR-V (or identified as intentionally invalid)

frasercrmck:
This PR constitutes a whole lot of changes that fixes up hundreds of SPIR-V LIT tests which were invalid SPIR-V (according to `spirv-val`, which isn't 100% complete in its coverage of the specification)

There's too many to summarize, but all tests should now be valid SPIR-V except:

* Any tests in `invalid-spir-v/*`, which are tests that are intentionally checking how we handle invalid SPIR-V
* Any tests in `codeplay/*` which may check Codeplay-specific SPIR-V extensions which `spirv-val` doesn't know about
* `intel_arbitrary_precision_integers.spvasm`, about which `spirv-val` complains that an invalid `OpTypeInt` is defined (even though that's the point of the extension)

In a follow-up PR I may experiment with adding `spirv-val` to all of our tests to ensure they're always valid and won't rot in the future. I've left it out of this PR to reduce noise as much as possible.


hvdijk:
The debug info changes look like they weren't meant to be part of this PR?

frasercrmck:
> The debug info changes look like they weren't meant to be part of this PR? Or, if they were meant to be, the PR title is very misleading.

The changes weren't originally planned, but the test was invalid (using `OpVariable` after an `OpStore`) but in order to keep the intended codegen I had to reorder the `OpLine`s, at which point I found that `spirv-ll` wasn't able to do the right thing.

That said it's probably best if I split that into an earlier PR. If I do it after, I can't rightly claim that this PR fixes all known validation issues.

frasercrmck:
Split the debug-info changes into https://github.com/codeplaysoftware/oneapi-construction-kit/pull/202

hvdijk (approved):

PR #203, opened 2023-11-13
Avoid memory leak in _cl_command_buffer_khr::updateCommandBuffer (alternative)

hvdijk:
# Overview

Avoid memory leak in _cl_command_buffer_khr::updateCommandBuffer.

# Reason for change

ASAN builds were showing a memory leak when running e.g. MutableDispatchUSMTest_InvalidArgValue_Test.

# Description of change

In _cl_command_buffer_khr::updateCommandBuffer we need to create a plain old data descriptor, which requires storing the data externally. We have no pre-existing memory which is guaranteed to persist for long enough that we can use it, but we can add a field to UpdateInfo to hold it.

This is an alternative to #199. This is a less invasive fix. If this is picked instead of #199, I will close that one. If #199 is picked, I will close this one.

# Checklist


frasercrmck (commented):
LGTM

PR #204, opened 2023-11-13
[spirv-ll] Validate LIT tests with `spirv-val`

frasercrmck:
This PR helps to ensure that all* LIT tests remain valid SPIR-V as we continue to add and modify LIT tests.

Since we support the LIT tests in two modes - online and offline - the functionality has to be added to both.

* For online, the LIT test itself runs `spirv-val` on the assembled binary. This can easily be controlled per test.
* For offline, cmake and the build system run `spirv-val` on the assembled binary. This is more fiddly to control per test - we use a separately tracked list of tests we explicitly don't want to validate.

*almost all - some are invalid and we're okay with that

### DIFF
--- a/modules/compiler/spirv-ll/test/glsl/CMakeLists.txt
+++ b/modules/compiler/spirv-ll/test/glsl/CMakeLists.txt
@@ -806,9 +806,19 @@ foreach(glsl ${GLSL_FILES})
   if(GLSL_UNSUPPORTED STREQUAL False)
     # Generate .spv from .comp source.
     file(RELATIVE_PATH spvOut ${CMAKE_BINARY_DIR} ${spv})
+
+    # Work out whether we're validating this SPIR-V file.
+    set(OPTIONAL_VALIDATE_DEPENDS)
+    set(OPTIONAL_VALIDATE_COMMAND)
+    if(TARGET spirv::spirv-val)
+      set(OPTIONAL_VALIDATE_DEPENDS spirv::spirv-val)
+      set(OPTIONAL_VALIDATE_COMMAND COMMAND spirv::spirv-val ${spv})
+    endif()
+
     add_custom_command(OUTPUT ${spv}
       COMMAND ${glslangValidator_EXECUTABLE} -s -V -o ${spv} ${comp}
-      DEPENDS ${glslangValidator_EXECUTABLE} ${comp}
+      ${OPTIONAL_VALIDATE_COMMAND}
+      DEPENDS ${glslangValidator_EXECUTABLE} ${OPTIONAL_VALIDATE_DEPENDS} ${comp}
       COMMENT "Building SPIR-V ${spvOut}")
     list(APPEND LitTestFiles ${spv})
   endif()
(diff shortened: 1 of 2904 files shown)


modules/compiler/spirv-ll/test/glsl/CMakeLists.txt
```diff
@@ -806,10 +806,20 @@ foreach(glsl ${GLSL_FILES})
...
+    if(TARGET spirv::spirv-val)
+      set(SPIRV_VAL_DEP spirv::spirv-val)
+      set(SPIRV_VAL_CMD spirv::spirv-val ${spv})
+    endif()
+
     add_custom_command(OUTPUT ${spv}
       COMMAND ${glslangValidator_EXECUTABLE} -s -V -o ${spv} ${comp}
-      DEPENDS ${glslangValidator_EXECUTABLE} ${comp}
-      COMMENT "Building SPIR-V ${spvOut}")
+      COMMAND ${SPIRV_VAL_CMD}
```
hvdijk, line 820: I don't think `add_custom_command` specifies that if a command is empty (`spirv-val` is not found), nothing runs, so I'm a bit cautious about relying on that. Can we avoid that somehow?

frasercrmck: Hmm good point, I hadn't considered that. I don't really know what else to do as my cmake isn't very good. Adding an additional optional `add_custom_command` would probably necessitate some dummy output?

hvdijk: We may be able to set a variable to multiple words, so that `COMMAND` is part of the variable and if no command is found, no second `COMMAND` is included. Not entirely sure that would work though.

frasercrmck: That might work, yeah. Not sure! For now I've gone with a second custom command creating a timestamp file, and using that as an additional dependency to the set of "lit test files"

frasercrmck:
I've made a slight change to this. Now the `spvasm` tests we don't want to verify are written up in a separate list. This should be simpler and more robust than using the directory names. I've updated the PR description accordingly.

modules/compiler/spirv-ll/test/glsl/CMakeLists.txt
hvdijk: Hmm, if we have two `add_command`s anyway, we might as well do
```
if(CONDITION)
    add_custom_command(COMMAND command1)
else()
    add_custom_command(COMMAND command1 COMMAND command2)
end()
```
? It looks equal in CMake complexity to what you have currently, but doesn't generate the unneeded stamp files.

frasercrmck: Fair point. I'd be worried about repetition but maybe it's then a good idea to try the "`COMMAND` inside a variable" thing you suggested earlier. I'm pretty sure we do that elsewhere (not for `COMMAND` but for other cmake arguments)

frasercrmck: I've updated with the first option that we discussed privately.

hvdijk (approved):

PR #205, opened 2023-11-13
[LLVM] Multi_llvm support for current tip changes

RossBrunton:
This patch adds multi_llvm support for some changes that have been made in upstream LLVM's main branch. This will not allow the build to succeed at the moment, but it will skip some errors.

Specifically:
* `CodeGenFileType` and `CodeGenOptLevel` are now enum classes and require a struct to "fake" this in LLVM 17 and earlier.
* `createSimpleTargetReduction` no longer accepts the `TargetTransformInfo` parameter. A proxy function has been added which calls the appropriate version.
* Use `getVirtualFileRef` on the file manager. Seems to work in both LLVM 17 and tip.

# Overview

Adds a bit of support for LLVM tip.

# Reason for change

We'd like to test support for LLVM tip sooner rather than later.

# Description of change

This change doesn't allow LLVM tip to build, but it does fix some of the trivial errors encountered while building.

N/A

# Checklist

### DIFF
--- a/modules/compiler/multi_llvm/include/multi_llvm/enums.h
+++ b/modules/compiler/multi_llvm/include/multi_llvm/enums.h
@@ -1,0 +1,46 @@
+// Copyright (C) Codeplay Software Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+// Exceptions; you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#ifndef MULTI_LLVM_ENUMS_H_INCLUDED
+#define MULTI_LLVM_ENUMS_H_INCLUDED
+
+#include <llvm/Support/CodeGen.h>
+#include <multi_llvm/llvm_version.h>
+
+namespace multi_llvm {
+#if LLVM_VERSION_MAJOR >= 18
+
+typedef llvm::CodeGenFileType CodeGenFileType;
+typedef llvm::CodeGenOptLevel CodeGenOptLevel;
+
+#else
+
+struct CodeGenFileType {
+  static constexpr auto AssemblyFile = llvm::CGFT_AssemblyFile;
+  static constexpr auto ObjectFile = llvm::CGFT_ObjectFile;
+  static constexpr auto Null = llvm::CGFT_Null;
+};
+
+struct CodeGenOptLevel {
+  static constexpr auto None = llvm::CodeGenOpt::None;
+  static constexpr auto Less = llvm::CodeGenOpt::Less;
+  static constexpr auto Default = llvm::CodeGenOpt::Default;
+  static constexpr auto Aggressive = llvm::CodeGenOpt::Aggressive;
+};
+
+#endif
+}  // namespace multi_llvm
+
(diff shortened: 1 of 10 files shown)


coldav (approved):

hvdijk (approved):

modules/compiler/vecz/source/transform/packetization_helpers.cpp
```diff
@@ -253,9 +253,10 @@ Value *createMaybeVPTargetReduction(IRBuilderBase &B,
                                     const TargetTransformInfo &TTI, Value *Val,
                                     RecurKind Kind, Value *VL) {
   assert(isa<VectorType>(Val->getType()) && "Must be vector type");
+  (void)TTI;
```
frasercrmck, line 256: Do we need this?

modules/compiler/multi_llvm/include/multi_llvm/enums.h
```diff
@@ -0,0 +1,47 @@
...
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#ifndef MULTI_LLVM_ENUMS_H_INCLUDED
+#define MULTI_LLVM_ENUMS_H_INCLUDED
+
+#include <multi_llvm/llvm_version.h>
+
+#include "llvm/Support/CodeGen.h"
```
frasercrmck, line 21: Our style is generally `<>` for including LLVM.

modules/compiler/multi_llvm/include/multi_llvm/loop_utils.h
```diff
@@ -0,0 +1,44 @@
...
+#define MULTI_LLVM_LOOP_UTILS_H_INCLUDED
+
+#include <multi_llvm/llvm_version.h>
+
+#include "llvm/Transforms/Utils/LoopUtils.h"
+
+namespace multi_llvm {
+#if LLVM_VERSION_MAJOR >= 18
+
+inline llvm::Value *createSimpleTargetReduction(
```
frasercrmck, line 26: In the past for `multi_llvm` we've usually had one unconditional function with the *body* doing the `#if/#else`.

Something like:

``` cpp
inline llvm::Value *createSimpleTargetReduction(
    llvm::IRBuilderBase &B, const llvm::TargetTransformInfo *TTI,
    llvm::Value *Src, llvm::RecurKind RdxKind) {
#if LLVM_VERSION_MAJOR >= 18
  (void)TTI;
  return llvm::createSimpleTargetReduction(B, Src, RdxKind);
#else
  return llvm::createSimpleTargetReduction(B, TTI, Src, RdxKind);
#endif
}

modules/compiler/vecz/source/transform/packetization_helpers.cpp
hvdijk: Without this, we may get warnings about the unused parameter depending on the compiler. Casting to `void` is the LLVM preferred style to avoid that, to indicate that it is intentional.

frasercrmck: But it is being passed to `multi_llvm::createSimpleTargetReduction` below?

RossBrunton: This is an explicit way of marking a value as unused. I know I could just not give it a name in the signature, but I think this is better because it makes it obvious that it isn't being forwarded to the call.

hvdijk: See conversation, I thought the same thing :)

RossBrunton: Ah, I see now. Sorry for missing that. I also didn't see the conversation because I assumed that Github would poll for new comments every few seconds.

frasercrmck (approved):
LGTM, thanks

modules/compiler/vecz/source/transform/packetization_helpers.cpp
frasercrmck: Yeah there's a lot more forced refreshing than I'm used to on GitLab.

PR #206, opened 2023-11-14
[spirv-ll] Fix LIT test build with 'make' build system

frasercrmck:
Certain tests live in sub-directories relative to the current binary dir. On certain build systems, these relative binary sub-directories may not have been created at the time `spirv-as` runs, which (may) then error. This fix adds an extra `COMMAND` to ensure the binary directories are always generated.

### DIFF
--- a/modules/compiler/spirv-ll/test/spvasm/CMakeLists.txt
+++ b/modules/compiler/spirv-ll/test/spvasm/CMakeLists.txt
@@ -2238,6 +2238,7 @@ function(ca_assemble_spirv_ll_lit_test)
     # Generate .spv from .spvasm source.
     file(RELATIVE_PATH spvOut ${CMAKE_BINARY_DIR} ${spv})
     add_custom_command(OUTPUT ${spv}
+      COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/${dir}
       COMMAND spirv::spirv-as --target-env ${args_TGT_ENV} -o ${spv} ${spvasm}
       DEPENDS spirv::spirv-as ${spvasm}
       COMMENT "Building SPIR-V ${spvOut}")


coldav (approved):

modules/compiler/spirv-ll/test/spvasm/CMakeLists.txt
```diff
@@ -2238,6 +2238,7 @@ function(ca_assemble_spirv_ll_lit_test)
     # Generate .spv from .spvasm source.
     file(RELATIVE_PATH spvOut ${CMAKE_BINARY_DIR} ${spv})
     add_custom_command(OUTPUT ${spv}
+      COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_CURRENT_BINARY_DIR}/${dir}
```
coldav, line 2241: is this okay if the directory already exists?

hvdijk (approved):

modules/compiler/spirv-ll/test/spvasm/CMakeLists.txt
frasercrmck: https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-E-arg-make_directory
``` quote
make_directory <dir>...

    Create <dir> directories. If necessary, create parent directories too. If a directory already exists it will be silently ignored.

    New in version 3.5: Support for multiple input directories.
```

coldav: According to https://cmake.org/cmake/help/v3.10/command/make_directory.html, this is a deprecated command.

frasercrmck: I think that refers to the "command" in CMake. Here we're using the "command-line tool" `make_directory`, which isn't marked as such. At least not in latest the documentation I linked above.

coldav: fair enough, cmake can be confusing.

PR #208, opened 2023-11-14
[spirv-ll] Rename a couple of getter methods for clarity

frasercrmck:
[spirv-ll] Rename 'get' method for clarify
    
    There are many overloaded 'get' functions in spirv-ll, and it can be
    confusing whether we're getting an Op from a spv::Id, an Op from a
    llvm::Value, or an Op from a llvm::Type. This is especially because
    there are many variables we call variants of "ty" or "type":
    
    * `spv::Id structType`
    * `OpType *sourceOpType`
    * `llvm::Type *resultType`
    
    Tracking the types of all these variables can be difficult when many
    things are declared with `auto`. One way of combatting this is by better
    naming the functions which declare these types.

Second commit:

    [spirv-ll] Rename 'getType' method for clarity
    
    This is to try to help prevent it getting confused with a method that
    returns a SPIR-V `OpType` value.

### DIFF
--- a/modules/compiler/spirv-ll/include/spirv-ll/module.h
+++ b/modules/compiler/spirv-ll/include/spirv-ll/module.h
@@ -864,7 +864,7 @@ class Module : public ModuleHeader {
   ///
   /// @return A pointer to the Op or nullptr if not found.
   template <class Op = OpCode>
-  const Op *get(llvm::Type *ty) const {
+  const Op *getFromLLVMTy(llvm::Type *ty) const {
     assert(!ty->isPointerTy() && "can't get the type of a pointer");
     auto found = std::find_if(
         Types.begin(), Types.end(),
--- a/modules/compiler/spirv-ll/include/spirv-ll/module.h
+++ b/modules/compiler/spirv-ll/include/spirv-ll/module.h
@@ -675,7 +675,7 @@ class Module : public ModuleHeader {
   /// @param[in] id The SPIR-V ID to fetch the value for.
   ///
   /// @return A pointer to the Type or nullptr if not found.
-  llvm::Type *getType(spv::Id id) const;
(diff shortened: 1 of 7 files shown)


modules/compiler/spirv-ll/include/spirv-ll/module.h
```diff
@@ -675,7 +675,7 @@ class Module : public ModuleHeader {
   /// @param[in] id The SPIR-V ID to fetch the value for.
   ///
   /// @return A pointer to the Type or nullptr if not found.
-  llvm::Type *getType(spv::Id id) const;
+  llvm::Type *getLLVMType(spv::Id id) const;
```
hvdijk, line 678: LLVM seems to often style its own name as `Llvm` when it appears in the middle of a camelcased identifier, like `getLlvmType`, see e.g. `constructLLVMLinkCommand` (old, been around for ages, removed in https://reviews.llvm.org/D130020) vs `constructLlvmLinkCommand` (new, added in https://reviews.llvm.org/D124151). Do we want to do the same?

frasercrmck: Huh, I didn't expect that. I'm conflicted. We already have a series of `*LLVM*` functions and variables in our codebase; see the results for `[a-z]LLVM` for example. We have no instances of `Llvm` from what I can see.

On the other hand, if it's what LLVM styles itself as, it's probably the right thing to do. Not that LLVM is itself consistent. Maybe that's for historical reasons.

What do you think?

hvdijk (approved):

modules/compiler/spirv-ll/include/spirv-ll/module.h
frasercrmck: I think that's the right call, yep.

PR #211, opened 2023-11-15
[spirv-ll] Fix various debug info generation issues

frasercrmck:
There were a few noted and unnoted issues with our generation of debug info from `OpLine` opcodes.

* Empty line ranges could crash the translator.
* `OpLine`s placed in empty basic blocks could crash the translator.
* `OpLabel` opcodes placed before `OpFunction` or `OpLabel` instructions (anywhere outside of a block) wouldn't influence the line/column of those entities, as is the intention.
* DISubprograms weren't attached to functions.

With this commit, all of these problems are resolved.

There is some refactoring here to make the code easier to work with.

Firstly, we change the representation of the current line range. This improves the internal representation of the internally tracked current "open" line range (i.e., one begun with `OpLine`). We were previously using special undocumented variables to denote that the current range was closed. This is better expressed as a `std::optional`, which is set to a valid pair when the range is open and `std::nullopt` when it's closed.

We also close line ranges more in adherence to the SPIR-V specification. We used to close open line ranges when starting a new basic block, or ending a function. The specification says that ranges close when they hit a *block termination instruction*: one of a certain set of opcodes that end a block or function. Note that the one such opcode we *don't* process is `OpTerminateInvocation`, but we have no support for that as it's too new (SPIR-V 1.6 and above).

Secondly, we process blocks targeted by branches "on demand". Before, we used to collect branch-like instructions (`OpBranch`, `OpBranchConditional`, and `OpSwitch`) and process them once the function was complete. This was harder to intuit, harder to maintain a self-consistent builder class (the special casing was done outside of its control) and was harder to maintain coherent set of debug scopes.

Now, if we process a branch instruction which forward-references a block, we create that basic block on the fly. When we come to the OpLabel for that block, we'll already have a sensible block. The only fixing up is for cosmetic reasons: we move the block to the end of the function to better maintain the layout of the SPIR-V program.

This also removes the special casing for branches when closing a line range. This was necessary before because there wasn't actually a branch at the time we closed the range, so the current iterators weren't quite right and we had to pre-empty the insertion of a branch later.


modules/compiler/spirv-ll/include/spirv-ll/builder.h
```diff
@@ -335,13 +335,32 @@ class Builder {
...
+  /// @brief Gets (or creates) a DICompileUnit for the given OpLine.
+  llvm::DICompileUnit *getOrCreateDICompileUnit(const OpLine *op_line);
+
+  /// @brief Gets (or creates) a DISubprogram for the given function and
+  /// OpLine.
+  llvm::DISubprogram *getOrCreateDebugFunctionScope(llvm::Function *function,
+                                                    const OpLine *op_line);
+
+  /// @brief Creates a new open line range for the given OpLine, contained
+  /// within the basic block.
```
hvdijk, line 350: This name "open line range" is used in various places. Is this based on the existing `OpLineRange`? I don't think `Op` means "open" there, I think it's just the range of `OpLine` where `Op` means "operation", and I don't think we use open ranges, which is an established term meaning ranges where the end points are included, as in the open range (1, 10) means the range of `x` such that `x > 1 && x < 10`.

frasercrmck: You're right about "Op", and no it's not to do with `OpLine` or `OpLineRange`.

I was meaning "open" to mean that it signifies a range that is  "currently open". That is, a range begun by `OpLine` that has not yet been closed by a `OpNoLine` or a block termination instruction.

I'll need to rethink the naming scheme. I probably got a bit lost during the refactoring.

hvdijk: Ah, thanks for the explanation, that makes sense. I don't have a better term to suggest right away either, sorry; if one comes to mind before you can update this I'll be sure to post it.

modules/compiler/spirv-ll/test/spvasm/debug_info_funcs_and_blocks.spvasm
```diff
@@ -0,0 +1,76 @@
...
+; Unless required by applicable law or agreed to in writing, software
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; RUN: %if online-spirv-as %{ spirv-as --target-env %spv_tgt_env -o %spv_file_s %s %}
+; RUN: %if online-spirv-as %{ spirv-val %spv_file_s %}
```
coldav, line 18: if we don't have online-spirv-as, then presumably this needs the offline spirv flag and all would be well?

coldav (approved):

modules/compiler/spirv-ll/include/spirv-ll/builder.h
frasercrmck: It could maybe be a `LineRangeBeginTy` as it's clearly only the beginning of a range, and the type itself doesn't "open" anything - it's other methods that do the opening. We could keep references to "open" in comments when it concerns `getCurrentOpLineRange` and can be clarified (or, you know, just use the word "current").

What do you think?

modules/compiler/spirv-ll/test/spvasm/debug_info_funcs_and_blocks.spvasm
frasercrmck: I'm not sure what you mean by "the offline spirv flag", sorry.

If the `online-spirv-as` isn't set then the test will use any pre-existing `%spv_file_s` file living in the build directory, which should normally have been configured by CMake to rebuild when you do `ninja spirv-ll-spvasm-lit` or `ninja check-spirv-ll-lit` or another target that depends on the files. Or not if you've misconfigured something or are running LIT yourself in a special way, but in that case the test will fail. This is how all of our `spirv-ll` LIT tests work, currently.

hvdijk (approved):

PR #212, opened 2023-11-15
[spirv-ll] Remove deferred processing of OpPhi

frasercrmck:
Following the recent change to branches, we can remove the deferred processing of Phis by:

1. Creating basic blocks on the fly if needed
2. Creating poison values if the value hasn't been visited yet, and finalizing that value later when it is visited

For the latter, this necessitates a change in how `addID` works. Now, when a `addID` is called on an ID which already has a different recorded value, the new value is assumed to be superior, and the old value is replaced with the new. In practice this only came up once in testing (exactly with phis) so I don't expect this to be an impactful change except for this scenario.

This change to `addID` may open up changes for further simplification of other forward-declared values.

### DIFF
--- a/modules/compiler/spirv-ll/include/spirv-ll/builder.h
+++ b/modules/compiler/spirv-ll/include/spirv-ll/builder.h
@@ -315,10 +315,6 @@ class Builder {
   template <class Op>
   cargo::optional<spirv_ll::Error> create(const Op *op);
 
-  /// @brief Populate the incoming edges/values for the given Phi node
-  /// @param op The SpirV Op for the Phi node
-  void populatePhi(OpPhi const &op);
-
   /// @brief A unification of the four very similar access chain functions
   ///
   /// @param opc The OpCode representing the access chain instruction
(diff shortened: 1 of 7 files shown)


hvdijk:
Does `llvm::PoisonValue::get(result_ty)` guarantee to return a new value every time it is evaluated? If not, if I am reading this right, it has the potential to cause errors when multiple PHIs use the same temporary `poison` value.

frasercrmck:
> Does `llvm::PoisonValue::get(result_ty)` guarantee to return a new value every time it is evaluated? If not, if I am reading this right, it has the potential to cause errors when multiple PHIs use the same temporary `poison` value.

Oh that's a good point; that slipped me by. Shame our testing didn't cover it. I'll look into that and get back to you. Worst case scenario is we can record all the forward refs that we need to fix up later, and end up in a similar scenario to what we currently have. At that point it might be worth abandoning this PR.

frasercrmck:
I doubt I'm doing to proceed with this.

PR #213, opened 2023-11-15
[spirv-ll] Fix a couple more crashes in debug info translation

frasercrmck:
This PR combines two commits fixing more crash issues in our translation of SPIR-V with debug info.

First commit:

    [spirv-ll] Fix and refactor parameter-type debug info

    We were passing `align` to the third parameter of
    `DIBuilder.createBasicType`, but that parameter is actually the
    `Encoding`. This meant that we were generating debug information for
    types which were, e.g., `float` for all 4-byte types or `unsigned_char`
    for all 8-byte types. The align is only relevant for pointer types.

    We also take the opportunity to refactor this method, taking in a ID
    rather than an LLVM Type. This gives us more control over the type. For
    example, LLVM pointer types won't be able to get their pointee types as
    they're now opaque. Using the SPIR-V type ID we can better trace the
    types.

    This fixes another bug, as before, generating debug info type for
    pointers was crashing as it tried to get the LLVM type for the pointee
    type.

Second commit:

    [spirv-ll] Fix a debug info crash with BuiltIns

    We were recording ranges of instructions to which to apply debug
    information, but first we were replacing the initial IR for BuiltIn
    variables with calls to the builtin functions. This meant that the
    instruction ranges were sometime invalid, and the translator would
    crash.

    This commit fixes that issue by applying debug info to the original IR,
    before we replace the BuiltIns.

### DIFF
--- a/modules/compiler/spirv-ll/source/builder.cpp
+++ b/modules/compiler/spirv-ll/source/builder.cpp
@@ -15,6 +15,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include <compiler/utils/target_extension_types.h>
+#include <llvm/BinaryFormat/Dwarf.h>
 #include <llvm/IR/Attributes.h>
 #include <llvm/Support/Debug.h>
 #include <llvm/Support/type_traits.h>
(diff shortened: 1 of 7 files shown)


modules/compiler/spirv-ll/source/builder.cpp
```diff
@@ -61,78 +62,68 @@ llvm::Value *spirv_ll::Builder::popFunctionArg() {
...
-llvm::DIType *spirv_ll::Builder::getDIType(llvm::Type *type) {
-  const llvm::DataLayout &datalayout =
+llvm::DIType *spirv_ll::Builder::getDIType(spv::Id tyID) {
+  const llvm::DataLayout &DL =
       IRBuilder.GetInsertBlock()->getModule()->getDataLayout();
 
-  uint32_t align = datalayout.getABITypeAlign(type).value();
+  auto *const llvmTy = module.getLLVMType(tyID);
+  auto *const opTy = module.get<OpType>(tyID);
+  assert(llvmTy && opTy && "Could not get type from spv::Id");
```
hvdijk, line 71: Can you split this into two separate asserts, so that if we get an assertion failure, we know which of the two failed?

hvdijk (approved):

modules/compiler/spirv-ll/source/builder.cpp
frasercrmck: Yes, good idea. Done.

PR #214, opened 2023-11-15
[spirv-ll] Attach names to values more often

frasercrmck:
We weren't doing a very thorough job at making use of `OpName`s when translating SPIR-V to LLVM IR. This is obviously fine as the names are purely decorative. However, they do have a use in making the translated LLVM IR easier to intuit.

This is primarily because we were manually attaching names to the LLVM IR instructions (or blocks, functions, etc.) when creating them. If we forgot to do that, the names would be forgotten.

Instead, if we attach names centrally, we can achieve a far higher coverage. We choose to do this in `addID` which is in practice called during the translation of almost every SPIR-V opcode.

The name is only attached if one hasn't been given already. This means that, while by default the name will be attached to the "last" instruction in the translated sequence, this can be overridden if desired. That said, since names are decorative, this isn't hugely important and it's not often we translate opcodes using multiple instructions. We can fix up any specific cases later if they're particularly egregious.

This gives us the opportunity to more cleanly prioritze the names of entry points over 'regular' functions. Previously we were doing this in a more convoluted two-step process. We can instead do it on the fly.

This also fixes a bug where we could accidentally pick up GLSL functions
called 'printf' as the C-style printf, and incorrectly make it variadic.
A test has been added to cover it.


hvdijk:
> With this commit we now generate:
> 
>     declare void @func(...)

This change will mean it is possible to match a special function name where we previously didn't. In particular, I was relying on GLSL `printf` being mangled to `printf(` so that we do not wrongly give it special treatment.

frasercrmck:
> > With this commit we now generate:
> > ```
> > declare void @func(...)
> > ```
> 
> This change will mean it is possible to match a special function name where we previously didn't. In particular, I was relying on GLSL `printf` being mangled to `printf(` so that we do not wrongly give it special treatment. We should probably have a test for that.

We probably should have a test for this, yes. As far as I can tell from looking at the code, the specific situation you're talking about is checking `name == "printf"`, but earlier on in that function, `name` was set to `name = name.substr(0, name.find_first_of('('))`. It looks to me like this problem you're hypothesizing is already present in our code?

This is kinda of what I'm alluding to at the end of the PR description - this pre-existing handling of `(` is strange. I wonder if it's working around oddities in `glslangValidator`, or whether it's specified what `(` means.

frasercrmck:
I've updated the PR now to undo the changes to how we name GLSL functions - the `(` is now kept. I've explicitly fixed the `printf` bug we discussed offline, and added a LIT test to cover it.

hvdijk (approved):

PR #215, opened 2023-11-16
[spirv-ll] Replace most internal uses of cargo types

frasercrmck:
This commit changes most internal uses of cargo data types:

* `cargo::optional<spirv_ll::Error>` -> `llvm:Error`
* `cargo::optional<T>` -> `std::optional<T>`
* `cargo::expected<T, std::string>` -> `llvm::Expected<T>`

The rationale is that `spirv-ll` is inherently an LLVM-based tool - we don't build the tool at all in an LLVM-less build - and so making use of more "standard" types should be more intuitive - returning `llvm::Error::success()` instead of `cargo::nullopt`, for instance. We can also make use of LLVM's more extensive support for dedicated error handling, compared with our hand-rolled version.

I haven't changed the 'external' (or public) APIs just yet: those which the `compiler` module makes use of. That can come later, if we so wish.

### DIFF
--- a/modules/compiler/spirv-ll/source/context.cpp
+++ b/modules/compiler/spirv-ll/source/context.cpp
@@ -147,7 +147,7 @@ cargo::expected<spirv_ll::Module, spirv_ll::Error> spirv_ll::Context::translate(
   llvm::SmallVector<OpIRLocTy, 8> Phis;
 
   for (auto op : module) {
-    cargo::optional<Error> error;
+    std::optional<llvm::Error> error;
     switch (op.code) {
         // Unsupported opcodes are ignored.
       default:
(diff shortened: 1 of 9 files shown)


modules/compiler/spirv-ll/source/context.cpp
```diff
@@ -146,8 +146,15 @@ cargo::expected<spirv_ll::Module, spirv_ll::Error> spirv_ll::Context::translate(
   // have been generated
   llvm::SmallVector<OpIRLocTy, 8> Phis;
 
+  llvm::Error error = llvm::Error::success();
+  // We need to check this error - even though it's obviously success - because
+  // we're going to re-assign it and that destroys it. *All* Error objects must
+  // be checked before they're destroyed.
```
hvdijk, line 152: We may be able to use `std::optional<llvm::Error>` to avoid having to do special weird things needing a big comment here? If we do that, I think it can also be local to the loop as it was before.

frasercrmck: Great idea. That works, thank you!

hvdijk (approved):

PR #216, opened 2023-11-16
Fix dependency on resources again.

hvdijk:
# Overview

Fix dependency on resources again.

# Reason for change

In #210, I added `OBJECT_DEPENDS` for resources and removed `add_dependencies` as it should no longer be needed. Actually, we need both: the `OBJECT_DEPENDS` documentation says target-level dependencies need to be added as well.

# Description of change

File-level dependencies require target-level dependencies as well. The Ninja generator works fine without them, but the Makefile generator does not.

This partially reverts #210.

# Checklist


frasercrmck (approved):
LGTM if CI is happy

PR #217, opened 2023-11-16
[mux][riscv] Fix getLocalSizeForSubGroupCount

frasercrmck:
We hadn't fully fleshed out this query on riscv. We were returning 0,0,0 (an invalid work-group size) for anything other than 1 sub-group. This was being picked up as a failure by the OpenCL-CTS.

Instead, we must return a work-group size for which the work-group would execute the same number of sub-groups as we've been asked for. The returned work-group size is always 1D, hence we only care about the 'X' dimension.

We do this by taking the maximum sub-group size we've compiled, and returning the count multiplied by that size. This is because we know the kernel variant picking algorithm favours the kernel variant with the highest legal sub-group size. If the required work-group size is invalid we must also report 0,0,0.

This fixes a test failure in the OpenCL CTS.

### DIFF
--- a/modules/mux/targets/riscv/source/kernel.cpp
+++ b/modules/mux/targets/riscv/source/kernel.cpp
@@ -117,19 +117,50 @@ mux_result_t kernel_s::getLocalSizeForSubGroupCount(size_t sub_group_count,
                                                     size_t *out_local_size_x,
                                                     size_t *out_local_size_y,
                                                     size_t *out_local_size_z) {
-  // FIXME: For a single sub-group, we know we can satisfy that with a
-  // work-group of 1,1,1. For any other sub-group count, we should ensure that
-  // the work-group size we report comes back through getKernelVariantForWGSize
-  // when it comes to run it. See CA-4784.
-  if (sub_group_count == 1) {
+  // Grab the maximum sub-group size we've compiled for.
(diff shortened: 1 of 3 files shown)


modules/mux/targets/riscv/source/kernel.cpp
```diff
@@ -117,18 +117,46 @@ mux_result_t kernel_s::getLocalSizeForSubGroupCount(size_t sub_group_count,
...
+  }
+
+  // Double-check that if we were to be asked for the kernel variant for this
+  // work-group size we've reported, we'd receive a kernel variant with the
+  // same sub-group size as we've assumed for the calculations.
+  mux::hal::kernel_variant_s variant;
+  mux_result_t res = getKernelVariantForWGSize(
+      *out_local_size_x, *out_local_size_y, *out_local_size_z, &variant);
+  if (res != mux_success || variant.sub_group_size != max_sub_group_size) {
+    return mux_error_internal;
```
hvdijk, line 159: This is basically an assert, right? It should probably be skipped in a build with assertions disabled, so `#ifndef NDEBUG`...`#endif`?

frasercrmck: It would signify an internal problem, yes. That's a good idea.

hvdijk (approved):

PR #218, opened 2023-11-16
[spirv-ll] Add an interface for OpExtInstImport handlers

frasercrmck:
This goes part way towards making the handling of SPIR-V extended instruction sets more flexible and less dependent on changing `spirv-ll` itself.

By adding an abstract interface through which the handlers are registered and are asked to handle a given `OpExtInst`, users of this library may in the future be able to register their own external handlers, if we so wish. This has been a long-standing wish for the project.

In the short-term, it helpers to separate out almost all of the handling of extended instruction sets into separate files. We were previously defining all the handlers in the common `builder.h` header file.

### DIFF
--- a/modules/compiler/spirv-ll/source/builder_core.cpp
+++ b/modules/compiler/spirv-ll/source/builder_core.cpp
@@ -32,6 +32,9 @@
 #include <multi_llvm/vector_type_helper.h>
 #include <spirv-ll/assert.h>
 #include <spirv-ll/builder.h>
+#include <spirv-ll/builder_glsl.h>
+#include <spirv-ll/builder_group_async_copies.h>
+#include <spirv-ll/builder_opencl.h>
 #include <spirv-ll/context.h>
 #include <spirv-ll/module.h>
 #include <spirv-ll/opcodes.h>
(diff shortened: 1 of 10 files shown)


modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -303,35 +306,35 @@ template <>
 llvm::Error Builder::create<OpExtInstImport>(const OpExtInstImport *op) {
   auto name = op->Name();
   if (name == "GLSL.std.450") {
+    ext_inst_handlers[ExtendedInstrSet::GLSL450] =
+        std::make_unique<GLSLBuilder>(*this, module);
```
hvdijk, line 310: Can we assert here that we didn't have this builder before (that is, that we only encountered the `OpExtInstImport` once) by using `ext_inst_handlers.insert` and asserting on the result?

RossBrunton (approved):
In the future I think there might be value to maintaining a "registry" map of `string IDs -> {ExtendedInstrSet, Constructors}` rather than have them all be hardcoded. But for now, lgtm besides the assert issue.

modules/compiler/spirv-ll/source/builder_core.cpp
frasercrmck: Yeah, something like that might be good. But from what I could tell from the spec (and `spirv-val`) there's no reason a program can't import the same extension multiple times:

```
%1 = OpExtInstImport "GLSL.std.450"
%2 = OpExtInstImport "GLSL.std.450"

%3 = OpExtInst %1 Foo ...
%4 = OpExtInst %2 Bar ...
```

So I think we'd want to handle that. I was planning on just making those share a builder (hence the current structure of enum to handler as opposed to `<id> Set` to handler) as I can't see why multiple builders would need their own private data.

In reality there's kind of a two-layer map going on. From `<id> Set` to `ExtendedInstrSet` to handler.

hvdijk: Oh, if that's valid, sharing the builder makes sense to me. In that case, something like
```c++
    auto &builder = ext_inst_handlers[ExtendedInstrSet::GLSL450];
    if (!builder)
        builder = std::make_unique<GLSLBUilder>(*this, module);
```
should work to avoid multiple instantiations.

frasercrmck: Yeah. I'll probably hide this behind a `Module` method too while I'm at it.

frasercrmck: Ah this currently is in `Builder`. We'll have to think later about how to make this more friendly to external users, as the `Builder` isn't exposed to users at all. Users they call `spirv_ll::Context::translate` as their entry point. Ideally we'd probably want some kind of `Context` method for registering builder handlers before calling `translate`, and that would have to set things up correctly. This would probably need a bit of re-working, and I think it can wait till later.

I'll just stick a method in `Builder` and we can move it later if we so wish.

frasercrmck: Done

frasercrmck:
> In the future I think there might be value to maintaining a "registry" map of `string IDs -> {ExtendedInstrSet, Constructors}` rather than have them all be hardcoded. But for now, lgtm besides the assert issue.

Yep this is definitely the first step towards something like that. We don't really have a mandate for it just yet though, as I'm not aware of anyone using `spirv-ll` in such a way. This was mostly as I need to add yet another builder and stuffing more and more things into `builder.h` was starting to look strange.

PR #220, opened 2023-11-20
[spirv-ll] Nominally support the DebugInfo/OpenCL.DebugInfo.100 sets

frasercrmck:
This adds a skeleton combined builder for these two sets. All of the instructions are ignored, meaning they are not translated to LLVM IR.

Since both instruction sets only produce non-semantic instructions, this is assumed to be safe enough for real-world usage, with a caveat noted in the code that it's probably possible to create a legal SPIR-V binary (from the verifier's point of view, at least) that reads the result of one of these opcodes from another instruction set, at which point we may internally error on not having produced an LLVM value for one a DebugInfo opcode IDs.

This risk is assumed to be minimal enough for real-world SPIR-V modules with debug info that nominally allowing the instruction set is more beneficial than not.

### DIFF
--- a/modules/compiler/spirv-ll/source/builder_debug_info.cpp
+++ b/modules/compiler/spirv-ll/source/builder_debug_info.cpp
@@ -1,0 +1,44 @@
+
+// Copyright (C) Codeplay Software Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+// Exceptions; you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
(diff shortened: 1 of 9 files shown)


modules/compiler/spirv-ll/source/builder_debug_info.cpp
```diff
@@ -0,0 +1,44 @@
...
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <spirv-ll/builder_debug_info.h>
+
+namespace spirv_ll {
+
+llvm::Error DebugInfoBuilder::create(OpExtInst const &) {
```
hvdijk, line 22: For `OpNop`, we return type `cargo::optional<Error>`, value `cargo::nullopt`. Since we are essentially treating it the same, we should probably to the same, so that the caller can assume that it only receives an `llvm::Error` if an error occurred?

frasercrmck: Are you perhaps looking at older code? I recently changed everything to `llvm::Error`, so `OpNop` now returns `llvm::Error::success()` too.

hvdijk: Oh! You changed that in #218, which I didn't have locally yet. In that case, looks good to me.

hvdijk (approved):

PR #221, opened 2023-11-20
Re-enable ARM math tests.

hvdijk:
# Overview

Re-enable ARM math tests.

# Reason for change

These tests were failing because we miscompiled our OpenCL CTS. Now that we have fixed that, all of these should pass.

# Description of change

When we run the OpenCL CTS, we will run math tests that we previously skipped.

I have verified that in a local build of OpenCL CTS, these tests do indeed pass.

# Checklist


frasercrmck (approved):
Nice one

PR #222, opened 2023-11-20
[compiler] Include file offset when loading builtins PCH

RossBrunton:
# Overview
Commit d4a912153488 changed how the offsets into AST bitstream files are handled. This patch changes our loading of the builtinsPCH to include the new `InputFilesOffsetBase` field.
    
As this only affects LLVM tip/18, we also introduce a multi_llvm toggle.

# Reason for change
This was causing LLVM tip to crash when reading the headers on tip.

# Description of change

Upstream LLVM added a new `InputFilesOffsetBase` field for the module file which should be included when offsetting into the module.

As this was crashing our CI systems, a completed build with a debug mode compiler should be sufficient to prove that it fixes things.

N/A

# Checklist

### DIFF
--- a/modules/compiler/source/base/source/module.cpp
+++ b/modules/compiler/source/base/source/module.cpp
@@ -238,7 +238,12 @@ static bool loadKernelAPIHeader(clang::CompilerInstance &compiler,
   // stored inside the PCH file.
   llvm::BitstreamCursor &Cursor = moduleFile->InputFilesCursor;
   clang::SavedStreamPosition SavedPosition(Cursor);
-  if (Cursor.JumpToBit(moduleFile->InputFileOffsets[0])) {
+  uint64_t Base = 0;
+#if LLVM_VERSION_GREATER_EQUAL(18, 0)
+  // LLVM 18 introduces a new offset that should be included
+  Base = moduleFile->InputFilesOffsetBase;
+#endif
+  if (Cursor.JumpToBit(Base + moduleFile->InputFileOffsets[0])) {
     return false;
   }
 
@@ -250,7 +255,7 @@ static bool loadKernelAPIHeader(clang::CompilerInstance &compiler,
   // warning.  However, this assert is guaranteed to be correct due to the
   // above `Cursor.JumpToBit` call, and provides enough information for
   // clang-tidy to understand that the undefined shift is impossible.
-  assert((Cursor.GetCurrentBitNo() == moduleFile->InputFileOffsets[0]) &&
+  assert((Cursor.GetCurrentBitNo() == Base + moduleFile->InputFileOffsets[0]) &&
          "Clang bitstream reader is in invalid state.");
   clang::ASTReader::RecordData Record;
   llvm::StringRef Filename;


modules/compiler/multi_llvm/include/multi_llvm/bitstream_reader.h
```diff
@@ -0,0 +1,36 @@
...
+//
+//     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
hvdijk, line 15: We seem to generally have a blank line after the header comment block. LGTM aside from that one trivial thing.

hvdijk (approved):

modules/compiler/multi_llvm/include/multi_llvm/bitstream_reader.h
```diff
@@ -0,0 +1,36 @@
...
+#ifndef MULTI_LLVM_BITSTREAM_READER_H_INCLUDED
+#define MULTI_LLVM_BITSTREAM_READER_H_INCLUDED
+
+#include <clang/Serialization/ASTReader.h>
```
RossBrunton, line 19: Is multi_llvm allowed to depend on clang headers, btw?

hvdijk: I don't know if there's a policy on that. To me the overhead of setting up a `multi_clang` that works just like `multi_llvm`, except for `clang`, doesn't seem to me like it would be worth it, I think it's fine like this, but happy to wait to see if other people disagree.

modules/compiler/source/base/source/module.cpp
```diff
@@ -238,7 +238,8 @@ static bool loadKernelAPIHeader(clang::CompilerInstance &compiler,
   // stored inside the PCH file.
   llvm::BitstreamCursor &Cursor = moduleFile->InputFilesCursor;
   clang::SavedStreamPosition SavedPosition(Cursor);
-  if (Cursor.JumpToBit(moduleFile->InputFileOffsets[0])) {
+  uint64_t Base = multi_llvm::GetInputFilesOffsetBase(moduleFile);
```
frasercrmck, line 241: Honestly since it's so localized to this one function, I'd be tempted to do this without `multi_llvm` and instead have something like:

``` cpp
uint64_t Base = 0;
Base = File->InputFilesOffsetBase;
#endif
```

We do this in other places, even in this same file.

modules/compiler/source/base/source/module.cpp
```diff
@@ -238,7 +238,12 @@ static bool loadKernelAPIHeader(clang::CompilerInstance &compiler,
   // stored inside the PCH file.
   llvm::BitstreamCursor &Cursor = moduleFile->InputFilesCursor;
   clang::SavedStreamPosition SavedPosition(Cursor);
-  if (Cursor.JumpToBit(moduleFile->InputFileOffsets[0])) {
+  uint64_t Base = 0;
```
hvdijk, line 241: I would expect this to trigger a warning 'warning "Base" set but not used' when building against LLVM 18, depending on the host compiler. Or is that a warning we suppress?

RossBrunton: Using version 11 of... Some compiler (seriously, `--version` should include the compiler name - I assume it's GCC based on the license), I get no warnings. Not sure why this would trigger a warning, Base is being used in the JumpToBit call below.

frasercrmck (approved):
LGTM, thanks. Sorry the policy of to-multi-llvm-or-not-to-multi-llvm isn't very well documented.

modules/compiler/source/base/source/module.cpp
hvdijk: This value being assigned isn't used: it's immediately overwritten with `moduleFile->InputFilesOffsetBase`. Hold on, let me try to get something together on Godbolt that demonstrates the warning and we'll see why it does or doesn't apply here.

hvdijk: Hmm, I must be misremembering, I can only find compiler versions that warn about the whole variable never being used, not one specific assignment. Alright then, if compilers are happy with it, I have no reason to complain.

PR #223, opened 2023-11-21
[spirv-ll] Add some preparation for DebugInfo translation

frasercrmck:
This PR combines three refactoring commits that help make way for the upcoming translation of DebugInfo extended instruction sets.

Commit 1:

    [spirv-ll] Return debug strings as optional values

    Knowing whether a debug string was actually provided is sometimes
    useful, e.g., for error handling. By returning an empty string we are
    unable to tell whether a string was provided and was empty, or whether
    one wasn't provided.

    Changing the return type of this function to
    `std::optional<std::string>` can provide this info for us.

Commit 2:

    [spirv-ll] Return debug strings as optional values

    Knowing whether a debug string was actually provided is sometimes
    useful, e.g., for error handling. By returning an empty string we are
    unable to tell whether a string was provided and was empty, or whether
    one wasn't provided.

    Changing the return type of this function to
    `std::optional<std::string>` can provide this info for us.

Commit 3:

    [spirv-ll] Add method to translate StorageClass to addrspace

    This will be used by the DebugInfo translations. This also makes
    failures in the translation more robust by making the functions return
    an error if the storage class is unknown/unsupported, rather than
    aborting.


modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -104,9 +104,8 @@ llvm::Error Builder::create<OpSource>(const OpSource *op) {
   source += ", Version: " + std::to_string(op->Version());
 
   if (op->wordCount() > 3) {
-    std::string file_path = module.getDebugString(op->File());
-    if (!file_path.empty()) {
-      source += ", Source file: " + file_path + "\r\n";
+    if (auto file_path = module.getDebugString(op->File())) {
```
frasercrmck, line 107: This is a change in behaviour if the file was provided but was empty. We'd now produce `Source file: `. I couldn't work out if this was better than the current behaviour or not.

hvdijk (approved):

PR #227, opened 2023-11-22
Re-enable a bunch more tests.

hvdijk:
# Overview

Re-enable a bunch more tests.

# Reason for change

These tests are currently passing as well, so re-enable them too.

# Description of change

This enables tests that were disabled for vector widths 2, 3, 4, but currently only fail for a vector width of 3, so re-enable the other ones.

# Checklist


RossBrunton (approved):

frasercrmck (approved):
Excellent, thanks! Good old vec3 - never fails to disappoint.

PR #230, opened 2023-11-23
Remove remapped constants in WorkItemLoopPass

PietroGhg:
# Overview

Removes constants after they have been remapped in WorkItemLoopPass.

# Reason for change

Not removing constants leads to a potentially very large increase of uses of the function being replaced.


modules/compiler/utils/source/work_item_loops_pass.cpp
```diff
@@ -1714,7 +1714,9 @@ Function *compiler::utils::WorkItemLoopsPass::makeWrapperFunction(
 
   // Remap any constant expression which take a reference to the old function
   // FIXME: What about the main function?
-  for (auto *user : refF.users()) {
+  for (auto it = refF.user_begin(); it != refF.user_end();) {
+    auto user = *it;
+    it++;
```
hvdijk, line 1719: For `user`, can we continue to have `auto *user` or does that not work? For the iterator, that should either be `user = *it++;` or `user = *it; ++it;`, but for non-builtin-types, best to use pre-increment when you don't use the result.

modules/compiler/test/lit/passes/work-item-loop-remove-replaced.ll
```diff
@@ -0,0 +1,31 @@
...
+;
+;     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+;
+; Unless required by applicable law or agreed to in writing, software
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
hvdijk, line 15: We're not 100% consistent across files but we generally have a blank line after the `SPDX-License-Identifier`, and another one after `RUN:`. (And I'm not sure if the double blank lines later are intentional.)

hvdijk:
This looks good, thanks, I only see some small trivial issues.

modules/compiler/utils/source/work_item_loops_pass.cpp
PietroGhg: No issues with `auto *`, and I've changed to use the pre increment operator, thank you

modules/compiler/test/lit/passes/work-item-loop-remove-replaced.ll
PietroGhg: Should be fixed now, thanks

modules/compiler/utils/source/work_item_loops_pass.cpp
```diff
@@ -1714,7 +1714,9 @@ Function *compiler::utils::WorkItemLoopsPass::makeWrapperFunction(
 
   // Remap any constant expression which take a reference to the old function
   // FIXME: What about the main function?
-  for (auto *user : refF.users()) {
+  for (auto it = refF.user_begin(); it != refF.user_end();) {
```
frasercrmck, line 1717: I've never used it myself but I wonder if this is what `llvm::make_early_inc_range` is for?

frasercrmck (approved):
LGTM

PR #231, opened 2023-11-23
[spirv-ll] Provide helper methods for working with OpExtInst  operands

frasercrmck:
Remembering and hard-coding the value '5' everywhere is brittle and can
be easy to miss when scanning code, leading to bugs. By introducing the
more intuitive notion of having operands of the extended instruction
modeled on top of the operands of the SPIR-V OpExtInst itself, we can
think at a higher level of abstraction.

Also move a whole bunch of GLSL/OpenCL definitions into their cpp files.


modules/compiler/spirv-ll/source/opcodes.cpp
```diff
@@ -421,10 +421,19 @@ spv::Id OpExtInst::Set() const { return getValueAtOffset(3); }
 
 uint32_t OpExtInst::Instruction() const { return getValueAtOffset(4); }
 
+uint16_t OpExtInst::opExtInstOperandCount() const {
+  return wordCount() - OpExtInstBaseOperandOffset;
+}
+
+uint32_t OpExtInst::getOpExtInstOperand(unsigned idx) const {
+  assert(5 + idx < wordCount() && "Invalid operand idx");
```
hvdijk, line 429:
```suggestion
  assert(OpExtInstBaseOperandOffset + idx < wordCount() && "Invalid operand idx");
```

hvdijk (approved):

PR #234, opened 2023-11-27
Use depfiles for clang -cc1.

hvdijk:
# Overview

Use depfiles for clang -cc1.

# Reason for change

When we make changes to header files, we want to ensure their dependees get rebuilt automatically.

# Description of change

This change adds `-dependency-file`, `-MT`, and `-sys-header-deps` to `clang -cc1` invocations, and adds the `DEPFILE` option to use the written file. It also sets the CMake `CMP0116` policy to make the depfiles Ninja-compatible.

# Checklist


frasercrmck (approved):
LGTM

PR #235, opened 2023-11-27
[spirv-ll] Refactor how debug ranges/scopes are handled

frasercrmck:
This commit aims to simplify how we handle the management of different types of debug scope.

Debug info is now attached to instructions when we close a range. We no longer store all of the ranges and process them at the end of the module. This keeps debug information better contained within the builder, and means we have to track less volatile data. The code should be simpler as a result, and hopefully easier to maintain.

We also introduce a new concept to help manage debug info. In addition to the old 'line' range which is built in to SPIR-V, we introduce another: a 'lexical scope'. This will be used for the translation of the various DebugInfo extended instruction sets.

The lexical scopes and line ranges do interact, in the same way that the DebugInfo instruction sets interact with the score spec: the DebugInfo instruction sets still rely on line number information provided by line ranges. A lexical scope is of no use without line information. We may define lexical scopes in one of two ways, in priority order:

1. The DebugInfo extended instruction sets generate them using dedicated instructions
2. We generate them on the fly when attaching debug info when we process line ranges

Thus, when we close a line range or a lexical scope, we apply debug info to all instructions within the range. For the scope information, we take the lexical scope information, if set, and else we generate one on the fly.

### DIFF
--- a/modules/compiler/spirv-ll/include/spirv-ll/builder.h
+++ b/modules/compiler/spirv-ll/include/spirv-ll/builder.h
@@ -21,6 +21,7 @@
 
 #include <llvm/ADT/StringSwitch.h>
 #include <llvm/IR/DIBuilder.h>
+#include <llvm/IR/DebugInfoMetadata.h>
 #include <llvm/IR/IRBuilder.h>
 #include <multi_llvm/multi_llvm.h>
 #include <multi_llvm/vector_type_helper.h>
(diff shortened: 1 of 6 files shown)


modules/compiler/spirv-ll/test/spvasm/debug_info.spvasm
```diff
@@ -95,11 +116,20 @@
...
+; CHECK-DAG: [[ifLexicalBlock]] = distinct !DILexicalBlock(scope: [[mainSubprogram]], file: !{{[0-9]+}}, line: 8, column: 3)
+; CHECK-DAG: [[ifJoinLocation]] = !DILocation(line: 9, column: 5, scope: [[ifLexicalBlock]])
+
+; CHECK: [[testfn2Subprogram]] = distinct !DISubprogram(name: "testfn2", linkageName: "testfn2", scope: null, file: [[File]],
+; CHECK-SAME: line: 5, type: [[mainSubroutineType]]
+; CHECK-SAME: {{(, isLocal: true, isDefinition: true)?}}, scopeLine: 1
+; CHECK-SAME: {{(, isOptimized: false)?}}
+; CHECK-SAME: {{(, spFlags: DISPFlagDefinition)?}}
+; CHECK-SAME: {{(, unit: ![0-9]+)?}}
+; CHECK-SAME: {{(, (variables|retainedNodes): ![0-9]+)?}}
```
hvdijk, line 134: These `CHECK-SAME`s will always pass, regardless of what is output, because they match an empty string.

frasercrmck:
> If the behavior changes, I don't think this can be described as a refactor. The test changes suggest that the behavior does change, but I may be misunderstanding?

Yes I feel it was a bit confusing with the test changes I made (updating to `CHECK-DAG` while locally debugging). There shouldn't be any test changes with this PR. I've updated to reflect that now.

modules/compiler/spirv-ll/test/spvasm/debug_info.spvasm
frasercrmck: Good point! I've updated them now. It's tricky to match against debug info because it changes so frequently. In this test, LLVM 16 puts `retainedNodes` leading to an empty list, but LLVM 17 omits it.

modules/compiler/spirv-ll/include/spirv-ll/builder.h
```diff
@@ -230,8 +231,56 @@ class Builder {
...
+    llvm::Metadata *scope = nullptr;
+    /// @brief An optional scope, representing where the scope was inlined. May
+    /// be nullptr.
+    llvm::Metadata *inlined_at = nullptr;
+  };
+
+  /// @brief Get the currently active debug scope in the current function the
+  /// builder is currently working on
+  ///
+  /// @return The function if one has been declared, otherwise nullptr
```
coldav, line 248: it'd s std::optional here so surely it's an std::nullopt?

modules/compiler/spirv-ll/include/spirv-ll/builder.h
```diff
@@ -230,8 +231,56 @@ class Builder {
...
+  std::optional<LexicalScopeTy> getCurrentFunctionLexicalScope() const;
+
+  /// @brief Set the currently active lexical scope in the current function the
+  /// builder is currently working on; nullptr signals no active scope, or the
```
coldav, line 252: std::nullopt for no active scope?

hvdijk:
> > If the behavior changes, I don't think this can be described as a refactor. The test changes suggest that the behavior does change, but I may be misunderstanding?
> 
> Yes I feel it was a bit confusing with the test changes I made (updating to `CHECK-DAG` while locally debugging). There shouldn't be any test changes with this PR. I've updated to reflect that now.

But there are still test changes in the PR now, and it's still called a refactor? Or is this a test change that would already pass on `main` too?

modules/compiler/spirv-ll/include/spirv-ll/builder.h
```diff
@@ -756,6 +805,12 @@ class Builder {
   llvm::Function *CurrentFunction;
   /// @brief A copy of the current function's argument list
   llvm::SmallVector<llvm::Value *, 8> CurrentFunctionArgs;
+  /// @brief Current debug scope of the function the builder is currently
+  /// working on (or nullopt if no debug scope is active)
+  std::optional<LexicalScopeTy> CurrentFunctionLexicalScope;
+  /// @brief Current line range - marked by the beginning of an OpLine
+  /// instruction - (or nullopt if no line range is active)
```
coldav, line 812:
```suggestion
  /// instruction - (or std::nullopt if no line range is active)
```

modules/compiler/spirv-ll/include/spirv-ll/builder.h
frasercrmck: Ta

modules/compiler/spirv-ll/include/spirv-ll/builder.h
```diff
@@ -230,8 +231,56 @@ class Builder {
...
+  /// @brief Called at the end of a lexical scope for book-keeping.
+  /// @param closing_line_range True if any open line range should be closed at
+  /// the same time.
+  void closeCurrentLexicalScope(bool closing_line_range = true);
+
+  /// @brief A type containing an OpLine line range and the beginning of the
+  /// range it corresponds to.
+  struct LineRangeBeginTy {
+    /// @brief A pointer to the OpLine that this line range corresponds to.
+    const OpLine *op_line;
```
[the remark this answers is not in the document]
frasercrmck: sure, it's more consistent that way.

hvdijk (approved):

PR #237, opened 2023-11-28
[refsi] Update Refsi memory specifications

RossBrunton:
# Overview

In the refsi simulator, there are two memory regions:
* "Main" memory, starting at 0x10000
* "Local" memory, starting at 0x10000000

When emitting the sections, we move the output cursor to location 0x10000000 before writing the sections for local memory. This has the effect of, in some unknown circumstances, resulting in the generated ELF file including padding in the file itself for 255MiB. Besides wasting memory, this also made the refsi simulator unable to load the binaries due to its 128MiB limit.

This patch updates the linker script to be more explicit about the memory layout, which also has the added advantage of having the linker verify that the binary isn't too large.

It also reverts 73c6af20, which was a workaround for this issue.

# Reason for change

This was causing a failure in our internal testing.

# Description of change

Fixes the internal elf file generation to stop it producing unloadable ~255MiB binaries.

I'm still not 100% sure why this issue came up in the first place, but this feels like a nicer way of doing things anyway.

# Checklist

### DIFF
--- a/examples/refsi/hal_refsi/external/refsidrv/source/refsidrv/refsi_device_g.cpp
+++ b/examples/refsi/hal_refsi/external/refsidrv/source/refsidrv/refsi_device_g.cpp
@@ -28,9 +28,7 @@
 // does not have dedicated (TCIM) memory for storing kernel exeutables, a memory
 // window is set up to map this memory area to a reserved area in DMA.
 constexpr const uint64_t REFSI_ELF_BASE = 0x10000ull;
-// The upper region can be up to the tdcm start point at 0x10000000
-// We will make it approx 128MB to give some latitude
-constexpr const uint64_t REFSI_ELF_SIZE = (1 << 27) - REFSI_ELF_BASE;
+constexpr const uint64_t REFSI_ELF_SIZE = (1 << 20) - REFSI_ELF_BASE;
 
 // Memory area for per-hart storage.
 constexpr const uint64_t G_HART_LOCAL_BASE = 0x20800000;
--- a/examples/refsi/hal_refsi/include/device/program.lds
+++ b/examples/refsi/hal_refsi/include/device/program.lds
@@ -17,6 +17,12 @@
 
 OUTPUT_ARCH( "riscv" )
 
+MEMORY
+{
+    mainmem : ORIGIN = 0x10000, LENGTH = (1 << 20)
+    localmem : ORIGIN = 0x10000000, LENGTH = 0x200000
+}
+
 SECTIONS
 {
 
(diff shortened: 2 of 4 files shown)


examples/refsi/hal_refsi/external/refsidrv/source/refsidrv/refsi_device_g.cpp
```diff
@@ -28,9 +28,7 @@
 // does not have dedicated (TCIM) memory for storing kernel exeutables, a memory
 // window is set up to map this memory area to a reserved area in DMA.
 constexpr const uint64_t REFSI_ELF_BASE = 0x10000ull;
-// The upper region can be up to the tdcm start point at 0x10000000
-// We will make it approx 128MB to give some latitude
-constexpr const uint64_t REFSI_ELF_SIZE = (1 << 27) - REFSI_ELF_BASE;
+constexpr const uint64_t REFSI_ELF_SIZE = (1 << 20) - REFSI_ELF_BASE;
```
hvdijk, line 31: Does reducing this risk us rejecting large binaries that were previously accepted?

RossBrunton: In theory, yes. This cuts the binary size down from 128MiB to 1MiB. However, I've not seen binaries anywhere near that big (most are about 1KiB). The only reason the change from `1 << 20` to `1 << 27` was made in the first place was, I think, to work around this issue, which is why I reverted it.

I'm honestly not sure if we should keep it. 1MiB does seem a bit low for complicated kernels.

hvdijk: Ah, https://github.com/codeplaysoftware/oneapi-construction-kit/pull/92, thanks, that makes sense. I do worry that it might be a bit low but if it's what we had before, I'm okay with it, we can always increase it again later if needed.

examples/refsi/hal_refsi/include/device/program.lds
```diff
@@ -94,10 +94,9 @@ SECTIONS
   /*--------------------------------------------------------------------*/
   /* Local memory                                                       */
   /*--------------------------------------------------------------------*/
+  . = ALIGN(0x1000);
 
-  . = 0x10000000;
-
-  .local :
+  .local 0x10000000 :
```
hvdijk, line 99: Can you explain what these two lines do? What I would expect is that if the second line here places `.local` at `0x10000000`, then the previous `. = ALIGN(0x1000);` is not necessary because `0x10000000` is always so aligned, but I am really not sure that my expectation is correct.

hvdijk: Is that actually the effect? When I try it locally, changing GNU ld's default linker script to include
```
...
  . = ALIGN(0x80000);
  .text 0x100000  :
...
```
with a simple test program, then the presence or absence of `. = ALIGN(0x80000);` has no effect on the generated file, they are bitwise identical.

examples/refsi/hal_refsi/include/device/program.lds
```diff
@@ -17,57 +17,57 @@
 
 OUTPUT_ARCH( "riscv" )
 
+MEMORY
+{
+    mainmem : ORIGIN = 0x10000, LENGTH = (1 << 20)
```
hvdijk, line 22: It seems like it's possible to mix explicit regions and implicit regions (looking at GNU ld's own linker scripts, which do that), so we could potentially drop `mainmem`. But I don't mind if you want to keep that in.

RossBrunton: I think it's useful to keep in - If we try to make a binary larger than `1 << 20` bytes, then the linker will fail with an appropriate error message rather than having the program loader crash.

examples/refsi/hal_refsi/external/refsidrv/source/refsidrv/refsi_device_g.cpp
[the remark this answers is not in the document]
RossBrunton: Both the SYCL and OpenCL CTSes pass. If it turns out we need to increase it again in the future, we can always do that as a separate task.

I assumed that `1 << 20` comes from wherever we got the simulator, and I think it makes sense to keep it like that until we know we have a real reason to support larger binaries.

examples/refsi/hal_refsi/include/device/program.lds
hvdijk: That makes sense to me, thanks.

examples/refsi/hal_refsi/include/device/program.lds
RossBrunton: Switched back to explicit memory regions, because specifying origin locations like this didn't actually work.

hvdijk (approved):

coldav (approved):

PR #238, opened 2023-11-29
Fix Doxygen.

hvdijk:
# Overview

Fix Doxygen.

# Reason for change

We did a search and replace in the documentation when we renamed the project, but this search and replace also included variables that came from outside the documentation that need to be kept at their old names.

# Description of change

Restore the CMake variable names that come from outside the documentation directory.

Tested by locally building the documentation.

# Checklist


frasercrmck (approved):
LGTM

RossBrunton (approved):

PR #241, opened 2023-12-02
Add build and run Resnet50 and VGG16 actions to the tartan CI job.

MaryaSharf:
As the networks take a long time to build, a separate action is added to build and run the networks to the tartan CI job on github.


.github/actions/build_vgg_resnet_action/action.yml
```diff
@@ -0,0 +1,75 @@
...
+    default: ${{ github.workspace }}
+
+runs:
+  # We don't want a new docker just a list of steps, so mark as composite
+  using: "composite"
+  steps:
+    - name: setup python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.8'
```
hvdijk, line 16: What's the reason for this specific version? Asking because current Python is 3.12; if there are problems with newer versions, it would be good to know about them.

.github/actions/build_vgg_resnet_action/action.yml
```diff
@@ -0,0 +1,75 @@
...
+
+    - name: Install Python libraries and system dependencies
+      shell: bash
+      run: |
+        pip install requests numpy h5py Pillow
```
hvdijk, line 21: If the below commands use `sudo`, I assume this is not run as root. In that case, best to explicitly use `pip install --user`, it will otherwise give you a warning that it can't do what you want and fall back to `--user`.

.github/actions/build_vgg_resnet_action/action.yml
```diff
@@ -0,0 +1,75 @@
...
+        sudo add-apt-repository universe
+        sudo apt-get update
+        sudo apt-get install -y libblas-dev libopenblas64-dev libopenblas-dev libpthread-stubs0-dev libboost-all-dev
+
+    - name: Clone portDNN
+      shell: bash
+      run: git clone --recursive https://github.com/codeplaysoftware/portDNN.git
```
hvdijk, line 28: Can we use `--depth 1` and `--single-branch` to reduce how much data we fetch here? Or use `wget` to fetch an archive?

.github/actions/build_vgg_resnet_action/action.yml
```diff
@@ -0,0 +1,75 @@
...
+      shell: bash
+      run: |
+        wget "https://github.com/intel/llvm/releases/download/nightly-2023-12-18/sycl_linux.tar.gz"
+        mkdir linux_nightly_release
+        tar -xzf sycl_linux.tar.gz -C linux_nightly_release
+
+    - name: Set up environment and build networks
+      shell: bash
+      run: |
+        ls
```
hvdijk, line 40: Was this left in for debugging during this PR, or do you want `ls` to be kept in?

.github/actions/build_vgg_resnet_action/action.yml
```diff
@@ -0,0 +1,75 @@
...
+        export LD_LIBRARY_PATH=${{ inputs.workspace }}/build/lib:${{ inputs.workspace }}/linux_nightly_release/lib/libsycl.so:${{ inputs.workspace }}/linux_nightly_release/lib:$LD_LIBRARY_PATH
+        export CMAKE_CXX_COMPILER=${{ inputs.workspace }}/linux_nightly_release/bin/clang++
+        export CMAKE_C_COMPILER=${{ inputs.workspace }}/linux_nightly_release/bin/clang
+        export CA_HAL_DEBUG=1
+        export CA_PROFILE_LEVEL=3
+        export ONEAPI_DEVICE_SELECTOR=opencl:acc
+        export OCL_ICD_FILENAMES=${{ inputs.workspace }}/build/lib/libCL.so
+        # To whitelist oneapi-construction-kit for the official Intel oneAPI basetoolkit,
+        # SYCL_CONFIG_FILE_NAME needs to be overridden
```
hvdijk, line 49: I do not understand what this comment means. Can you explain a bit more?

MaryaSharf: >    As the release has a whitelist of devices, it filters out RefSi. To override it, as a temporary solution we can point `SYCL_CONFIG_FILE_NAME` to ``. This way it doesn't set the default `sycl.conf`.

.github/actions/build_vgg_resnet_action/action.yml
MaryaSharf: I'll remove that, thanks

.github/actions/build_vgg_resnet_action/action.yml
MaryaSharf: Good call. :+1:

.github/actions/build_vgg_resnet_action/action.yml
MaryaSharf: I think when I was initially setting up the github ci for tartan, one of the components needed were supported by 3.8 version. I'll have a run with 3.12 to confirm everything works fine and then update this separately.

.github/actions/build_vgg_resnet_action/action.yml
hvdijk: Heh, you changed the other `pip install` that I didn't notice, while leaving this one. I assume that was accidental.

MaryaSharf: Right yeah I missed that. Updated !

.github/actions/build_vgg_resnet_action/action.yml
```diff
@@ -0,0 +1,86 @@
+name: Setup and build networks; VGG16, Resnet50
+description: Build and run VV16 and Resnet50 with oneAPI base toolkit and oneAPI-Construction-Kit
```
coldav, line 2:
```suggestion
description: Build and run VGG16 and Resnet50 with oneAPI base toolkit and oneAPI-Construction-Kit
```

.github/workflows/run_tartan.yml
```diff
@@ -57,4 +61,87 @@ jobs:
...
+        run: |
+          tar -cvzf artifacts.tar.gz build portDNN_build_dir portBLAS_build_dir
+
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: ock_demo_build
+          path: artifacts.tar.gz
+
+  build_and_run_networks:
```
coldav, line 76: not sure where to mention this, but I don't think we want to put the pictures in the ock repo (fine as part of an artifact).

.github/workflows/run_tartan.yml
```diff
@@ -1,8 +1,8 @@
-# Build, Run and Test Tartan EcoSystem
-name: Build and Run Tartan
+# Build, Run and Test OCK Demo EcoSystem
```
coldav, line 1: also rename the workflow

.github/workflows/run_tartan.yml
MaryaSharf: Its not straight forward to get the images from image-net. https://www.image-net.org/challenges/LSVRC/index.php
I can reduce the number of images.

coldav: would it not run in a different way with less images. I'm just not sure I want these binary objects in OCK.

.github/actions/build_vgg_resnet_action/action.yml
```diff
@@ -0,0 +1,90 @@
...
+        python $portDNN_source/samples/networks/resnet50/h5toBin.py resnet50_weights_tf_dim_ordering_tf_kernels.h5
+        cd ..
+
+        # Preparing an image
+        python $portDNN_source/samples/networks/img2bin.py $(pwd)/Labrador_Retriever_Molly.jpg
+
+        # Testing on image for VGG16
+        CA_HAL_DEBUG=1 CA_PROFILE_LEVEL=3 OCL_ICD_FILENAMES=${{ inputs.workspace }}/build/lib/libCL.so ONEAPI_DEVICE_SELECTOR=opencl:acc SYCL_CONFIG_FILE_NAME=  ${{ inputs.workspace }}/portDNN_build_dir/samples/networks/vgg/vgg vdata/ $(pwd)/Labrador_Retriever_Molly.jpg.bin
+        # Testing on image for Resnet50
+        CA_HAL_DEBUG=1 CA_PROFILE_LEVEL=3 OCL_ICD_FILENAMES=${{ inputs.workspace }}/build/lib/libCL.so ONEAPI_DEVICE_SELECTOR=opencl:acc SYCL_CONFIG_FILE_NAME=  ${{ inputs.workspace }}/portDNN_build_dir/samples/networks/resnet50/resnet50 rdata/ $(pwd)/Labrador_Retriever_Molly.jpg.bin
```
coldav, line 79: Should we test it recognises the image?

PR #245, opened 2023-12-04
More LLVM 18 fixups.

hvdijk:
# Overview

More LLVM 18 fixups.

# Reason for change

This fixes a few more build errors against LLVM 18.

# Description of change

* LLVM 18 moves clang::CodeGenOptions::VectorLibrary to llvm::driver::VectorLibrary. Use decltype to handle both.
* LLVM 18 drops CallingConv::WebKit_JS.
* LLVM 18 drops <llvm/Transforms/Vectorize.h> which we include needlessly.
* LLVM 18 requires us to link in libLLVMFrontendDriver, libclangAPINotes, libclangBasic.
* LLVM 18 moves <llvm/Support/Host.h> to <llvm/TargetParser/Host.h>.
* LLVM 18 is able to infer that we could potentially end up with a subvector size of zero, in which case we would end up with a division by zero. A subvector size of zero would be a bug elsewhere in OCK, so add an assert that it is not zero.

Tested against LLVM commit https://github.com/llvm/llvm-project/commit/1c55b227fe7d42b3ad18bf9e485fca66f14fa751.

# Checklist

### DIFF
--- a/modules/compiler/cookie/{{cookiecutter.target_name}}/source/module.cpp
+++ b/modules/compiler/cookie/{{cookiecutter.target_name}}/source/module.cpp
@@ -285,20 +285,23 @@ void {{cookiecutter.target_name.capitalize()}}Module::initializePassMachineryFor
   llvm::Triple TT = llvm::Triple(target_machine->getTargetTriple());
   auto TLII = llvm::TargetLibraryInfoImpl(TT);
 
-  switch (CGO.getVecLib()) {
-    case clang::CodeGenOptions::Accelerate:
+  // getVecLib()'s return type changed in LLVM 18.
+  auto VecLib = CGO.getVecLib();
+  using VecLibT = decltype(VecLib);
+  switch (VecLib) {
+    case VecLibT::Accelerate:
       TLII.addVectorizableFunctionsFromVecLib(
           llvm::TargetLibraryInfoImpl::Accelerate, TT);
       break;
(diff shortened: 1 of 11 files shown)


frasercrmck (approved):

RossBrunton (approved):

coldav (approved):

modules/compiler/cookie/{{cookiecutter.target_name}}/source/module.cpp
```diff
@@ -285,20 +285,22 @@ void {{cookiecutter.target_name.capitalize()}}Module::initializePassMachineryFor
   llvm::Triple TT = llvm::Triple(target_machine->getTargetTriple());
   auto TLII = llvm::TargetLibraryInfoImpl(TT);
 
-  switch (CGO.getVecLib()) {
-    case clang::CodeGenOptions::Accelerate:
+  auto VecLib = CGO.getVecLib();
+  using VecLibT = decltype(VecLib);
```
RossBrunton, line 289: Actually, would it make sense to leave a comment here saying why we are doing this? Just in case someone sees this in the far future when LLVM 22 is out and can undo it.

hvdijk: Sure, added a comment (in all three files that have this).

RossBrunton (approved):

PR #246, opened 2023-12-04
[spirv-ll] Fully support OpenCL.DebugInfo.100 instructions

frasercrmck:
This commit extends spirv-ll to consume OpenCL.DebugInfo.100 instructions and translate them to meaningful LLVM debug info.

Although the legacy DebugInfo instruction set is similar, there are differences, and we aren't seeing real-world binaries with this extended instruction set and so lack tests to confidently enable it.

The translation process differs slightly from the translation of other extended instruction sets. There is a small core of 'root' instructions in the debug info hierarchy, and many more 'leaves' which don't meaningfully add debug information on their own. Thus the translator only processes the roots on demand as we visit them in program order. Those roots may then reference other instructions (roots or leaves) which we translate on the fly. We cache the translation of each instruction as, in practice, they are usually referenced multiple times.

Generally, instructions are translated to `llvm::Metadata*`. The special instruction `DebugInfoNone` translates to `nullptr`: this is not an error; each instruction has to decide which operands may or may not be `nullptr`. This usually ends up depending on what the LLVM APIs accept - some accept `nullptr` as a valid value, others crash or assert.

One thing to note about the `OpenCL.DebugInfo.100` instruction set is that there are numerous bugs all over the ecosystem:

* Even recent versions of `spirv-val` complain about what should be valid SPIR-V binaries
* `llvm-spirv` (and thus DPC++) contains known bugs that won't be fixed soon. We try to work around these as best as possible
  * Mixed instruction opcodes, explain below
  * Extra dummy operands which shouldn't be there
  * Forward references where none should be allowed
  * etc
* `llvm-spirv` (and thus DPC++) support undocumented extensions of their own, such as the ability to translate more LLVM expression opcodes than the spec permits.

The most egregious bug is that `llvm-spirv` has mixed up two instruction
opcodes and thus generates binaries which a standards-compliant SPIR-V
consumer will struggle with. To cope with this, we add a bitfield of
'workarounds' to the DebugInfoBuilder, which can toggle on and off
behaviour.

The only workaround added so far instructs the translator to, when faced
with one of these two instructions, try and infer which one is intended.
It does by inspecting the operands, using rules to establish whether the
instruction opcode is correct or is swapped with the other one. This
workaround is *enabled* by default, as it's still not fixed upstream. We
can toggle this behaviour off later, perhaps.

### DIFF
--- a/modules/compiler/spirv-ll/include/spirv-ll/builder_debug_info.h
+++ b/modules/compiler/spirv-ll/include/spirv-ll/builder_debug_info.h
@@ -18,6 +18,8 @@
 #define SPIRV_LL_SPV_BUILDER_DEBUG_INFO_H_INCLUDED
 
 #include <spirv-ll/builder.h>
+#include <spirv-ll/module.h>
+#include <spirv-ll/opcodes.h>
 
 namespace spirv_ll {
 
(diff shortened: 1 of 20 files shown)


modules/compiler/spirv-ll/include/spirv-ll/builder_debug_info.h
```diff
@@ -18,22 +18,135 @@
...
 /// instruction sets.
 class DebugInfoBuilder : public ExtInstSetHandler {
  public:
+  enum Workarounds {
+    // Some versions of llvm-spirv mistakenly swap
+    // DebugTypeTemplateTemplateParameter and DebugTypeTemplateParameterPack
+    // opcodes, leading to incorrect binaries. When this workaround is enabled,
+    // we assume binaries have been created with this bug.
+    // See https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2248
+    TemplateTemplateSwapParamterPack = 1 << 0,
```
hvdijk, line 36: Typo in the name. As long as the name needs updating anyway, something like `TemplateTemplateSwappedWithParameterPack` might be slightly clearer.

frasercrmck: Good spot (I rushed that bit, sorry) and yeah I've taken your suggestion on the renaming.

Furthermore, the workaround now tries its best to infer which instruction it's received. We have tests compiled with `spirv-as` without this bug and real-world binaries with `llvm-spirv` which do have this bug, so it's unreasonable to statically choose one or the other. It's not a foolproof system, but it should be fairly accurate in its determinations.

modules/compiler/spirv-ll/include/spirv-ll/module.h
```diff
@@ -362,6 +363,26 @@ class Module : public ModuleHeader {
...
+  void setModuleProcess(const std::string &str);
+
+  /// @brief Gets the string used to hold the process/processor
+  const std::string &getModuleProcess() const;
+
+  /// @brief Check if this ID is an OpExtInst with the given opcode.
+  ///
+  /// @return Returns true if this opcode is an OpExtInst with the given
+  /// opcode, false otherwise.
+  bool isOpExtInst(spv::Id id, uint32_t opcode,
```
coldav, line 376: I find the name confusing with the fact that is seems to check it is an OpeExtInst and then checks against the opcode as something extra.

modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -353,12 +362,19 @@ llvm::Error Builder::create<OpExtInstImport>(const OpExtInstImport *op) {
     module.associateExtendedInstrSet(op->IdResult(),
                                      ExtendedInstrSet::GroupAsyncCopies);
   } else if (name == "DebugInfo") {
-    registerExtInstHandler<DebugInfoBuilder>(ExtendedInstrSet::DebugInfo);
+    // Work around a known llvm-spirv, until it's generally fixed upstream.
```
coldav, line 365: `a known llvm-spirv` do you mean  a known llvm-spirv bug (also below)

modules/compiler/spirv-ll/source/builder_debug_info.cpp
```diff
@@ -15,29 +15,2444 @@
...
+    return err;
+  }
+  unsigned addrspace = 0;
+  auto addrspace_or_error =
+      module.translateStorageClassToAddrSpace(op->StorageClass());
+  if (auto err = addrspace_or_error.takeError()) {
+    // Silently consume this error. We know that llvm-spirv will use ~0 to
+    // represent "no address space", despite this being invalid SPIR-V. We
+    // translate it to address space 0.
+    addrspace = 0;
```
RossBrunton, line 505: Does this address space come from Clang? LLVM IR doesn't have the concept of "no address space".

frasercrmck: You're right about LLVM.

What `llvm-spirv` does is check whether the `llvm::DIDerivedType` has an address space (which is a `std::optional<unsigned>`). So I suspect that during code generation, clang might omit the address space on things like `int *ptr`? I can double check that. I also just noticed that we don't necessarily have to set it to `0` - we can just forward on the `std::optional`.

modules/compiler/spirv-ll/include/spirv-ll/module.h
```diff
@@ -793,13 +814,14 @@ class Module : public ModuleHeader {
   ///
   /// @return A pointer to the Op or nullptr if not found.
   template <class Op = OpCode>
-  const Op *get(spv::Id id) const {
-    auto ty = Types.find(id);
-    if (ty != Types.end()) {
+  const Op *get_or_null(spv::Id id) const {
```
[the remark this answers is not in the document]
frasercrmck: I don't know what you mean here. You mean the name of the function?

modules/compiler/spirv-ll/source/builder_debug_info.cpp
```diff
@@ -15,29 +15,2444 @@
...
+    }
+    return llvm::Error::success();
+  }
+
+  return makeStringError("Unexpected imported entity kind");
+}
+
+llvm::Expected<llvm::MDNode *>
+DebugInfoBuilder::translateTemplateTemplateParameterOrTemplateParameterPack(
+    const OpExtInst *op) {
```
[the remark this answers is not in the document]
frasercrmck: Yeah I was really hopeful about that, but unfortunately it could also be `DebugInfoNone`. I believe `Template Name` can also be `DebugInfoNone` (not sure - it's not immediately clear if that makes sense to "optimize out").

I think seeing a positive `DebugSource` match for operand 2 is a sure-fire sign we have `DebugTypeTemplateParameterPack`.

I think seeing a positive `OpString` match for operand 2 is a sure-fire sign we have `DebugTypeTemplateTemplateParameter`.

If neither of these are true (which does happen, sadly), we necessarily need to check operands which may be  `spv::Id` or may be a literal number, depending on which instruction we are inspecting. Either 3 or 5 are interesting.

Fundamentally we don't ever really _know_ if an operand is a `spv::Id` or a literal number. So I think there'll always be a bit of ambiguity if we check operand 3 or 5. That's why I tried to go with matching pairs of operands 3 and 5, to hopefully mitigate the risk that we're just looking at numbers which happen to correspond to valid IDs.

This is the "algorithm" I tried to encode in this function. Maybe it's hard to follow?

modules/compiler/spirv-ll/include/spirv-ll/module.h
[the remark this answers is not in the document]
hvdijk: `dyn_cast_if_present` is one that accepts a `nullptr` as input. The version that returns `nullptr` on failure is `dyn_cast_or_null`, and `get_or_null` conforms to the naming of that.

modules/compiler/spirv-ll/source/builder_debug_info.cpp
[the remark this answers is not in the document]
hvdijk: Ah, right. Let me write this down:

* Operand 0: always `OpTypeVoid`, which tells us nothing.
* Operand 1: always `OpString`, which tells us nothing.
* Operand 2: either `OpString` (`DebugTypeTemplateTemplateParameter`) or `DebugSource` (`DebugTypeTemplateParameterPack`). or `DebugInfoNone`.
* Operand 3: either `DebugSource` (`DebugTypeTemplateTemplateParameter`) or `word` (`DebugTypeTemplateParameterPack`), which cannot be distinguished reliably.
* Operand 4: always `word`, which tells us nothing.
* Operand 5: either `word` (`DebugTypeTemplateTemplateParameter`) or `DebugTypeTemplateParameter` (`DebugTypeTemplateParameterPack`), which cannot be distinguished reliably.

So for operand 3 and 5, if it cannot be a valid ID, it must be a `word`, which tells us what instruction we're looking at. But it's possible that both are a valid ID, in which case we can't figure anything out. Thanks, that makes sense, and I think that's the best that we can do.

modules/compiler/spirv-ll/include/spirv-ll/module.h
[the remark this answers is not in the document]
frasercrmck: Oh good point, yes. I must have subconsciously used that pattern.

modules/compiler/spirv-ll/source/builder_debug_info.cpp
hvdijk: [`DIBuilder::getPointerType`](https://llvm.org/doxygen/classllvm_1_1DIBuilder.html#a8f163bf8a14c1c7823f6d8fce7fb5ac7) takes the same `std::optional<unsigned>` that the original `DerivedType` had. I think we should preserve that, I think we shouldn't use any particular address space as the default address space, and we should translate `~0U` back to `std::nullopt`.

modules/compiler/spirv-ll/source/builder_debug_info.cpp
```diff
@@ -15,29 +15,2466 @@
...
+    // This is 32, 64, or 0 if no memory model is specified.
+    uint64_t size = module.getAddressingModel();
+    type = dib.createPointerType(*base_ty, size, /*align*/ 0, addrspace);
+  }
+
+  if (flags & OpenCLDebugInfo100FlagObjectPointer) {
+    type = dib.createObjectPointerType(type);
+  } else if (flags & OpenCLDebugInfo100FlagArtificial) {
+    type = dib.createArtificialType(type);
+  }
```
hvdijk, line 530: These flags are not documented at all, are they? Given that this specifically calls out *object* pointers, and that DPC++ supports function pointers as an extension, could you add a test for whatever debug info DPC++ generates for those? As long as that is handled sensibly, even if debug information is left out for it, it looks good to me, thank you.

hvdijk: Oh, I just realised this maps to DWARF's `DW_AT_object_pointer` which does not mean object pointers in general, it just means the implicit `this` object parameter of non-static member functions. There is otherwise no difference between object and function pointers. Feel free to ignore this, then.

hvdijk (approved):

modules/compiler/spirv-ll/include/spirv-ll/module.h
frasercrmck: Yeah, I couldn't think of a better name. It's a bit confusing because there are two opcodes going on. `OpExtInst` is itself an opcode enumeration, but each `OpExtInst` has an operand which denotes the specific opcode of the extended instruction.

PR #247, opened 2023-12-06
Increase main memory size limit in memory map

RossBrunton:
# Overview
The SYCL CTS generates binaries larger than 1MiB, so this patch increases the limit.

# Reason for change

Reverts the size shrinking of e7f57440953ed62796de5459e63d398f29ef0df0 , which broke the CTS. Turns out the SYCL CTS requires binaries larger than 1MiB.

# Description of change

Size has been increased back to what it was before e7f57440953ed62796de5459e63d398f29ef0df0 

N/A

# Checklist


coldav (approved):

hvdijk:
> Doesn't seem to include the elf binary location (at `0x1000`) though.

It has comments `// Set up a memory window for ELF executables if needed.` and `// Allocate 'ELF' memory in DRAM, to store kernel executables.` though. What is it doing, then, and how does it relate to the ELF binaries?

hvdijk (approved):
Thanks, I'm not 100% sure I fully get it just yet but I do think you're right I'm concerned over nothing here.

PR #250, opened 2023-12-07
Basic shared USM support

RossBrunton:
# Overview
This implements shared allocations for USM using the same machinery as host allocations. A shared USM allocation can be associated with at most one device, and is accessable on host and that device.

This should be compliant with the extension spec, but does not yet support cross device access or migrating the allocation to the device. Both `CL_MEM_ALLOC_INITIAL_PLACEMENT_DEVICE_INTEL` and `clEnqueueMigrateMemINTEL` have no effect.

# Reason for change

Some tests and examples required shared USM support.

# Description of change

Hacked together shared USM support using host memory support. It's not efficient, but it should be compliant and not require any extra work by the backends.

I had to update a test which required cross-device support to consider shared USM supported, but I don't think that's required as per the spec.

Spec: https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_unified_shared_memory.html

# Checklist

### DIFF
--- a/source/cl/source/extension/include/extension/intel_unified_shared_memory.h
+++ b/source/cl/source/extension/include/extension/intel_unified_shared_memory.h
@@ -103,6 +103,11 @@ class allocation_info {
   /// @return A matching Mux buffer object on success, or nullptr on failure
   virtual mux_buffer_t getMuxBufferForDevice(cl_device_id device) const = 0;
 
+  /// @brief Return the USM memory type for this allocation
+  ///
+  /// @return The enum variant corresponding to this memory allocation type
+  virtual cl_unified_shared_memory_type_intel getMemoryType() const = 0;
+
   /// @brief Checks if a pointer belongs to USM allocation
   ///
   /// @param[in] ptr Pointer to host memory address.
@@ -192,6 +197,13 @@ class host_allocation_info final : public allocation_info {
   /// @return A matching Mux buffer object on success, or nullptr on failure
   mux_buffer_t getMuxBufferForDevice(cl_device_id device) const override;
 
+  /// @brief Return the USM memory type for this allocation
+  ///
+  /// @return The enum variant corresponding to this memory allocation type
+  cl_unified_shared_memory_type_intel getMemoryType() const override {
+    return CL_MEM_TYPE_HOST_INTEL;
+  }
+
   /// @brief Mux memory object bound to every device in the OpenCL context.
   cargo::dynamic_array<mux_memory_t> mux_memories;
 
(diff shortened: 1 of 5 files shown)


coldav (approved):

source/cl/source/extension/include/extension/intel_unified_shared_memory.h
```diff
@@ -192,6 +197,13 @@ class host_allocation_info final : public allocation_info {
   /// @return A matching Mux buffer object on success, or nullptr on failure
   mux_buffer_t getMuxBufferForDevice(cl_device_id device) const override;
 
+  /// @brief Return the USM memory type for this allocation
+  ///
+  /// @return The enum variant corresponding to this memory allocation type
+  virtual cl_unified_shared_memory_type_intel getMemoryType() const override {
```
hvdijk, line 203: The `virtual` keyword is redundant here. We are not consistent in whether we leave it in or take it out; the function right above this one has `override` without `virtual`, but there are plenty of functions that use both as well. Do we have a documented style that we should adhere to for new code?

source/cl/source/extension/source/intel_unified_shared_memory/intel_unified_shared_memory.cpp
```diff
@@ -37,10 +37,11 @@ cargo::expected<cl_mem_alloc_flags_intel, cl_int> parseProperties(
...
         case CL_MEM_ALLOC_FLAGS_INTEL: {
           if (0 == (seen & CL_MEM_ALLOC_FLAGS_INTEL)) {
-            if (value & (CL_MEM_ALLOC_INITIAL_PLACEMENT_DEVICE_INTEL |
-                         CL_MEM_ALLOC_INITIAL_PLACEMENT_HOST_INTEL)) {
-              // These options are valid only for shared memory allocations,
-              // which we don't yet support.
+            const auto PLACEMENT = CL_MEM_ALLOC_INITIAL_PLACEMENT_DEVICE_INTEL |
+                                   CL_MEM_ALLOC_INITIAL_PLACEMENT_HOST_INTEL;
+
+            if ((value & PLACEMENT) == PLACEMENT) {
```
hvdijk, line 43: This style works only if we have exactly two flags. Personally, I think this would be more readable as
```c++
switch (value & PLACEMENT) {
case CL_MEM_ALLOC_INITIAL_PLACEMENT_HOST_INTEL:
  ...; break;
case CL_MEM_ALLOC_INITIAL_PLACEMENT_DEVICE_INTEL:
  ...; break;
default:
  return cargo::make_unexpected(CL_INVALID_PROPERTY);
}
```

source/cl/source/extension/include/extension/intel_unified_shared_memory.h
coldav: I think you are right. I think we tend to rely on clang-tidy etc to bring up things.

source/cl/source/extension/include/extension/intel_unified_shared_memory.h
```diff
@@ -254,6 +273,74 @@ class device_allocation_info final : public allocation_info {
...
+
+  /// @brief Destructor.
+  ~shared_allocation_info() override;
+
+  /// @brief Get device used when allocation was made, or nullptr if no device
+  /// was given
+  cl_device_id getDevice() const override { return device; }
+
+  /// @brief Allocates memory for the USM allocation and binds it to a Mux
+  /// buffer object if a device is given. Sets 'mux_memory' and 'mux_buffer'
```
coldav, line 308: it says if a device is given - but the create requires a device?

source/cl/source/extension/source/intel_unified_shared_memory/intel_unified_shared_memory.cpp
```diff
@@ -418,6 +424,141 @@ cl_int device_allocation_info::allocate(cl_uint alignment) {
...
+    cl_uint max_align = 1;
+    for (auto device : context->devices) {
+      const auto device_align = device->min_data_type_align_size;
+      max_align = device_align > max_align ? device_align : max_align;
+    }
+    alignment = max_align;
+  }
+
+  auto usm_alloc = std::unique_ptr<shared_allocation_info>(
+      new (std::nothrow) shared_allocation_info(context, device, size));
```
hvdijk, line 500: This uses `new (std::nothrow)` allowing a `nullptr` return, but then unconditionally dereferences the result.

source/cl/source/extension/source/intel_unified_shared_memory/intel_unified_shared_memory.cpp
```diff
@@ -418,6 +424,141 @@ cl_int device_allocation_info::allocate(cl_uint alignment) {
...
+
+  cl_int error = usm_alloc->allocate(alignment);
+  OCL_CHECK(error != CL_SUCCESS, return cargo::make_unexpected(error));
+
+  usm_alloc->alloc_flags = alloc_properties.value();
+
+#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ < 9
+  // GCC <9 requires this redundant move, this branch of the #if can be
+  // deleted once the minimum supported version of GCC is at least 9.
+  return std::move(usm_alloc);
```
hvdijk, line 519: Could you give a description of the bug that you are working around here (can be a code snippet here, a GCC bug report link, a Godbolt link, whatever is most convenient) I tried to figure it out by experimenting on Godbolt, but even with non-movable types, even with a mismatch between local variable type and return type, GCC accepted my attempts before GCC 9 as well, so there is something more to it.

source/cl/source/extension/include/extension/intel_unified_shared_memory.h
RossBrunton: Create does not, `cl_device_id` is a typedefed pointer type.

coldav: ah got you it does not require a device.

source/cl/source/extension/source/intel_unified_shared_memory/intel_unified_shared_memory.cpp
```diff
@@ -418,6 +424,141 @@ cl_int device_allocation_info::allocate(cl_uint alignment) {
...
+  }
+};
+
+cargo::expected<std::unique_ptr<shared_allocation_info>, cl_int>
+shared_allocation_info::create(cl_context context, cl_device_id device,
+                               const cl_mem_properties_intel *properties,
+                               size_t size, cl_uint alignment) {
+  OCL_CHECK(size == 0, return cargo::make_unexpected(CL_INVALID_BUFFER_SIZE));
+
+  OCL_CHECK((alignment & (alignment - 1)) != 0,
```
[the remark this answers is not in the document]
coldav: I always think a comment for this sort of thing is useful as not everyone knows the bit hacks anyway.

source/cl/source/extension/source/intel_unified_shared_memory/intel_unified_shared_memory.cpp
RossBrunton: We only have two "location" flags as per the spec, and I don't see them adding any more ones. With that being the case, I'm not sure the switch statement here is more readable. Especially given that we don't do anything here with the arguments we parse, so they'd just be bare break statements.

source/cl/source/extension/source/intel_unified_shared_memory/intel_unified_shared_memory.cpp
RossBrunton: Created internal issue for this - it's in multiple places and we want to see if it's still needed.

source/cl/source/extension/source/intel_unified_shared_memory/intel_unified_shared_memory.cpp
```diff
@@ -418,6 +435,148 @@ cl_int device_allocation_info::allocate(cl_uint alignment) {
...
+              return cargo::make_unexpected(CL_INVALID_VALUE));
+  } else {
+    // If no device is given, all devices that support shared USM must meet
+    // the max_alloc_size and alignment requirements
+    for (auto device : context->devices) {
+      if (deviceSupportsSharedAllocations(device)) {
+        const auto device_align = device->mux_device->info->buffer_alignment;
+        OCL_CHECK(size > device->max_mem_alloc_size,
+                  return cargo::make_unexpected(CL_INVALID_BUFFER_SIZE));
+        OCL_CHECK(alignment > device_align,
```
coldav, line 492: Does this check and the line 482 check mean alignment == 0 won't get to line 499+?

hvdijk (approved):

coldav (approved):

PR #254, opened 2023-12-13
[vecz] Provide more context when CFG conversion fails

frasercrmck:
This should help debugging (using `--pass-remarks-missed=vecz`) why control-flow conversion failed to apply masks to the CFG.  The previous diagnostics would only print the name of the function that couldn't be converted, but not any more specific information.

This commit adds an extra level of information via a 'note', which is optionally printed on the line after the main diagnostic.

This is not intended to be a full solution to better vecz diagnostics, but a good first step.

### DIFF
--- a/modules/compiler/vecz/source/debugging.cpp
+++ b/modules/compiler/vecz/source/debugging.cpp
@@ -26,31 +26,39 @@ namespace vecz {
 ///
 /// @param[in] V The value (can be `nullptr`) to be included in the remark
 /// @param[in] Msg The main remark message
+/// @param[in] Note An optional additional note to provide more context/info.
 /// @return The remark message as it is to be printed
-static std::string createRemarkMessage(const Value *V, StringRef Msg) {
+static std::string createRemarkMessage(const Value *V, StringRef Msg,
+                                       StringRef Note = "") {
   std::string helper_str("Vecz: ");
   raw_string_ostream helper_stream(helper_str);
   helper_stream << Msg;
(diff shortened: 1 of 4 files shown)


RossBrunton (approved):

hvdijk (approved):

modules/compiler/vecz/source/debugging.cpp
```diff
@@ -26,31 +26,39 @@ namespace vecz {
...
+static std::string createRemarkMessage(const Value *V, StringRef Msg,
+                                       StringRef Note = "") {
   std::string helper_str("Vecz: ");
   raw_string_ostream helper_stream(helper_str);
   helper_stream << Msg;
   if (V) {
     if (isa<Instruction>(V)) {
       // Instructions are already prefixed by two spaces when printed
-      V->print(helper_stream, true);
+      V->print(helper_stream, /*IsForDebug*/ true);
```
hvdijk, line 39: The style used in [LLVM Coding Standards](https://llvm.org/docs/CodingStandards.html#comment-formatting) is `/*IsForDebug=*/true`, which we use in a few places already as well. Do we want to just stick with that?

coldav (approved):

modules/compiler/vecz/source/debugging.cpp
frasercrmck: Huh, I hadn't noticed that. Thanks, I've updated to use the proper style.

PR #256, opened 2023-12-14
[vecz] Run SROA as part of the vecz pipeline

frasercrmck:
Some kernels, particularly SYCL kernels, can make heavy use of temporary allocations which are passed between functions.

In SYCL's case, this is particularly harmful because the state stored and reloaded via these structs includes the work-item and work-group builtins (local IDs, global IDs, etc) which means the vectorizer is unable to analyze the program, leading to incorrect uniformity/divergence decisions and/or poor code generation.

LLVM's SROA is typically unable to do anything with this pattern before inlining, as the allocation is escaped by the called function. However, after vecz inlines these functions, SROA can eliminate much of this and present the vectorizer with a nicely analyzable program.

Some LIT tests using allocas were updated to use a more specific vecz pipeline which doesn't run SROA, as the allocas were being removed.


modules/compiler/vecz/source/vecz_pass_builder.cpp
```diff
@@ -181,7 +182,17 @@ bool vecz::buildPassPipeline(ModulePassManager &PM) {
...
   PM.addPass(createModuleToFunctionPassAdaptor(FixIrreduciblePass()));
 
-  // We have to run LLVM's Mem2Reg pass in case the front end didn't
+  // It's helpful to run SROA in case it opens up more opportunities to
+  // eliminate aggregates in (particularly SYCL) kernels. This is especially
+  // true after inlining - which we've (usually) just performed in the
+  // BuiltinInliningPass - because otherwise SROA is unable to analyze the
+  // lifetime of allocas due to them being "escaped" by the function callee.
+  PM.addPass(
+      createModuleToFunctionPassAdaptor(SROAPass(SROAOptions::ModifyCFG)));
```
hvdijk, line 191: This isn't a new issue, but we have a whole bunch of function passes in a row all wrapped in `createModuleToFunctionPassAdaptor`. It would be better to add them all to a single `FunctionPassManager` and wrap that, which would allow analysis results to be reused. Is that something that would make sense to change now, or is that better to leave as a separate PR for later at some point?

hvdijk (approved):

modules/compiler/vecz/source/vecz_pass_builder.cpp
frasercrmck: Oh yeah good point, that's been bugging me for a while. I can open a follow-up PR after this is merged (I'm already half way through testing)?

hvdijk: Sounds good to me, thanks.

coldav (approved):

PR #260, opened 2023-12-14
Revert "Fixed threading issues with usm and kernel execution"

coldav:
This reverts commit 695309df35a12a6bf23160327b51182e04d040ac.

# Overview

The fix for the usm threading issue introduces another threading issue, and so need reverting

# Reason for change

Tsan build showed possible threading issues from putting a context mutex around a queue mutex as there are other cases where a context mutex is called under a queue mutex, in this case when we delete a program.

# Description of change

revert 695309df35a12a6bf23160327b51182e04d040ac


hvdijk (approved):

frasercrmck (approved):
It can be good to include the reason in the revert commit message itself, but I won't hold this up on that account.

PR #262, opened 2023-12-18
[NFC] Silence CMake.

hvdijk:
# Overview

[NFC] Silence CMake.

# Reason for change

When I renamed Common.h to Common.h.in, I left a reference to include/Common.h untouched which resulted in a CMake warning:

  CMake Warning (dev) at cmake/AddCA.cmake:543 (add_executable):
    Policy CMP0115 is not set: Source file extensions must be explicit.  Run
    "cmake --help-policy CMP0115" for policy details.  Use the cmake_policy
    command to set the policy and suppress this warning.

    File:

      oneapi-construction-kit/source/cl/test/UnitCL/include/Common.h.in
  Call Stack (most recent call first):
    source/cl/cmake/AddCACL.cmake:92 (add_ca_executable)
    source/cl/test/UnitCL/CMakeLists.txt:59 (add_ca_cl_executable)
  This warning is for project developers.  Use -Wno-dev to suppress it.

# Description of change

Like Device.h, the reference should be explicitly to {CMAKE_CURRENT_BINARY_DIR}/include/Common.h.

# Checklist

### DIFF
--- a/source/cl/test/UnitCL/CMakeLists.txt
+++ b/source/cl/test/UnitCL/CMakeLists.txt
@@ -58,7 +58,7 @@ configure_file(
 
 add_ca_cl_executable(UnitCL
   # General testing machinery
-  include/Common.h source/Common.cpp
+  ${CMAKE_CURRENT_BINARY_DIR}/include/Common.h source/Common.cpp
   include/EventWaitList.h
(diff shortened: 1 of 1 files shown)


frasercrmck (approved):
Thanks! I only noticed this this morning but didn't look further.

PR #264, opened 2023-12-18
[vecz] Add support for masking atomic RMW instructions

frasercrmck:
This commit allows the vectorizer to vectorize kernels in which there are atomic RMW instructions that need masked for control-flow purposes: in a divergent if/else or a loop, etc.

It follows a fairly simple paradigm - similar to how we mask loads and stores - involving:

* Control-flow conversion replacing the atomic with a call to an 'internal' vecz builtin
* The packetizer widening this builtin, and replacing the call with another call (with packetized arguments)
* The post-vectorization `DefineBuiltinsPass` running and providing function bodies for these masked atomic builtins

The builtins themselves are simply loops over the entire vectorized arguments, conditionally doing an atomic operation one by one in sequence depending on the mask. This should be correct (i.e., not performing the whole atomic operation at once) since the results are undefined for how work-items run in parallel and which work-items would "win" if there was any contention in the atomic memory addresses. Note also that this is also essentially how plain atomics are vectorized: by scalarizing them.

There isn't yet support for the atomic cmpxhg instructions - those will be done separately.


modules/compiler/vecz/test/lit/llvm/masked_atomics.ll
```diff
@@ -0,0 +1,87 @@
...
+; CHECK-SAME: <4 x ptr> [[SPLAT_PTR]], <4 x i32> <i32 1, i32 1, i32 1, i32 1>, <4 x i1> [[CMP]]
+  %old0 = atomicrmw add ptr %p, i32 1 acquire
+; CHECK: = call <4 x i32> @__vecz_b_v4_masked_atomicrmw_add_align4_acquire_1_Dv4_u3ptrDv4_jDv4_b(
+; CHECK-SAME: <4 x ptr> [[VEC_PTR]], <4 x i32> <i32 1, i32 1, i32 1, i32 1>, <4 x i1> [[CMP]]
+  %old1 = atomicrmw add ptr %wi_p_i32, i32 1 acquire
+; CHECK: = call <4 x i32> @__vecz_b_v4_masked_atomicrmw_umin_align2_monotonic_1_Dv4_u3ptrDv4_jDv4_b(
+; CHECK-SAME: <4 x ptr> [[VEC_PTR]], <4 x i32> <i32 1, i32 1, i32 1, i32 1>, <4 x i1> [[CMP]]
+  %old2 = atomicrmw umin ptr %wi_p_i32, i32 1 monotonic, align 2
+; CHECK: = call <4 x float> @__vecz_b_v4_masked_atomicrmw_volatile_fmax_align4_seqcst_0_Dv4_u3ptrDv4_fDv4_b(
+; CHECK-SAME: <4 x ptr> [[VEC_PTR]], <4 x float> <float 1.000000e+00, float {{.*}}>, <4 x i1> [[CMP]]
```
hvdijk, line 45: The earlier checks all check that all elements of the vector are the same value. Why does this switch to checking `{{.*}}`?

frasercrmck: Good question, mostly just laziness I'm afraid. The lines were getting long. I've updated it now so it checks the vector elements are all the same value.

modules/compiler/vecz/source/vectorization_context.cpp
```diff
@@ -368,6 +374,227 @@ Function *VectorizationContext::getOrCreateMaskedFunction(CallInst *CI) {
...
+  } else {
+    return std::nullopt;
+  }
+
+  if (!FnName.consume_front("_align")) {
+    return std::nullopt;
+  }
+
+  uint64_t Alignment = 0;
+  if (FnName.consumeInteger(10, Alignment)) {
```
coldav, line 439: maybe a comment to say what 10 means here?

frasercrmck: Yep, good shout. It's just the radix.

coldav (approved):

modules/compiler/vecz/source/include/vectorization_context.h
```diff
@@ -150,14 +153,45 @@ class VectorizationContext {
...
+    llvm::ElementCount VF;
+    bool IsVectorPredicated = false;
+  };
+
+  /// @brief Check if the given function is a masked version of an atomic RMW
+  /// operation.
+  ///
+  /// @param[in] F The function to check
+  /// @return true if the function is a masked atomic RMW, or false otherwise
+  std::optional<MaskedAtomicRMW> isMaskedAtomicRMWFunction(
```
hvdijk, line 174: There is a mismatch between the documentation and the implementation here, the return type does not allow a `true`/`false` result.

frasercrmck: Thanks, updated!

modules/compiler/vecz/source/transform/control_flow_conversion_pass.cpp
```diff
@@ -1356,6 +1372,45 @@ bool ControlFlowConversionState::Impl::applyMaskToCall(CallInst *CI,
   return true;
 }
 
+bool ControlFlowConversionState::Impl::applyMaskToAtomicRMW(
```
hvdijk, line 1375: This function has a single `return` statement, `return true;`. There is no path where it returns that we cannot vectorize. Does this mean we can just return `void` and avoid having to deal with the possibility of failure in the caller, or is this anticipating future work?

hvdijk: Oh, hidden in the `VECZ_FAIL_IF` macro. Never mind.

PR #265, opened 2023-12-18
Use @llvm.sqrt.* intrincs rather than rolling our own.

hvdijk:
# Overview

Use @llvm.sqrt.* intrincs rather than rolling our own.

# Reason for change

All targets that we currently support have hardware floating point square root instructions that are more accurate than our own implementation.

# Description of change

points, etc.*

# Checklist


coldav (approved):

frasercrmck (approved):
Nice.

PR #266, opened 2023-12-19
[vecz] Add support for masking cmpxchg instructions

frasercrmck:
This finishes off the support for masked atomic instructions.

The scheme is essentially identical to that of atomic RMW instructions,
except that the instruction returns a literal struct containing the
value and a boolean success value. These must also be packetized for
efficient results.

The packetization of literal struct types - those unnamed structs
returned by cmpxchg - has gone through some refactoring, so now the
vectorized type of `{ i32, i1 }`  is (e.g.) `{ <4 x i32>, <4 x i1> }`.
This in practice makes it much more efficient to extract component
associated LIT tests. It also makes it possible to scalably vectorize
cmpxchg instructions.

Note that now even unmasked cmpxchg instructions generate the masked
builtin, albeit with an "all true" mask. This is to help maintain a
uniform internal representation of the packetized literal structure
type


modules/compiler/vecz/source/include/vectorization_context.h
```diff
@@ -153,36 +153,59 @@ class VectorizationContext {
...
-  struct MaskedAtomicRMW {
+  /// @brief Represents either an atomicrmw or cmpxchg operation.
+  ///
+  /// Most fields are shared, with the exception of CmpXchgFailureOrdering and
+  /// IsWeak, which are only to be set for cmpxchg, and BinOp, which is only to
+  /// be set to a valid value for atomicrmw.
+  struct MaskedAtomic {
     llvm::Type *PointerTy;
     llvm::Type *ValTy;
+    /// @brief Must be set to BAD_BINOP for cmpxchg instructions
```
coldav, line 164: It says above it should only be set to a valid value for atomicrmw?

frasercrmck: I'm taking `BAD_BINOP` to be an invalid binary operation. It you tried to create a `atomiccrmw` with `BAD_BINOP` you'd get an assert.

modules/compiler/vecz/source/transform/packetizer.cpp
```diff
@@ -2778,37 +2809,48 @@ ValuePacket Packetizer::Impl::packetizeMemOp(MemOp &op) {
...
-  Value *const ptr = CI.getArgOperand(0);
-  Value *const val = CI.getArgOperand(1);
-  Value *const mask = CI.getArgOperand(2);
+  bool const IsCmpXchg = AtomicInfo.isCmpXchg();
 
-  assert(AtomicInfo.ValTy == val->getType() && "AtomicInfo mismatch");
-  auto const packetWidth = getPacketWidthForType(val->getType());
+  Value *const ptrArg = CI.getArgOperand(0);
+  Value *const valOrCmpArg = CI.getArgOperand(1);
+  Value *const maskArg = CI.getArgOperand(2 + IsCmpXchg);
```
coldav, line 2820: personally I'd prefer `isCmpXchg ? 3 : 2` as being more clear, but feel free to ignore

modules/compiler/vecz/source/transform/packetizer.cpp
```diff
@@ -3716,6 +3805,68 @@ ValuePacket Packetizer::Impl::packetizeExtractElement(
...
+    return results;
+  }
+
+  Value *PackAggregate = packetizeIfVarying(Aggregate);
+  PACK_FAIL_IF(!PackAggregate);
+
+  Value *PackVal = packetizeIfVarying(Val);
+  PACK_FAIL_IF(!PackVal);
+
+  if (Aggregate == PackAggregate && Val != PackVal) {
```
coldav, line 3827: maybe clearer with 
```
IsAggregateVarying = Aggregate != PackAggregate;
IsValueVarying = Val != PackVal;
```

coldav (approved):

modules/compiler/vecz/source/transform/packetizer.cpp
frasercrmck: Good idea, thanks

modules/compiler/vecz/source/transform/packetizer.cpp
frasercrmck: Fair enough, yeah. I think I like this style well enough, though it's definitely subjective.

coldav: I think to me I think of a bool as true or false, and I need to think a little to see it as a 1, but yes feel free to leave as is.

PR #267, opened 2023-12-20
[compiler] Handle scalable structs as barrier live variables

frasercrmck:
The work-item loops pass would crash when faced with a live variable whose type was a struct containing scalable vectors.

We aren't legally allowed to store a struct type containing a mixture of scalable and fixed types (`{ <vscale x 1 x i8>, i8 }`) so we decompose such types into their constituent elements and store each individually. Note that scalable elements are stored in the scalable part of the live variables struct, and fixed elements are stored in the fixed part; in that way they are treated as if they were never a struct to begin with.

Note that there may be a future optimization possible here where we store all-scalable structs "whole", but this isn't currently a priority.

Note that this problem doesn't currently surface in the default pipeline in the main branch, because we only end up with scalable vectors when we vectorize as such, and we don't currently scalably vectorize any IR that's known to contain struct types, at least not in a way that creates a struct containing scalable vectors; see the new negative scalable vecz test as an example.

The plan is to start allowing this when we improve the vectorization of `cmpxhg` instructions. This should also improve the codegen for these structures; see the new fixed-length vecz test for an example of the poor codegen currently emitted.

### DIFF
--- a/modules/compiler/test/lit/passes/barriers-live-vars-literal-structs.ll
+++ b/modules/compiler/test/lit/passes/barriers-live-vars-literal-structs.ll
@@ -1,0 +1,119 @@
+; Copyright (C) Codeplay Software Limited
+;
+; Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+; Exceptions; you may not use this file except in compliance with the License.
+; You may obtain a copy of the License at
+;
+;     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+;
+; Unless required by applicable law or agreed to in writing, software
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; Run the verifier first, as working with scalable vectors in literal structs
+; can be fraught with illegality (as discovered when writing this test).
+; RUN: muxc --passes verify,work-item-loops,verify -S %s  | FileCheck %s
+
+target triple = "spir64-unknown-unknown"
+target datalayout = "e-p:64:64:64-m:e-i64:64-f80:128-n8:16:32:64-S128"
+
+; CHECK: %foo_live_mem_info = type { i8, [7 x i8], { <4 x i8>, i32, <8 x i1> }, [12 x i8], [0 x i8] }
+
+; Check that we can successfully save/reload fixed and scalable struct literals
+; across barriers. Scalable struct literals must be decomposed as it's invalid
+; to store them whole - they're "unsized". This might change in future versions
+; of LLVM.
+; CHECK: @foo.mux-barrier-region(ptr [[D:%.*]], ptr [[A:%.*]], ptr [[MEM:%.*]])
+; CHECK:  [[FIXED_ADDR:%.*]] = getelementptr inbounds %foo_live_mem_info, ptr [[MEM]], i32 0, i32 2
+; CHECK:  [[VSCALE:%.*]] = call i64 @llvm.vscale.i64()
+; CHECK:  [[NXV1I16_OFFS:%.*]] = mul i64 [[VSCALE]], 32
+; CHECK:  [[NXV1I16_ADDR:%.*]] = getelementptr inbounds %foo_live_mem_info, ptr [[MEM]], i32 0, i32 4, i64 [[NXV1I16_OFFS]]
+; CHECK:  [[NXV8I32_ADDR:%.*]] = getelementptr inbounds %foo_live_mem_info, ptr [[MEM]], i32 0, i32 4, i32 0
+; CHECK:  [[I8_ADDR:%.*]] = getelementptr inbounds %foo_live_mem_info, ptr [[MEM]], i32 0, i32 0
+; CHECK:  [[IDX:%.*]] = tail call i64 @__mux_get_global_id(i32 0)
+; We can store the fixed struct whole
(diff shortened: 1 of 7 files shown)


modules/compiler/test/lit/passes/barriers-live-vars-literal-structs.ll
```diff
@@ -0,0 +1,119 @@
...
+; Unless required by applicable law or agreed to in writing, software
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; Run the verifier first, as working with scalable vectors in literal structs
+; can be fraut with illegality (as discovered when writing this test).
```
coldav, line 18:
```suggestion
; can be fraught with illegality (as discovered when writing this test).
```

modules/compiler/utils/include/compiler/utils/barrier_regions.h
```diff
@@ -173,9 +176,13 @@ class Barrier {
...
 
-    /// @brief get a GEP instruction pointing to the given value in the barrier
-    /// struct.
-    llvm::Value *getGEP(const llvm::Value *live);
+    /// @brief get a GEP instruction pointing to the given value/idx pair in
+    /// the barrier struct.
+    llvm::Value *getGEP(const llvm::Value *live, unsigned member_idx = 0);
+
+    /// @brief Try and get a GEP instruction pointing to the given literal
+    /// stuct value in the barrier struct.
```
coldav, line 184:
```suggestion
    /// struct value in the barrier struct.
```

modules/compiler/utils/include/compiler/utils/barrier_regions.h
```diff
@@ -173,9 +176,13 @@ class Barrier {
...
         : barrier(b), gepBuilder(bb), barrier_struct(s) {}
 
-    /// @brief get a GEP instruction pointing to the given value in the barrier
-    /// struct.
-    llvm::Value *getGEP(const llvm::Value *live);
+    /// @brief get a GEP instruction pointing to the given value/idx pair in
+    /// the barrier struct.
+    llvm::Value *getGEP(const llvm::Value *live, unsigned member_idx = 0);
+
+    /// @brief Try and get a GEP instruction pointing to the given literal
```
coldav, line 183: When it says Try - does it return nullptr if it cannot? Is that an expected possibility or an error?

modules/compiler/utils/source/barrier_regions.cpp
```diff
@@ -855,9 +892,16 @@ void compiler::utils::Barrier::MakeLiveVariableMemType() {
   const auto &dl = module_.getDataLayout();
 
   struct member_info {
+    /// @brief The root `value` being stored.
     Value *value;
+    /// @brief The member index of this member inside `value`, if `value` is a
+    /// decomposed structure type. Zero otherwise.
```
coldav, line 898: is zero not a valid index or did you just want to come up with a good default value (although these values don't have defaults here).

modules/compiler/utils/source/barrier_regions.cpp
```diff
@@ -879,23 +923,33 @@ void compiler::utils::Barrier::MakeLiveVariableMemType() {
...
-    barrier_members.push_back({i, member_ty, alignment, size});
+      // Need to ensure that alloc alignment or preferred alignment is kept
+      // in the new struct so pad as necessary.
+      unsigned size = dl.getTypeAllocSize(member_ty_fixed);
+      alignment = std::max(dl.getPrefTypeAlign(ty).value(),
+                           static_cast<AlignIntTy>(alignment));
+      max_live_var_alignment = std::max(alignment, max_live_var_alignment);
+
+      barrier_members.push_back(
+          {i, static_cast<unsigned>(idx), ty, alignment, size});
```
coldav, line 951: since the loop is a bit  bigger, maybe a good time to rename `i` to something like `live_var`

coldav (approved):

modules/compiler/test/lit/passes/barriers-live-vars-literal-structs.ll
frasercrmck: And I thot I was good at spelling.

modules/compiler/utils/include/compiler/utils/barrier_regions.h
frasercrmck: Ta

modules/compiler/utils/source/barrier_regions.cpp
frasercrmck: Good idea, thanks

modules/compiler/utils/source/barrier_regions.cpp
frasercrmck: Zero is intended to be valid but it's rather arbitrary. It could also be `std::nullopt` but then we'd have to treat the regular and decomposed paths separately. By unifying the two code paths into iterating over a vector of types (the default case being a vector of one type), zero comes out naturally.

I was also kind of thinking of it like a representation of the address offset of the variable inside its parent (or itself), as in a struct. The pointer to the struct is the same address as the pointer to its first element.

modules/compiler/utils/include/compiler/utils/barrier_regions.h
frasercrmck: Returning `nullptr` is a possibility, as in `getGEP`, which this function just steals the `return nullptr` case from.

As for what it means for `getGEP` to return `nullptr`, I'm not sure. The APIs in this code aren't very well defined as they're all intended to be used internally. I think in practice `getGEP` is only used on variables that are contained within the live variables struct: variables contained within `live_vars_defs_in_kernel`, which are those contained within `whole_live_variables_set_`, which is also iterated over to insert the members into the structs. It's all a bit circuitous for my liking: both separate in terms of data and tracking, but also each are inextricably linked.

PR #269, opened 2023-12-20
[compiler] Remove the RemoveFencesPass

frasercrmck:
This pass was originally introduced back in LLVM 7 to account for the
fact that some of our targets couldn't handle the newly added `fence`
instruction. Removing them bypassed the issue.

While it is not semantically correct to remove fence instructions from
LLVM IR in general, we were relying on the fact that the memory model is
under-specified in OpenCL 1.2 and so could assume that fences were
redundant.

The ENORMOUS WARNING above this scheduling of this pass should be reason
enough not to run it. It has not been necessary for several LLVM
versions - it is claimed it is only in the host pipeline to ensure it's
tested. This is not a good enough reason to keep a dodgy unnecessary
pass.

There aren't enough good reasons to keep it around unused in the
codebase (the pass is trivial to implement by a downstream target if
they did require it) so this commit removes the pass.


hvdijk:
> Luckily we have better testing tools than that these days.

What better testing tool are we using that covers this pass? A quick grep for `remove-fences` or `RemoveFences` doesn't find it used anywhere else.

hvdijk:
Looking through the history, this was added at the end of 2018 when the current version of LLVM was LLVM 7, as a workaround for targets that were not yet updated to support this new instruction. Our minimal supported LLVM version now is LLVM 16. Do any targets still need this? If not, maybe we can remove the pass entirely?

frasercrmck:
> > Luckily we have better testing tools than that these days.
> 
> What better testing tool are we using that covers this pass? A quick grep for `remove-fences` or `RemoveFences` doesn't find it used anywhere else. A single lit test feels like it is not enough to see that the result actually works.

Yeah sorry, I wasn't very clear there. I meant more that `muxc` is sufficient to see that `RemoveFences` does indeed remove fences. Of course, it isn't sufficient to test that removing fences still passes OpenCL/SYCL conformance. I don't think it's valuable to actually test that though, as it's (almost certainly) not semantically correct to remove them. Even if it is correct (due to under-specification of the memory model) it's not something we need to do and I think it's bad form to keep on doing - it sends the wrong message to people using `host` as a reference.

> Looking through the history, this was added at the end of 2018 when the current version of LLVM was LLVM 7, as a workaround for targets that were not yet updated to support this new instruction. Our minimal supported LLVM version now is LLVM 16. Do any targets still need this? If not, maybe we can remove the pass entirely?

That's the other option I considered. We definitely don't need the pass. The current version of the PR was to keep the pass in case downstream targets did want it (either for working around backend issues or for performance) and to test the bare minimum, as I mentioned above. However, maybe this isn't a good enough reason to keep it. I'm open to removing it entirely.

hvdijk:
Personally, I would say that if a backend has not been updated to support `fence` instructions in the last 5 years, chances are, that is going to be the least of its issues. If the `fence` instructions are not needed for OpenCL 1.2, and if leaving them out improves performance for some target, what I think we should do instead is add an option to Clang to skip generating them in the first place, rather than generating them and then removing them. (But I also think we should only look at that if we have a concrete target that benefits from it.) As you say, removing `fence` instructions is not semantically correct at the LLVM IR level, so I would prefer to just remove the pass entirely.

frasercrmck:
> Personally, I would say that if a backend has not been updated to support `fence` instructions in the last 5 years, chances are, that is going to be the least of its issues. If the `fence` instructions are not needed for OpenCL 1.2, and if leaving them out improves performance for some target, what I think we should do instead is add an option to Clang to skip generating them in the first place, rather than generating them and then removing them. (But I also think we should only look at that if we have a concrete target that benefits from it.) As you say, removing `fence` instructions is not semantically correct at the LLVM IR level, so I would prefer to just remove the pass entirely.

Yep, agreed. I've updated the PR to remove the pass. Thanks :+1:

hvdijk (approved):

PR #273, opened 2024-01-02
[vecz] Fix and update a handful of LIT tests

frasercrmck:
These were all not working as intended for one of two reasons:

* The apparent `FileCheck` checks weren't working, due to a missing colon after `CHECK`
* The tests were marked `XFAIL` but were passing - an incorrect use of `FileCheck` was failing them instead.

Both of these cases were meaning we had tests which weren't checking what they were supposed to.

Each commit in the series explains a little better what's going on, but:

* Some tests can be fixed simply by adding the colon
* The one with `masked_printf` had different codegen, so was updated and simplified in the process.
* The `too_large_simdwith*` ones are in fact able to packetize/scalarize, so test for that.
* Some of the `call_instantiation_failure_*` ones were actually packetizing. They were all merged together, to save on the repeated test cases.


modules/compiler/vecz/test/lit/llvm/call_instantiation.ll
```diff
@@ -0,0 +1,160 @@
...
+}
+
+; If we can't duplicate a function, we can't packetize it.
+; CHECK-NOT: @__vecz_v4_cantduplicate
+define spir_kernel void @cantduplicate(ptr %in, ptr %out) {
+entry:
+  %call = tail call i64 @__mux_get_global_id(i32 0)
+  %arrayidx = getelementptr inbounds i32, ptr %in, i64 %call
+  %0 = load i32, ptr %arrayidx, align 4
+  %call1 = tail call spir_func i32 @_Z3clzi(i32 %0) #1
```
hvdijk, line 111: I'm aware you just took this from the existing test, but is there any reason why `clz` would have the `noduplicate` attribute applied to it? This seems like an unnecessarily confusing test. (Regardless of the answer, no need to update this as part of this PR.)

hvdijk (approved):

modules/compiler/vecz/test/lit/llvm/call_instantiation.ll
frasercrmck: I can't think of a reason, no. I couldn't work out if it *had* to be a builtin function for the test to work. I would think not, now, but maybe in the past we could only packetize known builtin functions?

PR #275, opened 2024-01-02
Fix boundary case in log2_extended_precision_half_safe.

hvdijk:
# Overview

Fix boundary case in log2_extended_precision_half_safe.

# Reason for change

log2_extended_precision_half_safe is a modified version of log2_extended_precision_half_unsafe that avoids denormal intermediate results via scaling. The unsafe version gives one input, 0x39f6, special treatment because the approximation is known to be incorrect, but the safe version did not translate this special treatment correctly: the special casing happens before downscaling, therefore the constant needs to be upscaled.

This fixes a pow test failure on FTZ targets.

# Description of change

One constant is changed. Tested by forcing this particular input to be used by modifying InputGenerator::half_edge_cases.

# Checklist

### DIFF
--- a/modules/compiler/builtins/abacus/include/abacus/internal/log2_extended_precision.h
+++ b/modules/compiler/builtins/abacus/include/abacus/internal/log2_extended_precision.h
@@ -280,7 +280,8 @@ T log2_extended_precision_half_safe(const T &x, T *ans_lo, T *hiExp, T *loExp) {
 
   // Single awkward boundary value fix:
   const SignedType edge(abacus::detail::cast::as<UnsignedType>(x) == 0x39f6);
-  remainder = __abacus_select(remainder, T(-0.000144362f16), edge);
+  // -0.14783 ==> -0.000144362 * 2^10
+  remainder = __abacus_select(remainder, T(-0.14783f16), edge);
 
   // Set return parameters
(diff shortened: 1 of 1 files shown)


modules/compiler/builtins/abacus/include/abacus/internal/log2_extended_precision.h
```diff
@@ -280,7 +280,8 @@ T log2_extended_precision_half_safe(const T &x, T *ans_lo, T *hiExp, T *loExp) {
 
   // Single awkward boundary value fix:
   const SignedType edge(abacus::detail::cast::as<UnsignedType>(x) == 0x39f6);
-  remainder = __abacus_select(remainder, T(-0.000144362f16), edge);
+  // -0.14783 ==> -0.000144362 * 2^10
```
coldav, line 283: Could we add a few comments here now you understand the boundary fix?

coldav (approved):

modules/compiler/builtins/abacus/include/abacus/internal/log2_extended_precision.h
hvdijk: We talked about this offline, going to try that as a followup PR.

PR #278, opened 2024-01-04
Improve UnitCL pow testing.

hvdijk:
# Overview

Improve UnitCL pow testing.

# Reason for change

We single out the value 0x39f6 in our implementation of log2_extended_precision, so in tests that use that function, we should ensure that that value is always tested.

# Description of change

This adds an overridable `GetEdgeCases()` function, defaulting to the `InputGenerator::half_edge_cases` we previously used, but overridden for `HalfMathBuiltinsPow` to additionally add `0x39f6` as an edge case.

Tested by locally reverting commit 90cf2faccb27adb19ce330bbe7c7ec62c53c7018 and verifying that UnitCL reliably picks up the error.

# Checklist

### DIFF
--- a/source/cl/test/UnitCL/include/kts/execution.h
+++ b/source/cl/test/UnitCL/include/kts/execution.h
@@ -347,6 +347,9 @@ class HalfParamExecution : public ExecutionWithParam<unsigned> {
   /// @brief Returns true if parameter at index has a scalar type
   /// @param[in] Index of parameter to check
   bool IsArgScalar(unsigned index) const;
+
+  /// @brief Returns a vector of edge cases that need extra testing.
+  virtual const std::vector<cl_ushort> &GetEdgeCases() const;
 };
 
 template <typename T>
--- a/source/cl/test/UnitCL/source/ktst_precision.cpp
+++ b/source/cl/test/UnitCL/source/ktst_precision.cpp
@@ -413,6 +413,17 @@ UCL_EXECUTION_TEST_SUITE_P(HalfOperatorTest, testing::Values(OPENCL_C),
 
 using HalfMathBuiltins = HalfParamExecution;
 
+struct HalfMathBuiltinsPow : HalfMathBuiltins {
+  const std::vector<cl_ushort> &GetEdgeCases() const override {
+    static const std::vector<cl_ushort> EdgeCases = [&] {
+      std::vector<cl_ushort> EdgeCases = HalfMathBuiltins::GetEdgeCases();
+      EdgeCases.push_back(0x39f6);
+      return EdgeCases;
+    }();
+    return EdgeCases;
+  }
+};
+
 TEST_P(HalfMathBuiltins, Precision_08_Half_Ldexp) {
   if (!UCL::hasHalfSupport(device)) {
     GTEST_SKIP();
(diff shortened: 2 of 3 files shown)


source/cl/test/UnitCL/source/ktst_precision.cpp
```diff
@@ -413,6 +413,17 @@ UCL_EXECUTION_TEST_SUITE_P(HalfOperatorTest, testing::Values(OPENCL_C),
 
 using HalfMathBuiltins = HalfParamExecution;
 
+struct HalfMathBuiltinsPow : HalfMathBuiltins {
+  const std::vector<cl_ushort> &GetEdgeCases() const override {
+    static std::vector<cl_ushort> EdgeCases = [&] {
```
frasercrmck, line 418: How come this is done in a lambda?

source/cl/test/UnitCL/include/kts/execution.h
```diff
@@ -347,6 +347,9 @@ class HalfParamExecution : public ExecutionWithParam<unsigned> {
   /// @brief Returns true if parameter at index has a scalar type
   /// @param[in] Index of parameter to check
   bool IsArgScalar(unsigned index) const;
+
+  /// @brief Returns a vector of edge cases that need extra testing.
+  virtual const std::vector<cl_ushort> &GetEdgeCases() const;
```
coldav, line 352: is it really worth doing as a reference? It could be prone to people introducing bugs. Would it not be simpler to just add the edge cases on request?

source/cl/test/UnitCL/source/ktst_precision.cpp
coldav: presumably to do it all in one static variable line.

coldav (approved):

source/cl/test/UnitCL/source/ktst_precision.cpp
hvdijk: Yes, this is to directly construct `EdgeCases` as the vector we want, rather than first constructing it and then adding an element, because this way we are guaranteed the initialization happens exactly once without needing extra guards.

source/cl/test/UnitCL/include/kts/execution.h
hvdijk: I don't quite get what you mean here, could you elaborate?

source/cl/test/UnitCL/source/ktst_precision.cpp
frasercrmck: Fair enough, yeah.

source/cl/test/UnitCL/include/kts/execution.h
coldav: I mean it can be called as is, it was more the reference, I'm just not sure that creating the vector at the time would really be noticeable time-wise

hvdijk: True, this isn't performance-critical. I did check though that the risk you mentioned, of accidentally returning a local variable, is detected by the compiler. It is, it results in "error: reference to local variable ‘EdgeCases’ returned". It would be a warning, but we build with `-Werror` anyway. Because of that, I think the risk of accidental wrong use is low enough that we don't need to worry about it.

coldav: I agree it does, feel free to merge as is.

PR #288, opened 2024-01-10
Use llvm.fma in abacus_fma

hvdijk:
# Overview

Use llvm.fma in abacus_fma

# Reason for change

Targets that have FMA instructions see a speedup when we use that instead of our custom `__abacus_fma` implementation. We know this helps ARM and RISC-V.

# Description of change

We add a new `__mux_hasnativefma()` builtin which, for now, only returns `true` on ARM and RISC-V (to be improved later). In `__abacus_fma`, if `__mux_hasnativefma()`, we use the `@llvm.fma` intrinsic, if not, we use our old implementation. This avoids problems on X86, where `@llvm.fma` needs libcalls which we do not expose.

# Checklist

### DIFF
--- a/modules/compiler/utils/source/replace_mux_math_decls_pass.cpp
+++ b/modules/compiler/utils/source/replace_mux_math_decls_pass.cpp
@@ -19,6 +19,7 @@
 #include <compiler/utils/replace_mux_math_decls_pass.h>
 #include <llvm/IR/IRBuilder.h>
 #include <llvm/IR/Module.h>
+#include <multi_llvm/triple.h>
 
 using namespace llvm;
 
@@ -26,7 +27,9 @@ PreservedAnalyses compiler::utils::ReplaceMuxMathDeclsPass::run(
     Module &M, ModuleAnalysisManager &MAM) {
   bool Changed = false;
 
+  Triple TT(M.getTargetTriple());
   auto &DI = MAM.getResult<DeviceInfoAnalysis>(M);
+
   // Abacus needs to know if denormal floats are supported, so it can avoid
   // Flush To Zero behaviour in intermediate operations of maths builtins.
   //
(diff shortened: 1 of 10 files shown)


modules/compiler/utils/source/replace_mux_math_decls_pass.cpp
```diff
@@ -51,15 +54,24 @@ PreservedAnalyses compiler::utils::ReplaceMuxMathDeclsPass::run(
...
   }
 
-  const std::array<std::pair<StringRef, const bool>, 3> MuxMathDecls = {
-      std::make_pair(MuxBuiltins::isftz, flush_denorms_to_zero),
-      std::make_pair(MuxBuiltins ::usefast, UseFast),
-      std::make_pair(MuxBuiltins::isembeddedprofile, is_embedded_profile)};
+  // "Native" FMA means we can translate @llvm.fma without libcall support.
+  // TODO We want to handle this for x86, when targetting a sufficiently recent
+  // CPU, too, but this information is not in the target triple.
+  bool has_native_fma = TT.isARM() || TT.isRISCV();
```
frasercrmck, line 60: How about `AArch64`? I don't see any reference to it in the code or in any comments.

hvdijk: Yep, should definitely add that, thanks.

modules/compiler/utils/source/replace_mux_math_decls_pass.cpp
```diff
@@ -51,15 +54,24 @@ PreservedAnalyses compiler::utils::ReplaceMuxMathDeclsPass::run(
...
+      {MuxBuiltins::usefast, UseFast},
+      {MuxBuiltins::isembeddedprofile, is_embedded_profile},
+      {MuxBuiltins::has_native_fma, has_native_fma},
+  };
 
   for (const auto &MathDecl : MuxMathDecls) {
     if (Function *const Fn = M.getFunction(MathDecl.first)) {
       Changed = true;
 
+      Fn->addFnAttr(Attribute::AlwaysInline);
```
frasercrmck, line 73: It's probably safest to wrap this in `if (!Fn->hasFnAttribute(Attribute::NoInline))` since the two options are incompatible and will cause a verification error if set simultaneously.

frasercrmck: Or strip the `noinline` attribute first, but I don't know if that's wise.

hvdijk: These builtins should never have `noinline`; if they do, something is wrong. I'll add a check so that if that ever happens, we diagnose that consistently.

modules/compiler/utils/source/replace_mux_math_decls_pass.cpp
hvdijk: Done.

modules/compiler/utils/source/replace_mux_math_decls_pass.cpp
hvdijk: Done, does that look okay to you?

frasercrmck: Yeah, looks fine to me, thanks.

I'm not sure I'd go as far to say something's "wrong" but I agree it's incredibly unlikely to come from our own in-tree code. There's nothing in our specs to say that downstream users aren't allowed to set `noinline` on those functions if they wanted to, though.

hvdijk: There's the general rule that identifiers starting with `__` are reserved for implementation use. That applies even to OpenCL, and that cannot work. It's similar to what would happen if a user defines `__mux_isftz` as returning `void`, in my opinion.

frasercrmck: That's true, though one could argue that a downstream user using OCK components is their own implementation? Either way, without us specifying these functions in our own specification, there's still room for maneuver. We could define that these builtins cannot be `noinline` if we wanted to.

hvdijk: Yeah, I think we need to: removing `noinline` would be changing the semantics, and leaving `noinline` would mean things break (if this is `noinline`, callers can't be optimized, and if callers aren't optimized, we end up generating libcalls on targets that have no hardware FMA). The only thing we can do is make it an error.

hvdijk: We already say "Later generic optimization passes, such as Dead Code Elimination, should be able
remove the unused control-flow in kernel code once the definitions of these builtins have been inlined." So implicitly, I think we already say these should not be declared `noinline`. I can update this PR to say so explicitly.

hvdijk: Done.

hvdijk: I think, but am not 100% sure, that ISel also does dead code elimination, so I think we should be fine as far as this is concerned even without `GlobalOptPass`. I'll check.

hvdijk: It turns out we do get an error without `GlobalOptPass`. Let me see if there is a quick way around that.

frasercrmck: Great, thanks for looking into that. I do hope to remove the problem with `.old` functions at some point!

frasercrmck (approved):
LGTM

PR #289, opened 2024-01-10
[compiler] Revert use of FMA in horner polynomial

frasercrmck:
This is unnecessarily precise for float and double types, and causes a lot of code bloat in the builtins library, which slowed down benchmarks.

We hope to improve the handling and performance of `__abacus_fma` in the short term, so we might revisit this when we've done that.

### DIFF
--- a/modules/compiler/builtins/abacus/include/abacus/internal/horner_polynomial.h
+++ b/modules/compiler/builtins/abacus/include/abacus/internal/horner_polynomial.h
@@ -27,7 +27,12 @@ inline T horner_polynomial(const T x, const TCoef *p_coef, size_t N) {
   T coef_sum = T(p_coef[N - 1]);
 
   for (size_t n = N - 1; n > 0; n--) {
-    coef_sum = __abacus_fma(coef_sum, x, T(p_coef[n - 1]));
+    if constexpr (sizeof(typename TypeTraits<T>::ElementType) == 2) {
+      // For half, we need the precision of FMA
+      coef_sum = __abacus_fma(coef_sum, x, T(p_coef[n - 1]));
+    } else {
+      coef_sum = T(p_coef[n - 1]) + x * coef_sum;
+    }
   }
 
   return coef_sum;


coldav (approved):

modules/compiler/builtins/abacus/include/abacus/internal/horner_polynomial.h
```diff
@@ -20,14 +20,33 @@
 #include <abacus/abacus_config.h>
 #include <abacus/abacus_type_traits.h>
 
+namespace {
+template <typename T, typename E = typename TypeTraits<T>::ElementType>
+struct fmuladd_helper;
+
+#ifdef __CA_BUILTINS_HALF_SUPPORT
+template <typename T>
+struct fmuladd_helper<T, abacus_half> {
```
hvdijk, line 29: [`fmuladd`](https://llvm.org/docs/LangRef.html#llvm-fmuladd-intrinsic) is the name of an LLVM intrinsic that specifically does not guarantee one way or the other whether it is implemented as `mul+add` or `fma`, but for half-precision, we need `fma`, I think it's confusing to use naming that indicates it might not be `fma`.

hvdijk: I think we can avoid the helper entirely, actually. Now that we build in C++17 mode, we can use `if constexpr`.

frasercrmck: Good point, thanks. I was unsure specifically about whether for half-precision we needed an FMA or just needed a multiply+add with extra precision (though not necessarily exactly FMA-level precision).

Do you have a suggestion for a better name? Would "multiply_add" clear up confusion between that and `llvm.fmuladd`? Or would that still hide the fact we need FMA?

frasercrmck: > I think we can avoid the helper entirely, actually. Now that we build in C++17 mode, we can use `if constexpr`.

I tried that, but was seemingly not allowed to include `type_traits` (I was trying to use `std::is_same`)

hvdijk: `if constexpr(sizeof(TypeTraits<T>::ElementType) == 2)` should work well enough for us. We could add more to `TypeTraits` to make this more readable too, if you prefer.

frasercrmck: Done

hvdijk (approved):

PR #291, opened 2024-01-10
Assume FMA support, bump x86-64 to x86-64-v3.

hvdijk:
# Overview

Assume FMA support, bump x86-64 to x86-64-v3.

# Reason for change

Our FMA implementation is slow and all the targets we aim to support at the moment have hardware support available.

# Description of change

This commit removes the support for targets that have no hardware FMA support; we always assume the availability of FMA now. For X86, hardware instructions are available from x86-64-v3 onwards, so when building for X86, set this as the minimum CPU level.

# Checklist


modules/compiler/targets/host/source/target.cpp
```diff
@@ -174,11 +175,13 @@ compiler::Result HostTarget::initWithBuiltins(
     if (host_device_info.half_capabilities) {
       FeatureMap["fp16"] = true;
     }
-  } else if (llvm::Triple::x86 == triple.getArch() && triple.isArch32Bit()) {
-    FeatureMap["sse2"] = true;
+  } else if (triple.isX86()) {
+    CPUName = "x86-64-v3";  // Default only, may be overridden below.
+    if (triple.isArch32Bit()) {
+      FeatureMap["sse2"] = true;
```
hvdijk, line 181: Do we still need this or would this already by implied by `x86-64-v3` anyway?

hvdijk: Testing shows we no longer need this, removed.

CHANGELOG.md
```diff
@@ -25,6 +25,12 @@ Upgrade guidance:
   IV names via a second `CreateLoopOpts` field.
 * Support for LLVM versions is now limited to LLVM 16 and LLVM 17. Support for
   earlier LLVM versions has been removed.
+* Support for FMA (fused multiply-add) is required. For x86-64, this means only
```
coldav, line 28: We should say this is only for the target accelerator and for  x86-64 this is only for the host target, where the cpu is used as an accelerator. General uses of x86 as a "host" can be any version.

hvdijk: Good point, thanks. Edited.

frasercrmck (approved):

PR #297, opened 2024-01-16
[NFC] Pick a style for const.

hvdijk:
# Overview

[NFC] Pick a style for const.

# Reason for change

This PR prepares for updating our clang-tidy, which wants us to add many more instances of const.

# Description of change

In cases where const may appear before or after the type, LLVM style is to place it before the type, and we have many more instances ourselves where we place it before the type than we do where we place it after. This commit sets that style in .clang-format and does a reformat across the code base.

I have no opinion on whether `const` is better to place before or after the type. If we want it after the type, consistently, I can update this PR to do so.

# Checklist

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -11,5 +11,6 @@ IncludeCategories:
   - Regex:           '.*'
     Priority:        3
 DerivePointerAlignment: false
+QualifierAlignment: Left
 PointerAlignment: Right
 ...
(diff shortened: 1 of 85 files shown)


frasercrmck (approved):
Thanks! Aligning ourselves with LLVM is the best direction for me. I don't personally prefer either style though.

PR #298, opened 2024-01-16
USM testing rework for better USM support

RossBrunton:
# Overview
This is a small rework of the USM testing framework to make it easier to add new USM memory types in the future, and also adds shared USM support to many tests.

# Reason for change

USM shared memory testing was lacking, this improves it by adding USM testing to most tests that currently do both device and host tests. It also removes duplication in the test suite itself.

# Description of change
This change brings a lot of duplicated code into
`cl_intel_unified_shared_memory_Test` (the base class of many tests). Specifically, it provides properties to check host and shared memory USM support, and helper functions to generate USM pointers.

One of the design changes was adding a way to iterate through all available USM pointer types without having to copy-paste code. This is used to add shared allocations to many places alongside device and host allocations.

This results in the tests doing more work, but this only causes a 10% reduction in the time taken for testing, which is about 300ms on my system.

N/A

# Checklist

### DIFF
--- a/source/cl/test/UnitCL/include/cl_intel_unified_shared_memory.h
+++ b/source/cl/test/UnitCL/include/cl_intel_unified_shared_memory.h
@@ -17,6 +17,8 @@
 #define UNITCL_USM_H_INCLUDED
 
 #include <CL/cl_ext.h>
+#include <cargo/error.h>
+#include <cargo/small_vector.h>
 
 #include "Common.h"
 
@@ -29,6 +31,13 @@ struct cl_intel_unified_shared_memory_Test : public virtual ucl::ContextTest {
     if (!isDeviceExtensionSupported("cl_intel_unified_shared_memory")) {
       GTEST_SKIP();
     }
+
+    ASSERT_SUCCESS(clGetDeviceInfo(
+        device, CL_DEVICE_HOST_MEM_CAPABILITIES_INTEL,
+        sizeof(host_capabilities), &host_capabilities, nullptr));
+    ASSERT_SUCCESS(clGetDeviceInfo(
+        device, CL_DEVICE_SINGLE_DEVICE_SHARED_MEM_CAPABILITIES_INTEL,
+        sizeof(shared_capabilities), &shared_capabilities, nullptr));
 
 #define CL_GET_EXTENSION_ADDRESS(FUNC)                            \
   FUNC = reinterpret_cast<FUNC##_fn>(                             \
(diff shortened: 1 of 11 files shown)


source/cl/test/UnitCL/include/cl_intel_unified_shared_memory.h
```diff
@@ -51,11 +60,116 @@ struct cl_intel_unified_shared_memory_Test : public virtual ucl::ContextTest {
...
+      cl_int err = clMemBlockingFreeINTEL(context, shared_ptr);
+      EXPECT_SUCCESS(err);
+      shared_ptr = nullptr;
+    }
+
+    if (host_ptr) {
+      cl_int err = clMemBlockingFreeINTEL(context, host_ptr);
+      EXPECT_SUCCESS(err);
+      host_ptr = nullptr;
+    }
```
hvdijk, line 80: What does `EXPECT_SUCCESS` do on error? If it aborts the process, or if it prints a warning and carries on, you can use your `allPointers()` helper function here.
```suggestion
    for (auto *ptr : allPointers()) {
      cl_int err = clMemBlockingFreeINTEL(context, device_ptr);
      EXPECT_SUCCESS(err);
    }
    device_ptr = shared_ptr = host_ptr = nullptr;
```

source/cl/test/UnitCL/include/cl_intel_unified_shared_memory.h
```diff
@@ -51,11 +60,116 @@ struct cl_intel_unified_shared_memory_Test : public virtual ucl::ContextTest {
...
+      case CL_MEM_TYPE_DEVICE_INTEL:
+        return clDeviceMemAllocINTEL(context, device, nullptr, size, alignment,
+                                     err);
+      case CL_MEM_TYPE_HOST_INTEL:
+        return clHostMemAllocINTEL(context, nullptr, size, alignment, err);
+      case CL_MEM_TYPE_SHARED_INTEL:
+        return clSharedMemAllocINTEL(context, device, nullptr, size, alignment,
+                                     err);
+      default:
+        return nullptr;
```
hvdijk, line 113: Can we ever get here? I would not expect returning `nullptr` to be useful here, but maybe I am not understanding what this default case handles.

source/cl/test/UnitCL/source/cl_intel_unified_shared_memory/usm_event_info.cpp
```diff
@@ -99,155 +77,128 @@ TEST_F(USMEventInfoTest, clEnqueueMemFillINTEL_EventInfo) {
...
-                               sizeof(cl_int), 0, nullptr, &events[1]);
-    EXPECT_SUCCESS(err);
-
-    ASSERT_SUCCESS(GetEventInfoHelper(
-        events[1], CL_EVENT_COMMAND_EXECUTION_STATUS, CL_QUEUED));
+  std::array<std::pair<void *, void *>, 4> pairs = {{
+      {host_ptr, device_ptr},
+      {host_ptr, shared_ptr},
+      {device_ptr, host_ptr},
+      {shared_ptr, host_ptr},
```
hvdijk, line 106: If it makes sense to test types in both orders (this has both `{host_ptr, device_ptr}` and `{device_ptr, host_ptr}`), does not make sense to test both orders for all types?

hvdijk, line 106: (Ignore my previous deleted comment here) Is it intentional that `{shared_ptr, device_ptr}` and the reverse are not in here?

source/cl/test/UnitCL/include/cl_intel_unified_shared_memory.h
RossBrunton: `EXPECT_*` functions continue the test while `ASSERT_*` functions halt it instantly (but only if the calling function has a specific return type). The helper function would work here, thanks.

source/cl/test/UnitCL/include/cl_intel_unified_shared_memory.h
RossBrunton: In theory we shouldn't, but we're in a test suite so anything can happen. I've put an explicit failure here.

source/cl/test/UnitCL/source/cl_intel_unified_shared_memory/usm_event_info.cpp
RossBrunton: It was, but only because for some reason I assumed you couldn't memcpy between shared and device pointers. Fixed now.

source/cl/test/UnitCL/include/cl_intel_unified_shared_memory.h
```diff
@@ -51,11 +60,109 @@ struct cl_intel_unified_shared_memory_Test : public virtual ucl::ContextTest {
...
     cl_char *byte_ptr = reinterpret_cast<cl_char *>(ptr) + offset;
     return reinterpret_cast<void *>(byte_ptr);
   }
 
+  /// Given a pointer, create a new USM allocation of the same type
+  ///
+  /// That is, if `old_ptr` points to a host USM allocation, this returns a new
+  /// host allocation. If the allocation type is unavailable (say, due to errors
+  /// querying the allocation data), nullptr is returned.
+  void *sameMemAlloc(void *old_ptr, size_t size, cl_uint alignment,
```
coldav, line 83: I think the name could be improved e.eg. sameMemTypeAlloc

source/cl/test/UnitCL/include/cl_intel_unified_shared_memory.h
```diff
@@ -51,11 +60,109 @@ struct cl_intel_unified_shared_memory_Test : public virtual ucl::ContextTest {
...
+                     cl_int *err) {
+    cl_int type_err;
+    cl_unified_shared_memory_type_intel alloc_type = CL_MEM_TYPE_UNKNOWN_INTEL;
+    type_err = clGetMemAllocInfoINTEL(context, old_ptr, CL_MEM_ALLOC_TYPE_INTEL,
```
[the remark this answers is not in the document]
RossBrunton: I'm not sure I see the problem with the extra testing. We could add infrastructure to store the types of all pointers, but that feels like it would involve a lot more code and complexity (what if a pointer is allocated in the body of a test instead of the initPointers method? What about offset pointers?).

All our tests also "test" things like allocating and freeing memory, so I think another function isn't that big a deal.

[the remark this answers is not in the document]
coldav: I think the purer the test the better, so we can identify the issues better. Also, by adding API calls, you can make somethings pass. I appreciate there are some things we can't remove, but this is superflous.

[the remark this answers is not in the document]
RossBrunton: Have removed this function entirely, replacing it with `if/else`s. It's only used in one place, so this should be fine.

coldav (approved):

hvdijk (approved):

PR #300, opened 2024-01-18
[compiler] Fix unchecked std::optional accesses

frasercrmck:
This cleans the project of the `bugprone-unchecked-optional-access` check which is present in newer versions of clang-tidy.

Some of these *are* checked as far as we're concerned, but clang-tidy can't know that that. Others are genuine oversights.

### DIFF
--- a/modules/compiler/spirv-ll/source/builder_debug_info.cpp
+++ b/modules/compiler/spirv-ll/source/builder_debug_info.cpp
@@ -470,12 +470,12 @@ llvm::Expected<llvm::MDNode *> DebugInfoBuilder::translate(
     return std::move(err);
   }
   // Without a size, we can't create a type.
-  if (!size_or_error.get()) {
+  std::optional<uint64_t> size = size_or_error.get();
+  if (!size) {
     return nullptr;
   }
-  uint64_t size = *size_or_error.get();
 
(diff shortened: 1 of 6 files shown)


modules/compiler/spirv-ll/source/builder_debug_info.cpp
```diff
@@ -1347,9 +1349,9 @@ llvm::Expected<llvm::MDNode *> DebugInfoBuilder::translate(
   llvm::DIType *const ty = ty_or_error.get();
 
   llvm::DIDerivedType *static_member_decl_ty = nullptr;
-  if (op->StaticMemberDecl()) {
+  if (auto static_member_decl = op->StaticMemberDecl(); static_member_decl) {
```
hvdijk, line 1352:
```suggestion
  if (auto static_member_decl = op->StaticMemberDecl()) {
```

hvdijk (approved):

coldav (approved):

modules/compiler/spirv-ll/source/builder_debug_info.cpp
frasercrmck: Ach yeah good spot, thanks!

PR #307, opened 2024-01-22
[compiler] Improve analysis of 'true uniform' values

frasercrmck:
This commit extends the Uniform Value Analysis with a fourth kind of uniformity: "true" uniformity. This represents a value that is uniform on both active and inactive lanes. The old class of "uniform" has been renamed to "active" uniformity to clarify this.

The analysis pass has been extended with a method to query whether a value is truly uniform. It processes this recursively on demand and caches the result. This is because the initial analysis run works from varying "roots" and marks all dependent values as varying - uniform values aren't handled at all. Rather than negatively affecting the performance of all Uniform Value Analysis runs, this on-demand method keeps costs the same except for users that need to query uniformity.

The query is still conservative, but less so. This can be seen in the test changes, where some cases which were previously conservatively handled as possibly varying/active uniform are now seen as truly uniform.

### DIFF
--- a/modules/compiler/vecz/source/analysis/uniform_value_analysis.cpp
+++ b/modules/compiler/vecz/source/analysis/uniform_value_analysis.cpp
@@ -18,6 +18,9 @@
 
 #include <compiler/utils/builtin_info.h>
 #include <compiler/utils/mangling.h>
+#include <llvm/Analysis/ValueTracking.h>
+#include <llvm/IR/Constants.h>
+#include <llvm/IR/InstrTypes.h>
 #include <llvm/IR/Instructions.h>
 #include <llvm/IR/Module.h>
 #include <llvm/Support/Debug.h>
@@ -76,6 +79,38 @@ bool isDivergenceReduction(const Function &F) {
           L.Consume("divergence_"));
 }
 
+bool isTrueUniformInternal(const Value *V, unsigned Depth) {
+  if (!V) {
+    return false;
+  }
+
+  // Constants and Arguments that can't be undef/poison are truly uniform
+  if (isa<Constant>(V) || isa<Argument>(V)) {
+    return isGuaranteedNotToBePoison(V);
+  }
+
+  constexpr unsigned DepthLimit = 6;
+
+  if (Depth < DepthLimit) {
+    // For a specific subset of instructions, if all operands are truly
+    // uniform, then the instruction is too.
+    // FIXME: This is pessimistic. We could improve this by extending the list
+    // of instructions covered. We could also use flow-sensitive analysis in
+    // isGuaranteedNotToBePoison to enhance its capabilities.
+    if (const auto *I = dyn_cast<Instruction>(V)) {
+      if (isa<UnaryOperator>(I) || isa<BinaryOperator>(I) || isa<CastInst>(I) ||
+          isa<CmpInst>(I) || isa<SelectInst>(I) || isa<PHINode>(I)) {
+        return isGuaranteedNotToBePoison(I) &&
+               llvm::all_of(I->operands(), [Depth](Value *Op) {
+                 return isTrueUniformInternal(Op, Depth + 1);
+               });
+      }
+    }
+  }
+
+  return false;
+}
+
 }  // namespace
 
 UniformValueResult::UniformValueResult(Function &F, VectorizationUnit &vu)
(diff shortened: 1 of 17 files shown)


modules/compiler/vecz/source/analysis/uniform_value_analysis.cpp
```diff
@@ -76,6 +77,32 @@ bool isDivergenceReduction(const Function &F) {
...
+  constexpr unsigned DepthLimit = 6;
+
+  if (Depth < DepthLimit) {
+    // For a specific subset of instructions, if all operands are truly
+    // uniform, then the instruction is too.
+    // FIXME: This is pessimistic.
+    if (const auto *I = dyn_cast<Instruction>(V)) {
+      if (isa<UnaryOperator>(I) || isa<BinaryOperator>(I) || isa<CastInst>(I) ||
+          isa<CmpInst>(I) || isa<FreezeInst>(I) || isa<SelectInst>(I) ||
+          isa<PHINode>(I)) {
```
hvdijk, line 95: This looks suspicious. `freeze` is used to handle `undef` and `poison`, but if we take `undef` and `poison` into account, is the logic still correct for any instructions? `freeze undef` and `freeze undef` are not guaranteed to be equal.

frasercrmck: Hmm you're probably right. I was initially thinking that each instance of `freeze` would be uniform across all _lanes_, even if it wasn't the exact same value between individual instances of the `freeze` instruction. On second though, I'm not sure this holds: once vectorized, `freeze <4 x i32> undef` isn't guaranteed to return the same value in all "lanes". It probably does, but we can't rely on it.

I'm not 100% certain that this holds. The problem is that the slow SYCL-DNN benchmark starts off with a `freeze` of an incoming argument. Maybe we can special case `freeze` to handle certain specific cases, I dunno. If not, then we're in a bit of trouble.

What do you think?

hvdijk: I think the logic holds if we don't have `undef` operands and the instructions do not produce `poison` or `undef` either. Clang will often add `noundef` to incoming arguments. Maybe checking that will be good enough?

modules/compiler/vecz/source/analysis/uniform_value_analysis.cpp
```diff
@@ -76,6 +77,32 @@ bool isDivergenceReduction(const Function &F) {
...
+  if (isa<Argument>(V) || isa<Constant>(V)) {
+    return true;
+  }
+
+  constexpr unsigned DepthLimit = 6;
+
+  if (Depth < DepthLimit) {
+    // For a specific subset of instructions, if all operands are truly
+    // uniform, then the instruction is too.
+    // FIXME: This is pessimistic.
```
coldav, line 91: I wonder why this is pessimistic - could you give an example/explanation here?

modules/compiler/vecz/source/analysis/uniform_value_analysis.cpp
frasercrmck: Yeah that's probably the way to go. 

Since we only check arguments and constants as the "leaves" we're quite easily able to prevent `undef`, `poison` or arguments without `noundef`.

Of course, we do still have to ensure that we don't end up seeing a `udiv 1, 0` or whatever. We shouldn't rely on expecting LLVM to have already replaced those with `poison`. I know there are ways of getting LLVM to tell us whether something is guaranteed not to produce poison/undef but I'll have to see if they're readily available.

modules/compiler/vecz/source/analysis/uniform_value_analysis.cpp
frasercrmck: "conservative" might be a better term. Just because we're not handling things like intrinsics, calls, loads, stores, allocas, vector insert/extract, etc. It's hard to get 100% full coverage here.

coldav: oh i see it's the specific subset that is pessimistic - maybe say that in the FIXME: This is pessimistic, improve by covering a larger set of instructions?

modules/compiler/vecz/source/analysis/uniform_value_analysis.cpp
frasercrmck: I think I've addressed this now @hvdijk. No more handling of freeze, and checking that everything we judge to be truly uniform is also guaranteed not to be undef or poison.

hvdijk: Looks good. We might also say, we do allow `freeze` instructions but only if the *operand* (as opposed to the result) is known not to be `undef` or `poison`. But not sure we have any test that would benefit from it.

hvdijk (approved):

modules/compiler/vecz/source/analysis/uniform_value_analysis.cpp
frasercrmck: Yes, that could also work. It would be interesting to find a test case for that. Most likely LLVM won't add a `freeze` to something it already knows is not poison/undef (at least from what I've seen in the sycl-dnn test I'm fixing with this) but it could happen through other means.

PR #311, opened 2024-01-23
Improve handling of error cases

RossBrunton:
Fix two issues I encountered while investigating something else.

# Overview
Fix two issues I encountered while investigating something else.

# Reason for change

Was working on something else to do with shared USM pointers, and noticed it not handling failing test cases gracefully.

# Description of change

Firstly, our tests don't initialize opencl events if they don't expect a failure. While in many cases, calls consuming these events shouldn't fail (and thus initialize the event correctly), it's possible that, due to a bug, they fail resulting in the event remaining uninitialized. If that event is then fed into functions later in the test, it's better for behaviour to be defined than undefined. We default initialise all opencl event pointers.

Secondly, if `MemFillImpl` gets into an unspecified state, it can return `CL_INVALID_COMMAND_QUEUE`. However, the manner in which it does this results in the event object not being properly tracked, resulting in an event that never terminates yet is depended on by an allocation. This results in a hang when deconstructing the USM pointer. I don't know if this issue can occur when running sound OpenCL code (it's possible if we don't have enough memory to allocate a slot in the USM allocation's event list). We ensure that the event is always marked as completed, even if the function exits early due to an error.

# Checklist


source/cl/source/extension/source/intel_unified_shared_memory/usm-exports.cpp
```diff
@@ -531,20 +531,24 @@ cl_int MemFillImpl(cl_command_queue command_queue, void *dst_ptr,
       // Push Mux fill buffer operation
       auto mux_error =
           ExamineUSMAlloc(usm_alloc, device, return_event, mux_buffer);
-      OCL_CHECK(mux_error, return CL_OUT_OF_RESOURCES);
+      OCL_CHECK(mux_error, return_event->complete(CL_OUT_OF_RESOURCES);
```
hvdijk, line 534: Other code here tries to handle the possibility that `return_event` is a null pointer, this probably needs to do the same. This all looks far too easy to get wrong, would be good to have a think about how we can refactor this to put all the `return_event->complete` calls in a single place, but if we can't find anything I'm happy with just plugging up all the places where we forgot.

source/cl/test/UnitCL/source/clEnqueueNDRangeKernel.cpp
```diff
@@ -526,7 +526,7 @@ TEST_F(clEnqueueNDRangeKernelTest, ConcurrentContextCreateAndEnqueue) {
         clCreateContext(nullptr, 1, &device, nullptr, nullptr, nullptr);
     cl_command_queue queue = clCreateCommandQueue(ctx, device, 0, nullptr);
 
-    std::vector<cl_event> kernel_complete_events;
+    std::vector<cl_event> kernel_complete_events{};
```
coldav, line 529: I think this has the same effect? We don't do this elsewhere in UnitCL.

hvdijk: You reverted one `std::vector` change in response to Colin's comment, but not the one he commented on, so I think this was resolved too soon. There is also another `std::vector` later in this file that did not need updating. (Mind you, it does the same thing, and if you had responded that you would prefer to keep them in for some reason, I'd also be fine with that.)

RossBrunton: Done.

hvdijk (approved):

PR #315, opened 2024-01-25
Add "deviceless" shared USM pointer

RossBrunton:
# Overview
Add a new pointer to the available pointers (used by allPointers()) for a "deviceless" shared USM pointer. This is a shared USM pointer without an associated device.

# Reason for change

Improves USM testing

# Description of change
Added a new pointer type to the pointers list.

The tests should all still pass.

# Checklist

### DIFF
--- a/source/cl/test/UnitCL/source/cl_intel_unified_shared_memory/usm_enqueue.cpp
+++ b/source/cl/test/UnitCL/source/cl_intel_unified_shared_memory/usm_enqueue.cpp
@@ -54,6 +54,18 @@ TEST_F(USMCommandQueueTest, MemFill_InvalidUsage) {
         {{0x0, 0x1, 0x2, 0x3, 0x4, 0x5, 0x6, 0x7, 0x8, 0x9, 0xA, 0xB, 0xC, 0xD,
           0xE, 0xF}}};
 
+    if (ptr == host_shared_ptr) {
+      // In our test suite, `host_shared_ptr` is a shared USM pointer with no
+      // associated device. This means that no device is able to read it (we
+      // currently don't support devices with cross-device access capability).
+      // The spec doesn't say what to return when the queue is for an incorrect
+      // device, so we return CL_INVALID_COMMAND_QUEUE.
+      cl_int err = clEnqueueMemFillINTEL(queue, ptr, pattern, sizeof(pattern),
+                                         sizeof(pattern), 0, nullptr, nullptr);
+      EXPECT_EQ_ERRCODE(err, CL_INVALID_COMMAND_QUEUE);
+      continue;
+    }
+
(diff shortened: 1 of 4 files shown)


coldav (approved):

source/cl/test/UnitCL/source/cl_intel_unified_shared_memory/usm_enqueue.cpp
```diff
@@ -47,6 +47,11 @@ struct USMCommandQueueTest : public cl_intel_unified_shared_memory_Test {
 // Test for invalid API usage of clEnqueueMemFillINTEL()
 TEST_F(USMCommandQueueTest, MemFill_InvalidUsage) {
   for (auto ptr : allPointers()) {
+    if (ptr == host_shared_ptr) {
+      // Shared pointers without devices can't be memfill'd
+      continue;
+    }
```
hvdijk, line 53: Since this test is about reporting the right errors for invalid uses of `MemFill`, maybe we should attempt it anyway and make sure it does not crash?

RossBrunton: Does the updated test and comment make sense? I've reworded it, but I'm not sure how understandable it is.

hvdijk: I'm confused now, I thought after our other conversation that this was meant to be valid, but we have a bug and don't support it and that's why we're skipping testing for now. When I look at the spec, I get that same impression. The spec doesn't say the queue needs to be associated with the allocation's device (you're right, there isn't any such device), the spec says the allocation needs to be accessible by the device associated with the queue (which does exist).

RossBrunton: Reworded the comment. host_shared_pointer has no associated device.

coldav (approved):

source/cl/test/UnitCL/source/cl_intel_unified_shared_memory/usm_enqueue.cpp
RossBrunton: > device is an optional OpenCL device ID to associate with the allocation. If device is NULL then the allocation is not associated with any device. Allocations with no associated device are accessible by the host and OpenCL devices in the context that have cross-device access capabilities.

None of our devices support the cross-device access capability (yet).

hvdijk: Now it's starting to make sense to me, thanks. Looks good, then.

hvdijk (approved):

PR #319, opened 2024-01-29
Improve error handling.

hvdijk:
# Overview

Improve error handling.

# Reason for change

This addresses a few warnings by clang-tidy-17.

# Description of change

clc's driver::saveBinary and oclc's oclc::Driver::BuildProgram and oclc::Driver::WriteToFile have a result that allows them to indicate failure, but several I/O functions were not checked for errors. Additionally, driver::saveBinary would close stdin when it had not opened it, which is not its responsibility.

# Checklist

### DIFF
--- a/source/cl/tools/clc/clc.cpp
+++ b/source/cl/tools/clc/clc.cpp
@@ -700,31 +700,34 @@ result driver::saveBinary() {
     binary = binary_storage;
   }
 
+  FILE *fp = nullptr;
+  bool owns_fp = false;
   if (output_file == "-") {
-    if (!dry_run) {
-      if (fwrite(binary.data(), 1, binary.size(), stdout) != binary.size()) {
-        perror("error: Could not write all of the binary to the output file");
-        return result::failure;
-      }
-      fflush(stdout);
-    }
-  } else {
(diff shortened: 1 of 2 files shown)


source/cl/tools/clc/clc.cpp
```diff
@@ -700,31 +700,34 @@ result driver::saveBinary() {
...
+    fp = stdin;
+  } else if (verbose) {
+    (void)std::fprintf(stderr, "info: writing binary to %s\n",
+                       generated_output_file.c_str());
+  }
+  if (!dry_run) {
+    if (fp == nullptr) {
+      fp = fopen(generated_output_file.c_str(), "wb");
+      if (fp == nullptr) {
+        (void)perror("error: Could not open output file");
```
RossBrunton, line 715: I know this is not an addition in this patch, but do we need the `error: ` in the perror string? If not, might it be worth removing it?

hvdijk: We are fairly consistent in prefixing messages with a severity (`info: ...`, `error: ...`), so I think we should keep it here.

RossBrunton (approved):

PR #320, opened 2024-01-29
Update findDbgDeclares for LLVM 18

PietroGhg:
# Overview

Updates uses of `findDbgDeclares` to account for upstream LLVM changes (https://github.com/llvm/llvm-project/commit/304119860ac0ded0a126ab1c8cc30367e29ebd01)

### DIFF
--- a/modules/compiler/utils/source/barrier_regions.cpp
+++ b/modules/compiler/utils/source/barrier_regions.cpp
@@ -971,8 +971,7 @@ void compiler::utils::Barrier::MakeLiveVariableMemType() {
     // Check if the alloca has a debug info source variable attached. If
     // so record this and the matching byte offset into the struct.
 #if LLVM_VERSION_GREATER_EQUAL(18, 0)
-    SmallVector<DbgDeclareInst *, 1> DbgIntrinsics;
-    findDbgDeclares(DbgIntrinsics, member.value);
+    auto DbgIntrinsics = findDbgDeclares(member.value);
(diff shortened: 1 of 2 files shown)


hvdijk (approved):

modules/compiler/utils/source/barrier_regions.cpp
```diff
@@ -971,8 +971,7 @@ void compiler::utils::Barrier::MakeLiveVariableMemType() {
     // Check if the alloca has a debug info source variable attached. If
     // so record this and the matching byte offset into the struct.
 #if LLVM_VERSION_GREATER_EQUAL(18, 0)
-    SmallVector<DbgDeclareInst *, 1> DbgIntrinsics;
-    findDbgDeclares(DbgIntrinsics, member.value);
+    auto DbgIntrinsics = findDbgDeclares(member.value);
```
coldav, line 974: Isn't this the same as >= 17? Can't we just remove the 18 part altogether here since auto resolve it all?

hvdijk: LLVM 17 has `auto DbgIntrinsics = FindDbgDeclareUses(member.value);` rather than `auto DbgIntrinsics = FindDbgDeclares(member.value);` (different function name).

coldav (approved):

modules/compiler/utils/source/barrier_regions.cpp
coldav: yes sorry I deleted my comment when i realised.

PR #323, opened 2024-01-30
Extend CA_ENABLE_TESTS to include check targets

coldav:
# Overview

Disable check targets when CA_ENABLE_TEST=false

# Reason for change

Remove target name conflict when used with SYCL Native CPU, also seems more correct

# Description of change

This extends the cmake flag to include check targets.

### DIFF
--- a/cmake/AddCA.cmake
+++ b/cmake/AddCA.cmake
@@ -685,8 +685,11 @@ macro(get_target_link_libraries variable target)
   get_target_link_libraries_recursive(${target})
 endmacro()
 
-# Add the check target to run all registered checks, see add_ca_check() below.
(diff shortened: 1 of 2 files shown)


cmake/AddCA.cmake
```diff
@@ -730,6 +732,9 @@ endif()
       can be specified, each must be of the form: "VAR=<value>"
 #]=======================================================================]
 function(add_ca_check name)
+  if (NOT CA_ENABLE_TESTS)
+    return()
+  endif()
```
hvdijk, line 737: This makes the function no longer match its documentation. I think the change makes sense if you update the documentation too though.

hvdijk (approved):

PR #324, opened 2024-01-30
Added $<BUILD_INTERFACE on target_includes

coldav:
This allows libraries to be included easier from sycl-native cpu.

# Overview

Added $<BUILD_INTERFACE on target_include_directories

# Reason for change

Simplifies including from sycl-native-cpu.

# Description of change

Added $<BUILD_INTERFACE on target_include_directories if they are from within the project.


modules/compiler/CMakeLists.txt
```diff
@@ -46,9 +46,9 @@ if(CA_RUNTIME_COMPILER_ENABLED)
     ${CMAKE_CURRENT_SOURCE_DIR}/library/source/library.cpp
     ${COMPILER_CONFIG_SOURCE})
   target_include_directories(compiler-static PUBLIC
-    ${CMAKE_CURRENT_SOURCE_DIR}/include
-    ${CMAKE_CURRENT_SOURCE_DIR}/utils/include
-    ${CMAKE_CURRENT_SOURCE_DIR}/library/include
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/utils/include>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/library/include>
```
hvdijk, line 51: Does it work if we do `$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include ${CMAKE_CURRENT_SOURCE_DIR}/utils/include ${CMAKE_CURRENT_SOURCE_DIR}/library/include>` rather than repeating `$<BUILD_INTERFACE:` every time? (Likewise for other `CMakeLists.txt` files.) If that doesn't work, then this looks fine to me as it is.

hvdijk (approved):

modules/compiler/CMakeLists.txt
coldav: it does seem to work.

coldav: actually it doesn't seem to work fully. I think I'd rather go with what we have,.

PR #325, opened 2024-01-31
Enable extra warnings, adjust code.

hvdijk:
# Overview

Enable extra warnings, adjust code.

# Reason for change

We would like to be able to build with -pedantic -Wcast-qual, the former because it flags accidental use of non-standard language constructs, the latter because it is also enabled by LLVM.

# Description of change

These are now enabled, and code is updated to address the warnings raised.
    
The C standard is bumped from C99 to C11 to account for the fact that we were already using C11 language features, which now got flagged by -pedantic. The move to C11 also enables the checkDeprecatedOrUnsafeBufferHandling warning, which tells us not to use scanf and to use scanf_s instead, despite scanf_s being optional and not supported on platforms we support, and despite the use of scanf being safe. Therefore, this warning is disabled.

GoogleTest is bumped to version 1.14.0. The old version of GoogleTest we were using uses macros in a non-C++17-conforming way. This has been improved (though not fully fixed) in more recent versions, requiring much fewer workarounds.
    
Unfortunately, we were using an internal GoogleTest function that we should have left alone. This function still exists, but is now static and uses a new type in an anonymous namespace, so cannot be referred to from external code without patching GoogleTest. We have locally patched GoogleTest before when it did not quite work for us, so this patches it again.

# Checklist


.clang-tidy
```diff
@@ -208,6 +208,7 @@ Checks: "bugprone-*,\
          -clang-analyzer-cplusplus.NewDelete,\
          -clang-analyzer-cplusplus.NewDeleteLeaks,\
          -clang-analyzer-optin.performance.Padding,\
+         -clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling,\
```
coldav, line 211: why has this been added?

coldav (approved):

.clang-tidy
hvdijk: It is a warning that was previously implicitly disabled because it's a C11+-only warning, it's bogus in that it suggests that calls that are safe are unsafe (all the `scanf`s it complains about have a format string that specifies no strings), and it's bogus in that the suggested replacement (`scanf_s`) is optional in C11 and not widely implemented (it's basically a Microsoft-only function).

PR #328, opened 2024-02-01
Include FMA in manual type legalization.

hvdijk:
# Overview

Include FMA in manual type legalization.

# Reason for change

It too is promoted incorrectly by current LLVM versions.

# Description of change

Current LLVM promotes half-precision FMA to single-precision, which is not sufficient to get a correct result. Promote it to double-precision instead.

# Checklist

### DIFF
--- a/modules/compiler/utils/source/manual_type_legalization_pass.cpp
+++ b/modules/compiler/utils/source/manual_type_legalization_pass.cpp
@@ -20,6 +20,7 @@
 #include <llvm/IR/IRBuilder.h>
 #include <llvm/IR/InstrTypes.h>
 #include <llvm/IR/Instructions.h>
+#include <llvm/IR/IntrinsicInst.h>
 #include <llvm/IR/Module.h>
 #include <llvm/IR/PassManager.h>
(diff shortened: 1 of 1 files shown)


modules/compiler/utils/source/manual_type_legalization_pass.cpp
```diff
@@ -31,25 +32,39 @@ using namespace llvm;
...
+  // implement it correctly at the moment.
+  //
+  // Both for operators and FMA, whether the target implements the operation
+  // correctly may depend on the target feature string. We ignore that here for
+  // simplicity.
   const llvm::Triple TT(F.getParent()->getTargetTriple());
-  if (TTI.isTypeLegal(HalfT) || TT.isX86() || TT.isRISCV()) {
+
+  auto &TTI = FAM.getResult<TargetIRAnalysis>(F);
+  const bool HaveCorrectHalfOps =
```
coldav, line 54: won't some ARM target support this? Also is RISCV a guarantee  if it doesn't have hardware feature?

RossBrunton (approved):

PR #329, opened 2024-02-01
Fix include path for compiler-binary-metadata

PietroGhg:
Fixes the include path for the `compiler-binary-metadata` CMake target for Native CPU


PietroGhg:
Closing this since this is better fixed with a change in DPC++

PR #330, opened 2024-02-01
Upgrade to clang-tidy-17.

hvdijk:
# Overview

Upgrade to clang-tidy-17.

# Reason for change

Our current version, clang-tidy-9, does not understand C++17 constructs and regardless, it will be nice to not be stuck on old versions for no real reason.

# Description of change

* misc-include-cleaner is disabled globally because at the moment, it requires annotations in external headers we have no control over.
* misc-use-anonymous-namespace is disabled globally because it goes against LLVM style, and the warning is triggered by use of LLVM macros that we have no control over.
* modernize-macro-to-enum is disabled globally because it also warns on macros that should be usable in preprocessor conditions.

Everything else either has the code adjusted according to the intent of the warning, or has the warning suppressed as appropriate. Warnings are suppressed for false positives (intentionally unused function return values, and two instances of clang-tidy warning about undefined code that is actually well-defined), and for code that we cannot improve without a bigger refactor (unused function return values in places where we have no way to indicate an error).

Two warnings that we had suppressed to work around a clang-tidy-9 limitation are no longer suppressed.

# Checklist

### DIFF
--- a/modules/compiler/utils/source/lld_linker.cpp
+++ b/modules/compiler/utils/source/lld_linker.cpp
@@ -55,16 +55,48 @@ Expected<std::unique_ptr<MemoryBuffer>> lldLinkToBinary(
     const ArrayRef<uint8_t> rawBinary, const std::string &linkerScriptStr,
     const uint8_t *linkerLib, unsigned int linkerLibBytes,
     const SmallVectorImpl<std::string> &additionalLinkArgs) {
-  SmallString<128> objFileName;
-  SmallString<128> elfFileName;
-  SmallString<128> linkerScriptFileName;
-  SmallString<128> linkRTFileName;
-  sys::fs::createTemporaryFile("lld", "o", objFileName);
-  sys::fs::createTemporaryFile("lld", "elf", elfFileName);
-  sys::fs::createTemporaryFile("lld", "ld", linkerScriptFileName);
+  struct TemporaryFile {
+    TemporaryFile() = default;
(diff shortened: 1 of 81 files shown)


modules/compiler/utils/source/lld_linker.cpp
```diff
@@ -59,11 +59,11 @@ Expected<std::unique_ptr<MemoryBuffer>> lldLinkToBinary(
   SmallString<128> elfFileName;
   SmallString<128> linkerScriptFileName;
   SmallString<128> linkRTFileName;
-  sys::fs::createTemporaryFile("lld", "o", objFileName);
-  sys::fs::createTemporaryFile("lld", "elf", elfFileName);
-  sys::fs::createTemporaryFile("lld", "ld", linkerScriptFileName);
+  (void)sys::fs::createTemporaryFile("lld", "o", objFileName);
```
coldav, line 62: I wonder whether we should just check for failure here - but it's no worse than before, so up to you.

hvdijk: Oh right, we can actually indicate failure here, will update.

coldav (approved):

RossBrunton (approved):

modules/compiler/utils/source/lld_linker.cpp
coldav: looks good

PR #334, opened 2024-02-05
Support LLVM already being set up in OCK rather than importing it.

coldav:
# Overview

Support LLVM already being set up in OCK rather than importing it.

# Reason for change

OCK previously required CA_LLVM_INSTALL_DIR to be set in order to import LLVM. If OCK is included from LLVM, then we can't do this.

# Description of change

In ImportLLVM.cmake, we check if LLVM CMake variables are set first. We also change to use the more general LLVM method of using add_include_dirs rather than adding target include for LLVM, as that has issues with using OCK through FetchContent (or as an external LLVM project).


cmake/ImportLLVM.cmake
```diff
@@ -39,6 +41,12 @@ Once the modules are included we verify the LLVM version is supported and set
 some additional compile definitions.
 #]=======================================================================]
 
+# If we already have the LLVM CMake variables set then we don't need to include from
+# CA_LLVM_INSTALL_DIR
+if (DEFINED LLVM_VERSION_MAJOR)
```
hvdijk, line 46: Clang also supports building either inside LLVM or outside, it checks it by `CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR`, would it work for us to do it the same way?

cmake/ImportLLVM.cmake
```diff
@@ -105,3 +113,6 @@ if(CMAKE_SIZEOF_VOID_P EQUAL 4 OR CA_BUILD_32_BITS OR
   add_definitions(-D_LARGEFILE_SOURCE)
   add_definitions(-D_FILE_OFFSET_BITS=64)
 endif()
+
+# Set LLVM Include directories for all targets
+include_directories(SYSTEM ${LLVM_INCLUDE_DIRS})
```
hvdijk, line 118: Does this mean we also get `LLVM_INCLUDE_DIRS` added for targets that previously did not have it? That's probably fine, but good to make sure.

cmake/ImportLLVM.cmake
coldav: no, because we promote doing this for building OCK customer targets, so we can't really make this check in the same way. Also, since it's not really an LLVM project as such it feels a bit general.

cmake/ImportLLVM.cmake
coldav: When you say targets I assume you mean cmake targets. If so, then yes it does - although we wouldn't link with it and it would fail then.

cmake/ImportLLVM.cmake
hvdijk: I don't quite get what you mean here, sorry. I know we support at least two ways of building OCK, one is from OCK itself, one is as part of DPC++ with the `native_cpu` plugin, but it sounds like those are not the only two ways?

modules/compiler/spirv-ll/CMakeLists.txt
```diff
@@ -68,7 +68,7 @@ target_include_directories(spirv-ll PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/source>)
 target_include_directories(spirv-ll SYSTEM PUBLIC
-  ${spirv-headers_SOURCE_DIR}/include ${LLVM_INCLUDE_DIR})
+  ${spirv-headers_SOURCE_DIR}/include)
```
frasercrmck, line 71: We could maybe just chuck this on a single line now

cmake/ImportLLVM.cmake
hvdijk: It would not necessarily fail: I've tried to use header-only LLVM functions before in places that don't and shouldn't have a dependency on LLVM. It is useful, though not absolutely required, for that to fail as soon as possible.

cmake/ImportLLVM.cmake
coldav: I think if it's already set anyway, we are likely to get into a messy cmake situation if we try to import it here, but I'm open to suggestions.

hvdijk: Thinking about this, is there some target that in-LLVM-tree builds will definitely have available, and out-of-LLVM-tree builds can be reasonably assumed won't even if they have imported LLVM? For instance, `if(TARGET llvm-tblgen)` except I think that one might actually also be available as an imported target. Trying to find something better...

coldav: The only reason I went with this is that we use LLVM_VERSION_MAJOR elsewhere. We do `if(TARGET LLVMCore)` - maybe something like that is guaranteed?

cmake/ImportLLVM.cmake
coldav: Fair enough in theory you could include some llvm, but you'd need to do it pretty knowingly via include <llvm/....>

PR #335, opened 2024-02-05
[Abacus] Reimplement ldexp.

hvdijk:
# Overview

[Abacus] Reimplement ldexp.

# Reason for change

Our ldexp implementation was overly complicated and failing to handle some corner cases.

# Description of change

points, etc.*

# Checklist


modules/compiler/builtins/abacus/source/abacus_math/ldexp.cpp
```diff
@@ -20,312 +20,119 @@
...
-  n_hi += exp_bias;
-  n_lo += exp_bias;
-  n_lo_lo += exp_bias;
+  const SignedType n_c = convert_sat<SignedType>(n);
 
-  const SignedType mant_bits = Shape::Mantissa();
-  T mul_hi = abacus::detail::cast::as<T>(n_hi << mant_bits);
-  T mul_lo = abacus::detail::cast::as<T>(n_lo << mant_bits);
-  T mul_lo_lo = abacus::detail::cast::as<T>(n_lo_lo << mant_bits);
+  // Split n into 3 such that ldexp(ldexp(x, n_1), n_1) does not turn normal
```
coldav, line 35: could we explain what split n into three really means and how the algorithm works, possibly with a link?

coldav: that's a good explanation, can you add that?

frasercrmck: Thanks for the explanation, makes sense (I was also unsure).

If you could add some version of this comment to the source code that'd be really helpful.

hvdijk: Expanded the comment, does it look okay like this?

modules/compiler/builtins/abacus/source/abacus_math/ldexp.cpp
```diff
@@ -20,312 +20,121 @@
...
+  // which requires avoiding turning normals into subnormals early.
+  const SignedType n_1_min =
+      SignedType(1) - (SignedType(((abacus::detail::cast::as<UnsignedType>(x) &
+                                    Shape::ExponentMask()) >>
+                                   Shape::Mantissa())) /
+                       2);
+  const SignedType n_1 = __abacus_max(SignedType(n_c / 3), n_1_min);
+  const SignedType n_2 = n_c - (n_1 << 1);
+
+  const UnsignedType biased_n_1 =
```
coldav, line 47: Is this the bit where it calculates pow(2,n_1) and pow(2,n_2) - I think just an extra line to say what you said before that ldexp -> x * pow(2,n); it just makes it easxier to understand without the full knowledge.

coldav (approved):

modules/compiler/builtins/abacus/source/abacus_math/ldexp.cpp
coldav: yes that is much better - I've added a comment below though just to add a bit more so it's obvious that ldexp -> x * pow(n)

modules/compiler/builtins/abacus/source/abacus_math/ldexp.cpp
hvdijk: I have been working on the assumption that someone working on a standard function would be expected to know what the function is supposed to do, as we tend to not repeat what is already directly in the relevant standards, and that `ldexp(x, n)` returns `x * pow(2, n)` is just its definition, not an implementation detail. I don't think we should single out `ldexp` in documenting things like that, but I don't mind if we want to change that as a general thing across Abacus. That said, agreed that the fact that these lines calculate `pow(2, n_1)` and `pow(2, n_2)` is not obvious, will add a comment about that too.

coldav: I appreciate this is a standard method, but it does make things easier to review etc, and allows for a quick understanding when there is an issue.  I agree that's part of it's definition though.

hvdijk: We talked about this off of GitHub, we do have that documentation at https://github.com/codeplaysoftware/oneapi-construction-kit/blob/f6e8b035978f25ff25cf265bc999104cadafb50d/modules/compiler/builtins/abacus/generate/abacus_math/main.in#L1014-L1022 but it's not in an easily accessible place. We'll work on improving that later.

PR #337, opened 2024-02-06
[compiler] Provide a safe GDB JIT registration listener

frasercrmck:
This commit adds a version of LLVM's GDBRegistrationListener that is safe to use in our libraries.

This and the original listener is responsible for registering and deregistering dynamically compiled kernels with GDB's JIT compilation interface. It does this through access to global symbols understood by GDB. This requires that it must be made correctly thread-safe.

LLVM ensures this by having only one static singleton version of its GDB listener, which locks itself when notified of a new object, or notified or a freed one. This does not work for our use-case, as the static singleton may be destroyed before the OpenCL library is destroyed. Thus our modules and kernels may inadvertently use an already freed object.

To solve this, our version of this listener is no longer a singleton, and instead requires the callers to correctly lock access to it. In principle this requires the host target to lock when jitting a kernel and when freeing a kernel. Both of these are now locked under the same LLVM global mutex to ensure thread safety.

The original also accesses these globals in its destructor, but our version requires (asserts) that resources are already freed up at the time the listener is freed. This requires us to lock in one fewer instance and should simplify the mental model of how the listener works.

### DIFF
--- a/modules/compiler/utils/include/compiler/utils/gdb_registration_listener.h
+++ b/modules/compiler/utils/include/compiler/utils/gdb_registration_listener.h
@@ -1,0 +1,27 @@
+// Copyright (C) Codeplay Software Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+// Exceptions; you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <memory>
+
+namespace llvm {
+class JITEventListener;
+}
+
+namespace compiler {
+namespace utils {
+std::unique_ptr<llvm::JITEventListener> createGDBRegistrationListener();
+}
+}  // namespace compiler
--- a/modules/compiler/utils/source/gdb_registration_listener.cpp
+++ b/modules/compiler/utils/source/gdb_registration_listener.cpp
@@ -1,0 +1,239 @@
+// Copyright (C) Codeplay Software Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+// Exceptions; you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


modules/compiler/utils/source/gdb_registration_listener.cpp
```diff
@@ -0,0 +1,238 @@
...
+// Debuggers puts a breakpoint in this function.
+extern "C" void __jit_debug_register_code();
+}
+
+namespace {
+
+// FIXME: lli aims to provide both, RuntimeDyld and JITLink, as the dynamic
+// loaders for it's JIT implementations. And they both offer debugging via the
+// GDB JIT interface, which builds on the two well-known symbol names below.
+// As these symbols must be unique accross the linked executable, we can only
```
coldav, line 76:
```suggestion
// As these symbols must be unique across the linked executable, we can only
```

coldav (approved):

modules/compiler/utils/include/compiler/utils/gdb_registration_listener.h
```diff
@@ -0,0 +1,26 @@
+//
```
hvdijk, line 1: Missing copyright notice

coldav (approved):

modules/compiler/utils/include/compiler/utils/gdb_registration_listener.h
frasercrmck: Good catch, thanks

modules/compiler/utils/source/gdb_registration_listener.cpp
frasercrmck: Thanks. That's in the original LLVM source code but it makes sense to fix locally. I'll push something upstream to LLVM to fix while I'm here.

hvdijk:
I'm a bit concerned that we rely on this working when LLVM globals are already destroyed. Do we know that none of what we call into access any globals?

frasercrmck:
> I'm a bit concerned that we rely on this working when LLVM globals are already destroyed. Do we know that none of what we call into access any globals?

Do you mean across the entire project, or specifically this event listener? Because while I don't think this GDB listener uses any _LLVM_ globals (as opposed to the external GDB global symbols) so should be safe, we do of course need to ensure that _nothing_ we do across the entire library makes use of LLVM globals after they are destroyed. I'm not sure we've ever properly vetted anything like that - we just take things on a case-by-case basis.

hvdijk:
> Do you mean across the entire project, or specifically this event listener?

I mean the LLVM code that we call into here specifically, yes.

frasercrmck:
> > Do you mean across the entire project, or specifically this event listener?
> 
> I mean the LLVM code that we call into here specifically, yes.

Ah yeah in that case I think this listener is fine in that regard.

PR #340, opened 2024-02-06
[UnitCL] Test *Half* tests under the 'COMPILER' filter

frasercrmck:
These fp16 tests are important enough that we should be running them in more configurations more often - with vectorization, with different compilation flags, etc. Relegating them to only the `check-UnitCL-half` target means we miss out on this coverage.

These will all be skipped if half/fp16 isn't enabled so shouldn't negatively impact testing time in the majority of configs.

### DIFF
--- a/source/cl/test/UnitCL/CMakeLists.txt
+++ b/source/cl/test/UnitCL/CMakeLists.txt
@@ -417,7 +417,7 @@ function(add_ca_unitcl_check name)
     # our testing by minimising duplication.
     # Allow users to tack on any extra tests to run. They may also disable any
     # tests that we run by default.
-    set(filter "--gtest_filter=*Compile*:*Link*:*Build*:*Execution*:*print*:*WorkGroupCollective*:*SubGroup*:${args_FILTER}")
+    set(filter "--gtest_filter=*Compile*:*Link*:*Build*:*Execution*:*Half*:*print*:*WorkGroupCollective*:*SubGroup*:${args_FILTER}")
   elseif(args_FILTER)
     set(filter "--gtest_filter=${args_FILTER}")
   endif()


coldav (approved):

hvdijk:
Agreed with the idea but the naming doesn't look right. The comment block above reads: "We run UnitCL in many compiler configurations, but many UnitCL tests are not especially affected by compiler flags. Thus for the compiler variations we set a filter to select the most compiler intensive tests." This wasn't intended to be a general "these tests matter" filter; if we want to turn it into that, can we rename that?

frasercrmck:
> Agreed with the idea but the naming doesn't look right. The comment block above reads: "We run UnitCL in many compiler configurations, but many UnitCL tests are not especially affected by compiler flags. Thus for the compiler variations we set a filter to select the most compiler intensive tests." This wasn't intended to be a general "these tests matter" filter; if we want to turn it into that, can we rename that?

Yeah I was wondering about that.

My personal interpretation of what we want to do in CI is have a subset of tests which are most affected by changes to the compiler or mux runtimes, as opposed to the CL API or whatever which change less often. This is primarily a compromise for MR testing times, which is fine.

Fundamentally though this is an inflexible system as ideally we'd _also_ test the whole of UnitCL in all the various configurations at another point in time (say nightly). We can't currently do that without adding extra checks with different combinations of filters and options/environment. Consequently we've got some gaps in our test coverage. Whether or not those are important gaps to fill is perhaps another question.

That said, I do think that it could be argued the fp16 tests I'm including *are* affected by compiler flags (i.e., vectorization, where we may turn scalar fp16 IR into vector fp16 IR and want to ensure the precision is still correct) an so even then it does fit with the current description. I'll admit though that I'm mostly ignoring the word "intensive" from the description as I don't think it necessarily agrees with the first sentence you quoted.

Either way I think renaming whatever the `COMPILER` arg means, and whether we could improve the system more generally, could be done separately and offline.

hvdijk (approved):
Hmm, yeah, I can agree it includes both tests that fit the criteria and tests that don't, and it's more important for tests that do fit the criteria to be included than for tests that don't to not be. So alright, that's probably good to merge and to further refine later.

PR #345, opened 2024-02-07
Fix in-tree builds' use of LLVM tools for compiling builtins

frasercrmck:
This looks like it was intended to work, by checking whether or not the `$tools_dir` was empty. Unfortunately, unless the tools directory is manually set to empty by a user, it will always be at least `/bin`. This meant that in practice is was trying to pick up tools from the path, leading to unintended behaviour.

This commit fixes this behaviour, by only setting
`CA_BUILTINS_TOOLS_DIR` if `CA_LLVM_INSTALL_DIR` is set.

If in-tree targets are used, we set the CMake targets using an alias executable. This gives us the full path to the in-tree target executable without worrying about accidentally using `clang` from the PATH, for example. This fixes another latent bug in the old code, which would import a `builtins::compiler` target called "clang" which may actually be on the PATH.

It also deletes some dead code, and consolidates the setting and use of `CA_BUILTINS_TOOLS_DIR` in one place. It removes references to `opt` which we no longer use when compiling builtins.


modules/compiler/builtins/CMakeLists.txt
```diff
@@ -117,7 +129,7 @@ add_custom_target(internal_builtins)
 #                 the dependencies.
 #
 # Variables expected to exist:
-# * BUILTINS_COMPILER: compiler binary with which to build the builtins.
+# * builtins::compiler: compiler binary with which to build the builtins.
```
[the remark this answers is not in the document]
frasercrmck: Good point! Will fix

modules/compiler/builtins/cmake/FindBuiltinsTools.cmake
```diff
@@ -22,93 +22,109 @@ set(BUILTINS_TOOLS_CMAKE INCLUDED)
...
-        "'<llvm-build>/bin/clang${CA_HOST_EXECUTABLE_SUFFIX}' to fix this "
-        "error.")
-    endif()
-    if(NOT EXISTS ${COMPILER_RESULT})
-      message(FATAL_ERROR
-        "${COMPILER_RESULT} does not exist. Clean the build directory and set "
-        "CA_BUILTINS_TOOLS_DIR to '<llvm-build>' or BUILTINS_COMPILER to "
-        "'<llvm-build>/bin/clang${EXTENSION}' to fix this error.")
-    endif()
+  string(REGEX MATCH ".*clang${CA_HOST_EXECUTABLE_SUFFIX}$"
```
hvdijk, line 85: I realize you did not change this beyond the indentation, but this seems wrong. `CA_HOST_EXECUTABLE_SUFFIX` would be expected to be something like `.exe`, and we don't want to interpret the `.` in there as a wildcard character. Could we go over the touched CMake bits to see if there's more that doesn't make sense and could be fixed while we're here anyway?

hvdijk (approved):
The changes look good, the existing CMake that this preserves less so. If you would prefer to leave that for a followup PR though, that would be fine with me, let's mark this as approved.

modules/compiler/builtins/cmake/FindBuiltinsTools.cmake
frasercrmck: Yeah, good point. But I don't really understand the point of his regex matching at all. I can't see a way that `BUILTINS_COMPILER` is anything but `${tools_dir}/clang${CA_HOST_EXECUTABLE_SUFFIX}` so isn't this regex trivially always matching? Or am I misunderstanding? Can the user set `BUILTINS_COMIPLER` themselves or something?

hvdijk: Hmm, I think you're right. I wonder if at one point this was planned to be a cache variable (which the user can adjust) but I cannot find it ever having been one. That allows for a nice cleanup.

frasercrmck: Yeah, that's a pleasant surprise. Thanks!

coldav (approved):

PR #346, opened 2024-02-07
[compiler] Add missing 'override' keyword. NFC

frasercrmck:


coldav (approved):

hvdijk (approved):
Commit message should say `override` rather than `virtual`, but looks good otherwise. (Ah, you already updated!)

PR #350, opened 2024-02-08
Do not import LLVM if LLVM targets already exist.

coldav:
# Overview

Do not import LLVM if the LLVMCore target already exists

# Reason for change

If we are being built within LLVM or we have basically imported LLVM already we do not wish to import LLVM again. 

# Description of change

Do not import LLVM if the LLVMCore target already exists
Furthermore we set a flag CA_IN_LLVM_TREE and do not add to the multi_llvm INTERFACE libraries.


hvdijk (approved):

RossBrunton (approved):

modules/compiler/multi_llvm/CMakeLists.txt
```diff
@@ -18,7 +18,9 @@ add_ca_interface_library(multi_llvm)
 
 target_include_directories(multi_llvm INTERFACE
 $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
-target_include_directories(multi_llvm SYSTEM INTERFACE
-  ${LLVM_INCLUDE_DIR})
+if (NOT CA_IN_LLVM_TREE)
```
frasercrmck, line 21: This doesn't feel too good to me. I know you've explained something to us offline but this could do with an explanation in code as there must surely be a better way of handling this. Even if it's a code comment saying it's unclear why it's necessary, or it's a hack, it still alerts future people to maybe question it later on, or not rely on it as unquestionably correct.

cmake/ImportLLVM.cmake
```diff
@@ -39,6 +41,14 @@ Once the modules are included we verify the LLVM version is supported and set
 some additional compile definitions.
 #]=======================================================================]
 
+# Set a flag for whether we are in an LLVM Tree and return early.
+# This is done by checking for the LLVMCore target which does not exist
+# until we import LLVM.
+set(CA_IN_LLVM_TREE OFF)
+if (TARGET LLVMCore)
+  set(CA_IN_LLVM_TREE ON)
```
frasercrmck, line 49: Am I missing something - should this not `return()` here? I thought if this falls through in an in-tree build, the next line would `FATAL_ERROR` because `CA_LLVM_INSTALL_DIR` isn't set.

cmake/ImportLLVM.cmake
```diff
@@ -16,7 +16,9 @@
 
 #[=======================================================================[.rst:
 Module imports LLVM into the ComputeAorta build by pulling in LLVM CMake
-modules from the ``install`` directory of a build.
+modules from the ``install`` directory of a build. This only happens of we have not
```
frasercrmck, line 19: I think there's also an argument for taking `CA_IN_LLVM_TREE` out of `ImportLLVM`, and only calling into `ImportLLVM` if `CA_IN_LLVM_TREE`/`OCK_IN_LLVM_TREE` is false. That way the logic of `ImportLLVM` is clearer - it imports LLVM. If we're already inside LLVM, we don't need to do that.

See how the documentation still says `User option CA_LLVM_INSTALL_DIR is required to be set to the filepath of the LLVM install before including this module`. This documentation is now incorrect and/or misleading.

I think if this module is to auto-detect being in-tree, it should probably be renamed (to `DetectLLVM` or something) and redocumented.

frasercrmck: In fact, none of the documentation after this paragraph is correct for `CA_IN_LLVM_TREE`.

frasercrmck: Also this new documentation overflows the 80 character column limit.

cmake/ImportLLVM.cmake
```diff
@@ -39,6 +41,14 @@ Once the modules are included we verify the LLVM version is supported and set
 some additional compile definitions.
 #]=======================================================================]
 
+# Set a flag for whether we are in an LLVM Tree and return early.
+# This is done by checking for the LLVMCore target which does not exist
+# until we import LLVM.
+set(CA_IN_LLVM_TREE OFF)
```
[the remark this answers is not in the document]
coldav: done

cmake/ImportLLVM.cmake
coldav: moved as suggested, so now no need for the return(). not sure what happened to the return() but you are correct.

frasercrmck (approved):
Thanks

modules/compiler/multi_llvm/CMakeLists.txt
coldav: I've added explanation. I agree that the original design of tagging it into multi_llvm seems wrong, but I think we need to discuss what would be better. This is a relatively small change to how it currently works and we can discuss what should be done going forward.

frasercrmck: Yep, my gut feeling is that `multi_llvm` looks like it's too much work and shouldn't be acting as the thing that pulls in LLVM headers

PR #351, opened 2024-02-09
[muxc] Link against compiler-base

frasercrmck:
This tool includes code from 'base' and clang headers and so should link against the compiler-base library.

This will transitively add the appropriate clang headers to the include path. In a pre-installed/external LLVM build they'll be included anyway because they live at the same location as the LLVM headers. However, in a (eventually correct) in-tree LLVM build, the clang headers will be in a different location and will only be included by the compiler-base library, and not the compiler-static library.


frasercrmck:
Wrong source repo, sorry

PR #352, opened 2024-02-09
[muxc] Link against compiler-base

frasercrmck:
This tool includes code from 'base' and clang headers and so should link against the compiler-base library.

This will transitively add the appropriate clang headers to the include path. In a pre-installed/external LLVM build they'll be included anyway because they live at the same location as the LLVM headers. However, in a (eventually correct) in-tree LLVM build, the clang headers will be in a different location and will only be included by the compiler-base library, and not the compiler-static library.

### DIFF
--- a/modules/compiler/tools/muxc/CMakeLists.txt
+++ b/modules/compiler/tools/muxc/CMakeLists.txt
@@ -28,11 +28,10 @@ target_compile_definitions(muxc PRIVATE
   muxc_CL_STD_30=$<BOOL:${muxc_CL_STD_30}>)
 target_include_directories(muxc SYSTEM PUBLIC ${LLVM_INCLUDE_DIR})
 target_include_directories(muxc PUBLIC
-  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/modules/compiler/source/base/include>
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/source/cl/source/compiler/include>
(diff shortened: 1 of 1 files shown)


modules/compiler/tools/muxc/CMakeLists.txt
```diff
@@ -28,11 +28,10 @@ target_compile_definitions(muxc PRIVATE
   muxc_CL_STD_30=$<BOOL:${muxc_CL_STD_30}>)
 target_include_directories(muxc SYSTEM PUBLIC ${LLVM_INCLUDE_DIR})
 target_include_directories(muxc PUBLIC
-  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/modules/compiler/source/base/include>
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/source/cl/source/compiler/include>
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/modules/compiler/multi_llvm/include>)
 target_link_libraries(muxc PRIVATE
-  compiler-static mux builtins cargo LLVMBitWriter LLVMIRReader)
+  compiler-static compiler-base mux builtins cargo LLVMBitWriter LLVMIRReader)
```
coldav, line 34: does it not get builtins and cargo as well from compiler-base?

coldav (approved):

modules/compiler/tools/muxc/CMakeLists.txt
frasercrmck: Yep, good point - and `mux`.

PR #354, opened 2024-02-12
Use include_directories for llvm includes under compiler

coldav:
# Overview

Use include_directories for llvm includes under compiler

# Reason for change

llvm suggests using include_directories for llvm projects and it is currently does not feel very clean.

# Description of change

At the moment we effectively add target include directories by having dependencies on multi_llvm which in itself adds the target include directories.

llvm suggests using include_directories for llvm projects and adding it at the compiler level feels a cleaner solution.

### DIFF
--- a/source/vk/CMakeLists.txt
+++ b/source/vk/CMakeLists.txt
@@ -90,6 +90,11 @@ target_include_directories(VK PUBLIC
   ${CMAKE_CURRENT_SOURCE_DIR}/include)
 target_include_directories(VK SYSTEM PUBLIC
   ${CMAKE_CURRENT_SOURCE_DIR}/external/Khronos/include)
+
+# TODO: has some llvm dependencies via spirv-ll/module.h. This should be looked at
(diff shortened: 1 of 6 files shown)


RossBrunton (approved):

source/vk/CMakeLists.txt
```diff
@@ -90,6 +90,11 @@ target_include_directories(VK PUBLIC
   ${CMAKE_CURRENT_SOURCE_DIR}/include)
 target_include_directories(VK SYSTEM PUBLIC
   ${CMAKE_CURRENT_SOURCE_DIR}/external/Khronos/include)
+
+# TODO: has some llvm dependencies via spirv/module.h. This should be looked at
```
frasercrmck, line 94: Just to check  - do you mean `spirv-ll/module.h` here?

coldav: updated comment

frasercrmck (approved):

PR #355, opened 2024-02-12
[compiler] Fix in-tree detection of clang includes/libs

frasercrmck:
1. When building against an in-tree LLVM, the clang headers are not in the same place as the LLVM headers, as in a pre-installed build.
2. When building against an in-tree LLVM, we don't have to provide the full paths to the (static) clang libraries we link against, because the libnames should already be library targets, which will be correctly resolved to the correct built in-tree libraries.

### DIFF
--- a/modules/compiler/source/base/CMakeLists.txt
+++ b/modules/compiler/source/base/CMakeLists.txt
@@ -70,6 +70,23 @@ target_include_directories(compiler-base PUBLIC
 target_include_directories(compiler-base SYSTEM PUBLIC
   ${spirv-headers_SOURCE_DIR}/include)
 
+if(OCK_IN_LLVM_TREE)
+  # When in-tree, we must explicitly specify the clang include directories, as
+  # they are not in the same place as the generic LLVM includes, as in a
+  # pre-installed LLVM.
+  if(NOT "clang" IN_LIST LLVM_ENABLE_PROJECTS
+      OR "${LLVM_EXTERNAL_CLANG_SOURCE_DIR}" STREQUAL "")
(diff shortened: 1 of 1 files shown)


modules/compiler/source/base/CMakeLists.txt
```diff
@@ -70,6 +70,23 @@ target_include_directories(compiler-base PUBLIC
 target_include_directories(compiler-base SYSTEM PUBLIC
   ${spirv-headers_SOURCE_DIR}/include)
 
+if(OCK_IN_LLVM_TREE)
+  # When in-tree, we must explicitly specify the clang include directories, as
+  # they are not in the same place as the generic LLVM includes, as in a
+  # pre-installed LLVM.
+  if(NOT "clang" IN_LIST LLVM_ENABLE_PROJECTS
+      OR "${LLVM_EXTERNAL_CLANG_SOURCE_DIR}" STREQUAL "")
+    message(FATAL_ERROR "The 'clang' project is not enabled")
```
coldav, line 79: could we have clang not enabled but only want compiler-pipeline for example?

frasercrmck: Er yeah probably. Is that something we want to support now? And if we did support that build configuration, would we want to be entering this subdirectory without clang present, or would we catch it earlier and avoid descending into `base` when `clang` isn't enabled? If so, this error could stay.

coldav: It's probably unlikely we would have that situation, true and you are right we could catch it earlier.

coldav (approved):

PR #358, opened 2024-02-13
Replacing the ock-demo workflow to use multiple jobs for its components instead of using the same workflow.

MaryaSharf:
# Overview

Replacing the workflow of creating oneapi demo to use multiple jobs for its components instead of using the same workflow.

### DIFF
--- a/.github/actions/build_portBLAS_action/action.yml
+++ b/.github/actions/build_portBLAS_action/action.yml
@@ -14,17 +14,39 @@ runs:
       shell: bash
       run: git clone --recursive https://github.com/codeplaysoftware/portBLAS
 
+       # installs tools, ninja, installs llvm and sets up sccahe
+    - name: setup ubuntu
+      uses:  ./.github/actions/setup_ubuntu_build
+
+    - name: Get Intel OneAPI BaseToolkit
+      shell: bash
+      run: |
+        wget "https://github.com/intel/llvm/releases/download/nightly-2024-02-08/sycl_linux.tar.gz"
+        mkdir linux_nightly_release
+        tar -xzf sycl_linux.tar.gz -C linux_nightly_release
+
     - name: Set up Environment and Build portBLAS
       shell: bash
       run: |
-        ls
-        export LD_LIBRARY_PATH=${{ inputs.workspace }}/linux_nightly_release/lib:$LD_LIBRARY_PATH
-        export CMAKE_CXX_COMPILER=${{ inputs.workspace }}/linux_nightly_release/bin/clang++
-        export CMAKE_C_COMPILER=${{ inputs.workspace }}/linux_nightly_release/bin/clang
+        export LD_LIBRARY_PATH=$(pwd)/linux_nightly_release/lib:$LD_LIBRARY_PATH
+        export CMAKE_CXX_COMPILER=$(pwd)/linux_nightly_release/bin/clang++
+        export CMAKE_C_COMPILER=$(pwd)/linux_nightly_release/bin/clang
         export OCL_ICD_FILENAMES=${{ inputs.workspace }}/install/lib/libCL.so
-        export CXX=${{ inputs.workspace }}/linux_nightly_release/bin/clang++
+        export CXX=$(pwd)/linux_nightly_release/bin/clang++
+
         # Note: With default options enabled, portBLAS supports complex math using
         # <ext/oneapi/experimental/sycl_complex.hpp>. It was removed in DPC++ in 0b5757bf.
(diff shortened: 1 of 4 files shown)


MaryaSharf:
I'll remove the trigger on PR once it runs completely and successfully. 
Next step is to update the getting_started guide.

coldav (approved):

.github/actions/build_portBLAS_action/action.yml
```diff
@@ -17,12 +17,13 @@ runs:
     - name: Set up Environment and Build portBLAS
       shell: bash
       run: |
-        ls
-        export LD_LIBRARY_PATH=${{ inputs.workspace }}/linux_nightly_release/lib:$LD_LIBRARY_PATH
-        export CMAKE_CXX_COMPILER=${{ inputs.workspace }}/linux_nightly_release/bin/clang++
-        export CMAKE_C_COMPILER=${{ inputs.workspace }}/linux_nightly_release/bin/clang
+        export LD_LIBRARY_PATH=${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/lib/libsycl.so:${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/lib:$LD_LIBRARY_PATH
```
frasercrmck, line 20: Does setting `LD_LIBRARY_PATH` to a specific `.so` actually do anything? I've have thought that pointing it to `/lib` as you're also doing would be enough.

.github/actions/build_portDNN_action/action.yml
```diff
@@ -17,11 +17,12 @@ runs:
     - name: Set up Environment and build portDNN
       shell: bash
       run: |
-        ls
-        export LD_LIBRARY_PATH=${{ inputs.workspace }}/linux_nightly_release/lib:$LD_LIBRARY_PATH
-        export CMAKE_CXX_COMPILER=${{ inputs.workspace }}/linux_nightly_release/bin/clang++
-        export CMAKE_C_COMPILER=${{ inputs.workspace }}/linux_nightly_release/bin/clang
+        export LD_LIBRARY_PATH=${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/lib/libsycl.so:${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/lib:$LD_LIBRARY_PATH
```
frasercrmck, line 20: Same question about the `.so`

.github/actions/build_portDNN_action/action.yml
```diff
@@ -17,11 +17,12 @@ runs:
...
+        export CMAKE_CXX_COMPILER=${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/bin/compiler/clang++
+        export CMAKE_C_COMPILER=${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/bin/compiler/clang
         export OCL_ICD_FILENAMES=${{ inputs.workspace }}/install/lib/libCL.so
+        export DPCPP_INCLUDE_DIR=${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/include/sycl
+        export PATH=${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/include/sycl:${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/bin:${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/lib:$PATH
 
-        cmake -B portDNN_build_dir portDNN -GNinja -DCMAKE_CXX_COMPILER=$(pwd)/linux_nightly_release/bin/clang++ -DSNN_BUILD_BENCHMARKS=OFF -DSNN_BENCH_SYCLBLAS=OFF -DSNN_BUILD_DOCUMENTATION=OFF
+        cmake -B portDNN_build_dir portDNN -GNinja -DCMAKE_CXX_COMPILER=${{ inputs.workspace }}/intel/oneapi/compiler/2024.0/bin/compiler/clang++ -DSNN_BUILD_BENCHMARKS=OFF -DSNN_BENCH_SYCLBLAS=OFF -DSNN_BUILD_DOCUMENTATION=OFF
```
frasercrmck, line 27: You're exporting `CMAKE_CXX_COMPILER` above to the same value as you're setting `-DCMAKE_CXX_COMPILER` here. Could you just have, say, `-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}`? That's not getting into the question of whether both are actually required to be set or not.

.github/actions/build_portBLAS_action/action.yml
coldav: I think you are right, it's just a path.

PR #365, opened 2024-02-15
Add risc-v to the host target supported targets.

coldav:
# Overview

This gives support for running clc for RISC-V as an additional compiler for the "host" target as well as building RISC-V OCK host target as a cross compile.

# Reason for change

RISC-V is not supported as a "host" target but is required for various projects.

# Description of change

This gives support for running clc for RISC-V as an additional compiler for the "host" target as well as building RISC-V OCK host target as a cross compile.
Note this also disables the JIT path as RISC-V does not support this.
We no longer go through the JIT constructs to create the TargetMachine if deferred compilation is disabled.


modules/compiler/targets/host/source/target.cpp
```diff
@@ -180,6 +231,20 @@ compiler::Result HostTarget::initWithBuiltins(
...
     CPUName = "x86-64-v3";  // Default only, may be overridden below.
+  } else if (triple.isRISCV()) {
+    if (triple.isArch32Bit()) {
+      Abi = "ilp32d";
+      CPUName = "generic-rv32";
+    } else {
+      Abi = "lp64d";
+      CPUName = "generic-rv64";
+    }
+    // We assume the following features must be supported.
```
hvdijk, line 242: Why these? What is this based on?

modules/mux/targets/host/source/device.cpp
```diff
@@ -552,6 +554,12 @@ host::arch device_info_s::detectHostArch() {
 #elif defined(UTILS_SYSTEM_64_BIT)
   arch = host::arch::X86_64;
 #endif
+#elif defined(UTILS_SYSTEM_RISCV)
+#if UTILS_SYSTEM_64_BIT
```
hvdijk, line 558: This could probably use the same style as the rest here (`#ifdef UTILS_SYSTEM_32_BIT`/.../`#elif defined(UTILS_SYSTEM_64_BIT)`/...)?

modules/mux/targets/host/test/UnitCL/host_clGetDeviceInfo.cpp
```diff
@@ -55,6 +55,10 @@ TEST_F(host_clGetDeviceInfoTest, Name) {
   const char *arch = "x86";
 #elif defined(__x86_64__) || defined(_M_X64)
   const char *arch = "x86_64";
+#elif defined(__riscv) && __riscv_xlen == 64
```
hvdijk, line 58: The check for `__riscv_xlen` should not be part of the same condition, it can cause warnings on non-RISC-V platforms due to `__riscv_xlen` being used without being defined.

modules/mux/targets/host/test/UnitCL/host_clGetDeviceInfo.cpp
```diff
@@ -55,6 +55,10 @@ TEST_F(host_clGetDeviceInfoTest, Name) {
...
+  const char *arch = "riscv64";
+#elif defined(__riscv__)
```
hvdijk, line 60: Why the difference between `__riscv` and `__riscv__`? Is that really how they're defined?

modules/compiler/targets/host/source/target.cpp
coldav: I can't recall the origins but it's what we use for refsi.

modules/mux/targets/host/test/UnitCL/host_clGetDeviceInfo.cpp
coldav: I can change it, but it's already like this in `modules/utils/include/utils/system.h`

frasercrmck: Yeah, it's probably best to check `__riscv`, then inside that, check `__riscv_xlen` is defined as 64 or 32 or `#error` otherwise.

modules/mux/targets/host/test/UnitCL/host_clGetDeviceInfo.cpp
frasercrmck: Yeah this doesn't look right. As I said above, it *should* be possible to check `__riscv_xlen` against 32 (carefully, mind)

modules/compiler/targets/host/source/target.cpp
frasercrmck: I think they're a fair default but could maybe use documentation to explain?

* We can't do anything without atomics (`a`).
* We *could* do stuff without multiplication/division (`m`) but we'd need extra compiler-rt support to do the most basic things and it'd be really slow.
* Assuming hardware support for fp32/fp64 (`f`/`d`) seems sensible to me. Same problem as above with compiler-rt, software emulation and performance.

All of these should probably be overrideable, as the TODO suggests, but I don't think it's required for now.

modules/mux/targets/host/test/UnitCL/host_clGetDeviceInfo.cpp
hvdijk: https://github.com/riscv-non-isa/riscv-toolchain-conventions claims "__riscv: defined for any RISC-V target. Older versions of the GCC toolchain defined __riscv__." but when I look through GCC's history, it never had `__riscv__`, it only had `__riscv` ever since it was first supported as a target (https://github.com/gcc-mirror/gcc/commit/09cae7507d9e88f2b05cf3a9404bf181e65ccbac#diff-6427e55c1bb73417ca77ff9c33ade2fe85a3796eb99c7de329c67717d45fb27d, look for `riscv_cpu_cpp_builtins`). This was GCC 7 and we already require newer GCC than GCC 7. I think we can simply get rid of `__riscv__` entirely.

modules/mux/targets/host/test/UnitCL/host_clGetDeviceInfo.cpp
```diff
@@ -55,6 +55,12 @@ TEST_F(host_clGetDeviceInfoTest, Name) {
   const char *arch = "x86";
 #elif defined(__x86_64__) || defined(_M_X64)
   const char *arch = "x86_64";
+#elif defined(__riscv) || defined(__riscv__)
+#if __riscv_xlen == 64
+  const char *arch = "riscv64";
+#elif __riscv_xlen == 32
+  const char *arch = "riscv32";
```
frasercrmck, line 62: Could probably do with an `#else #error` in case someone fudged a 128-bit riscv or something

modules/mux/targets/host/source/device.cpp
```diff
@@ -552,6 +554,12 @@ host::arch device_info_s::detectHostArch() {
 #elif defined(UTILS_SYSTEM_64_BIT)
   arch = host::arch::X86_64;
 #endif
+#elif defined(UTILS_SYSTEM_RISCV)
+#ifdef UTILS_SYSTEM_64_BIT
+  arch = host::arch::RISCV64;
+#elif defined(UTILS_SYSTEM_64_BIT)
+  arch = host::arch::RISCV32;
```
hvdijk, line 561: This isn't right, this checks the wrong macro. Minor nit, we seem to consistently have 32-bit logic first, then 64-bit logic, but that's reversed here. Not a big deal but if changes are needed anyway, might as well do this too.

frasercrmck (approved):

hvdijk (approved):

modules/mux/targets/host/test/UnitCL/host_clGetDeviceInfo.cpp
```diff
@@ -55,6 +55,14 @@ TEST_F(host_clGetDeviceInfoTest, Name) {
   const char *arch = "x86";
 #elif defined(__x86_64__) || defined(_M_X64)
   const char *arch = "x86_64";
+#elif defined(__riscv)
+#if __riscv_xlen == 64
+  const char *arch = "riscv64";
+#elif __riscv_xlen == 32
+  const char *arch = "riscv32";
+#else
+#error "32 bit or 64 bit xlen not define for RISC-V"
```
frasercrmck, line 64: nit: *defined* would be more correct

modules/compiler/targets/host/source/target.cpp
```diff
@@ -180,6 +231,24 @@ compiler::Result HostTarget::initWithBuiltins(
     }
   } else if (triple.isX86()) {
     CPUName = "x86-64-v3";  // Default only, may be overridden below.
+  } else if (triple.isRISCV()) {
+    // These are reasonable defaults, which has been used for various RISC-V
+    // target so far. We should allow overriding of this in the future
```
frasercrmck, line 236: nit: I think the CPU is overrideable, through `CA_HOST_TARGET_CPU`. Unless you mean the ABI?

frasercrmck (commented):
just a couple of final nits. LGTM otherwise

PR #372, opened 2024-02-22
[oclc] Fix input and output of 'half' values

frasercrmck:
Internally we were storing half data as cl_half, which is correct, but this is not an actual floating-point data type. Therefore when we converted those values to/from floating-point types, things would go wrong (e.g., we'd static cast 2.0 to the integer 2 instead of 0x4000).

The OpenCL headers provide some methods for correct translation between float and half, so we can use those instead.

### DIFF
--- a/source/cl/tools/oclc/oclc.cpp
+++ b/source/cl/tools/oclc/oclc.cpp
@@ -15,6 +15,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 // System headers
+#include <CL/cl_half.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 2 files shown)


source/cl/tools/oclc/oclc.cpp
```diff
@@ -1391,6 +1392,19 @@ T *oclc::Driver::CastToTypeFloat(const std::vector<std::string> &source,
   return casted_buffers.back().data();
 }
 
+cl_half *oclc::Driver::CastToTypeHalf(const std::vector<std::string> &source,
+                                      vector2d<cl_half> &casted_buffers,
+                                      size_t &size) {
+  std::vector<cl_half> data(source.size());
+  for (size_t i = 0; i < source.size(); ++i) {
+    auto f = atof(source[i].c_str());
+    data[i] = cl_half_from_float(f, cl_half_rounding_mode::CL_HALF_RTE);
```
hvdijk, line 1401: You can make `CastToTypeFloat<cl_half>` work with
```
auto f = atof(source[i].c_str());
if constexpr (std::is_same_v<T, cl_half>) {
  data[i] = cl_half_from_float(f, cl_half_rounding_mode::CL_HALF_RTE);
} else {
  data[i] = static_cast<T>(f);
}
```
so you don't need to duplicate the rest of the function.

coldav (approved):

source/cl/tools/oclc/oclc.cpp
frasercrmck: wonderful, thank you

hvdijk (approved):

PR #373, opened 2024-02-22
Support reqd_work_group_size with less than 3 operands

PietroGhg:
# Overview

Adds support for handling the `reqd_work_group_size` metadata node with less than three operands.

# Reason for change

This is needed for the SYCL Native CPU integration in DPC++. DPC++ may generate metadata nodes with less than 3 operands.

# Description of change

Defaults to 1 for the unspecified dimensions, similarly to what the `LLVM-SPIRV Translator` [does](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/747ef0f0ec03a25b5fb4edf5f5bf43e794503806/lib/SPIRV/PreprocessMetadata.cpp#L131).

Added a lit test for the `reqd_work_group_size` metadata with one and two operands.

### DIFF
--- a/modules/compiler/test/utils.cpp
+++ b/modules/compiler/test/utils.cpp
@@ -127,3 +127,34 @@ TEST_F(CompilerUtilsTest, ReplaceFunctionInMetadata) {
   ASSERT_EQ(Kernels.size(), 1);
   ASSERT_EQ(Kernels[0].Name, Bar->getName());
 }
+
+TEST_F(CompilerUtilsTest, ParseReqWGSize) {
+  auto M = parseModule(R"(
+  define void @foo() !reqd_work_group_size !1 {
+  entry:
+    ret void
+  }
+
+  define void @bar() !reqd_work_group_size !2 {
+  entry:
+    ret void
+  }
+
+  !1 = !{i32 30}
+  !2 = !{i32 30, i32 20}
+  )");
+
+  auto check = [&M](llvm::StringRef Name, const std::array<uint64_t, 3> &Res) {
+    auto *const F = M->getFunction(Name);
+    if (auto Parsed = parseRequiredWGSMetadata(*F)) {
+      for (auto [El, Exp] : llvm::zip(*Parsed, Res)) {
+        EXPECT_EQ(El, Exp);
+      }
+    } else {
+      GTEST_FAIL();
+    }
+  };
+
+  check("foo", {30, 1, 1});
+  check("bar", {30, 20, 1});
+}
--- a/modules/compiler/utils/source/metadata.cpp
+++ b/modules/compiler/utils/source/metadata.cpp
@@ -293,14 +293,12 @@ std::optional<unsigned> isSchedulingParameter(const Function &f, unsigned idx) {
 std::optional<std::array<uint64_t, 3>> parseRequiredWGSMetadata(
     const Function &f) {
   if (auto mdnode = f.getMetadata(ReqdWGSizeMD)) {
-    auto *const op0 = mdconst::extract<ConstantInt>(mdnode->getOperand(0));
-    auto *const op1 = mdconst::extract<ConstantInt>(mdnode->getOperand(1));
-    auto *const op2 = mdconst::extract<ConstantInt>(mdnode->getOperand(2));
-
-    // KLOCWORK "UNINIT.STACK.ARRAY.MUST" possible false positive
-    // This is normal std::array initialization
-    std::array<uint64_t, 3> wgs = {
-        {op0->getZExtValue(), op1->getZExtValue(), op2->getZExtValue()}};
+    std::array<uint64_t, 3> wgs = {0, 1, 1};
+    assert(mdnode->getNumOperands() >= 1 && mdnode->getNumOperands() <= 3 &&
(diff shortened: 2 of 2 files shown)


coldav (approved):

modules/compiler/utils/source/metadata.cpp
```diff
@@ -293,14 +293,23 @@ std::optional<unsigned> isSchedulingParameter(const Function &f, unsigned idx) {
...
+    auto op2 = numOps == 3
+                   ? mdconst::extract<ConstantInt>(mdnode->getOperand(2))
+                         ->getZExtValue()
+                   : 1;
 
     // KLOCWORK "UNINIT.STACK.ARRAY.MUST" possible false positive
     // This is normal std::array initialization
-    std::array<uint64_t, 3> wgs = {
-        {op0->getZExtValue(), op1->getZExtValue(), op2->getZExtValue()}};
+    std::array<uint64_t, 3> wgs = {{op0, op1, op2}};
```
hvdijk, line 312: If we flip this around, if we do `std::array<uint64_t, 3> wgs = {{0, 1, 1}};` first, we can then just do a loop over the `mdnode` operands.

modules/compiler/test/lit/passes/reqd_work_group_size_operands.ll
```diff
@@ -0,0 +1,54 @@
...
+;
+; Unless required by applicable law or agreed to in writing, software
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; RUN: muxc --passes work-item-loops,verify -S %s | FileCheck %s --check-prefix=CHECK1
```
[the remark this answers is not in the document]
PietroGhg: I'm not sure how reliable the order of the output function is, will it always stay the same when changing LLVM version etc?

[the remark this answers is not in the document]
frasercrmck: Ah I see what you're doing. I think it's fair to assume that the order is stable. We do it almost everywhere else in our LIT tests.

Though I was wondering if a `UnitCompiler` unit test might be preferable? Then we don't need to assume about the IR (neither the order of functions nor what comparisons we check against).

modules/compiler/utils/source/metadata.cpp
PietroGhg: Thanks for the suggestion, should be done :+1:

modules/compiler/utils/source/metadata.cpp
```diff
@@ -293,14 +293,13 @@ std::optional<unsigned> isSchedulingParameter(const Function &f, unsigned idx) {
...
-    auto *const op1 = mdconst::extract<ConstantInt>(mdnode->getOperand(1));
-    auto *const op2 = mdconst::extract<ConstantInt>(mdnode->getOperand(2));
-
-    // KLOCWORK "UNINIT.STACK.ARRAY.MUST" possible false positive
-    // This is normal std::array initialization
-    std::array<uint64_t, 3> wgs = {
-        {op0->getZExtValue(), op1->getZExtValue(), op2->getZExtValue()}};
+    std::array<uint64_t, 3> wgs = {0, 1, 1};
+    assert(mdnode->getNumOperands() >= 1 && mdnode->getNumOperands() <= 3 &&
+           "Unsupported number of operators in reqd_work_group_size");
```
hvdijk, line 298: Nit: operands, not operators.

modules/compiler/utils/source/metadata.cpp
```diff
@@ -293,14 +293,13 @@ std::optional<unsigned> isSchedulingParameter(const Function &f, unsigned idx) {
...
+    for (const auto &op : enumerate(mdnode->operands())) {
```
frasercrmck, line 299: nit: you could also do `[idx, op]` if you wanted?

modules/compiler/test/utils.cpp
```diff
@@ -127,3 +127,34 @@ TEST_F(CompilerUtilsTest, ReplaceFunctionInMetadata) {
...
+  define void @bar() !reqd_work_group_size !2 {
+  entry:
+    ret void
+  }
+
+  !1 = !{i32 30}
+  !2 = !{i32 30, i32 20}
+  )");
+
+  auto check = [&M](const char *Name, const std::array<uint64_t, 3> &Res) {
```
frasercrmck, line 147: nit: could be `llvm::StringRef` I spose

frasercrmck (approved):
Cheers!

modules/compiler/test/lit/passes/reqd_work_group_size_operands.ll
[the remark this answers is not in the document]
frasercrmck: I suggested a unit test because it's by far the best way of testing the API and the change to it. There's no guarantee that the work-item-loops pass actually parses that metadata, and if it does it might not use that API, and there's no guarantee it emits those comparisons, nor in that order. It's far flakier with a LIT test.

PR #374, opened 2024-02-22
Fix cmake variable CA_HOST_TARGET_CPU affecting all host targets

coldav:
# Overview

Fix cmake variable CA_HOST_TARGET_CPU affecting all host targets

# Reason for change

CA_HOST_TARGET_CPU affecting all host targets shows failures in UnitCompiler tests where it tries to apply the change to targets which are not suited for the value.

# Description of change

The cmake variable `CA_HOST_TARGET_CPU` has been split into capitalized architecture variants of the form `CA_HOST_TARGET_<ARCH>_CPU` e.g `CA_HOST_TARGET_X86_64_CPU`, `CA_HOST_TARGET_AARCH64_CPU` and `CA_HOST_TARGET_RISCV64_CPU`. The environment variable `CA_HOST_TARGET_CPU` remains the same name.


hvdijk:
Correct me if I'm wrong, but it looks like this drops `CA_HOST_TARGET_CPU_NATIVE` with no mention in the PR or commit message or documentation? It should be fine to drop as long as it's possible to set the CPU to `"native"` but it may be worth warning users about, because existing CMake configurations would continue to work but would start doing the wrong thing.

hvdijk (approved):

frasercrmck (approved):

PR #376, opened 2024-02-23
Update for LLVM 19.

hvdijk:
# Overview

Update for LLVM 19.

# Reason for change

# Description of change

* When using the new debug info format, make sure to actually use the new debug info format by converting the builtins module after loading.
* When checking debug info, make sure to check both the old and the new debug info formats.
* When inserting debug info, make sure to insert in the appropriate format.
* When inserting PHIs, ensure they are inserted before any debug info. We are also inconsistent in whether trying to insert PHIs before or after existing PHIs; this commit tries to preserve the existing insertion points as much as possible.
* In add.cl, do not check that no vsetvli is generated. Even if we do not generate it, LLVM 19 generates it itself. This is fine.

The changes in `scalarizer.cpp` are covered by the existing `llvm/insert_element_debug_info.ll` test.
The changes in `work_item_loops_pass.cpp` are covered the existing `passes/barriers-dbg.ll` test.

# Checklist

### DIFF
--- a/modules/compiler/utils/source/barrier_regions.cpp
+++ b/modules/compiler/utils/source/barrier_regions.cpp
@@ -982,6 +982,12 @@ void compiler::utils::Barrier::MakeLiveVariableMemType() {
         debug_intrinsics_.push_back(std::make_pair(dbgDeclare, offset));
       }
     }
+#if LLVM_VERSION_GREATER_EQUAL(19, 0)
+    const auto DPVDeclares = findDPVDeclares(member.value);
+    for (auto *const DPVDeclare : DPVDeclares) {
+      debug_dp_values_.push_back(std::make_pair(DPVDeclare, offset));
+    }
+#endif
     offset += member.size;
     live_variable_index_map_[std::make_pair(member.value, member.member_idx)] =
         field_tys.size();
--- a/modules/compiler/utils/source/work_item_loops_pass.cpp
+++ b/modules/compiler/utils/source/work_item_loops_pass.cpp
@@ -273,10 +273,8 @@ struct ScheduleGenerator {
       const compiler::utils::BarrierWithLiveVars &barrier, BasicBlock *block,
       StoreInst *SI) {
     DIBuilder DIB(module, /*AllowUnresolved*/ false);
-    for (auto debug_pair : barrier.getDebugIntrinsics()) {
-      auto *old_dbg_declare = debug_pair.first;
-      const unsigned live_var_offset = debug_pair.second;
-
+    auto RecreateDebugIntrinsic = [&](DILocalVariable *const old_var,
+                                      const unsigned live_var_offset) {
       const uint64_t dwPlusOp = dwarf::DW_OP_plus_uconst;
       // Use a DWARF expression to point to byte offset in struct where
       // the variable lives. This involves dereferencing the pointer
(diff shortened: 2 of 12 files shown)


modules/compiler/vecz/source/transform/control_flow_conversion_pass.cpp
```diff
@@ -2944,8 +2974,12 @@ bool ControlFlowConversionState::Impl::blendInstructions() {
     Type *T = opDef->getType();
     const unsigned numPreds = std::distance(pred_begin(B), pred_end(B));
     Value *blend = nullptr;
-    PHINode *PHI =
-        PHINode::Create(T, numPreds, opDef->getName() + ".merge", &B->front());
+    PHINode *PHI = PHINode::Create(T, numPreds, opDef->getName() + ".merge");
+#if LLVM_VERSION_GREATER_EQUAL(18, 0)
+    PHI->insertBefore(B->begin());
```
frasercrmck, line 2979: Do you think this could become a sensible `multi_llvm` helper? `insertAtBeginningOfBlock`? or something?

modules/compiler/utils/source/work_item_loops_pass.cpp
```diff
@@ -301,6 +301,36 @@ struct ScheduleGenerator {
       auto *const DVIntrinsic = cast<DbgVariableIntrinsic>(DII);
       ConvertDebugDeclareToDebugValue(DVIntrinsic, SI, DIB);
     }
+#if LLVM_VERSION_GREATER_EQUAL(19, 0)
+    for (auto debug_pair : barrier.getDebugDPValues()) {
```
frasercrmck, line 305: Can this share any code with the above?

frasercrmck (commented):
Could we test the work-item-loops and scalarizer changes in LIT?

modules/compiler/utils/source/work_item_loops_pass.cpp
hvdijk: Probably, yes. Will update, thanks.

modules/compiler/vecz/source/transform/control_flow_conversion_pass.cpp
frasercrmck: Ah, because of `getFirstNonPHI`. Uh, if you add `insertAtIterator` would we not still need to use `#if LLVM*` at every use to distinguish between `begin()`/`front()`?

modules/compiler/utils/source/barrier_regions.cpp
```diff
@@ -972,6 +972,9 @@ void compiler::utils::Barrier::MakeLiveVariableMemType() {
     // so record this and the matching byte offset into the struct.
 #if LLVM_VERSION_GREATER_EQUAL(18, 0)
     auto DbgIntrinsics = findDbgDeclares(member.value);
+#if LLVM_VERSION_GREATER_EQUAL(19, 0)
```
RossBrunton, line 975: Nitpick, but would it be possible to move the DPVDeclares outside of the `#if LLVM_VERSION_GREATER_EQUAL(18, 0)`? The nesting of the ifdefs makes it a bit hard to skim.

modules/compiler/utils/source/work_item_loops_pass.cpp
hvdijk: Done.

modules/compiler/utils/source/barrier_regions.cpp
hvdijk: Yeah, we don't need it here yet, I've moved it below.

frasercrmck (approved):

PR #386, opened 2024-02-29
[host] Add initial support for RISC-V relocations

frasercrmck:
This adds support for RISC-V relocations - enough to permit the running
of UnitCL (in the default configuration). There are obviously many other
relocations for RISC-V but without test coverage those have been left
for future work.

Note that relocation support is still incomplete. There is no support
yet for the PLT or GOT sections, so if the executable is loaded too far
away from external symbols, the relocation process will fail at program
load time, complaining of an unreachable symbol. This situation is
entirely down to the in-process memory of the OpenCL location and the
machine it's running on, so cannot be easily pre-determined.

Locally, at least, all (offline) UnitCL tests pass in isolation but fail
when run together, because the binary is loaded too far away from some
symbols such as memcpy.

The ADD32 and SUB32 relocations only seem to appear for the PIC
relocation model, which is not being used in main nor in this PR, but
they are simple enough that including them should be fairly obviously
correct.

### DIFF
--- a/modules/compiler/targets/host/source/target.cpp
+++ b/modules/compiler/targets/host/source/target.cpp
@@ -55,14 +55,14 @@
 namespace host {
 
 // Create a target machine, `abi` is optional and may be empty
-static llvm::TargetMachine *createTargetMachine(llvm::StringRef CPU,
-                                                llvm::StringRef Triple,
+static llvm::TargetMachine *createTargetMachine(llvm::Triple TT,
+                                                llvm::StringRef CPU,
                                                 llvm::StringRef Features,
                                                 llvm::StringRef ABI) {
   // Init the llvm target machine.
   llvm::TargetOptions Options;
   std::string Error;
-  const std::string TripleStr = Triple.str();
+  const std::string &TripleStr = TT.str();
   const llvm::Target *LLVMTarget =
       llvm::TargetRegistry::lookupTarget(TripleStr, Error);
   if (nullptr == LLVMTarget) {
@@ -73,6 +73,14 @@ static llvm::TargetMachine *createTargetMachine(llvm::StringRef CPU,
     Options.MCOptions.ABIName = ABI;
   }
 
+  std::optional<llvm::CodeModel::Model> CM;
+  // RISC-V needs the 'medium' code model as we don't ensure that the code and
+  // its statically defined symbols are loaded somewhere in the range of
+  // absolute addresses -2GB and +2GB, which is a requirement of the 'low' code
+  // model.
+  if (TT.isRISCV()) {
+    CM = llvm::CodeModel::Medium;
+  }
   // Aarch64 fails on UnitCL test
   // Execution/Execution.Barrier_02_Barrier_No_Duplicates/OfflineOpenCLC if we
   // don't set `JIT` to true. JIT for x86_64 and Aarch64 set the large code
@@ -83,8 +91,8 @@ static llvm::TargetMachine *createTargetMachine(llvm::StringRef CPU,
   // TODO: Investigate whether we can use a loader that does not have this
   // issue.
   return LLVMTarget->createTargetMachine(
-      Triple, CPU, Features, Options, /*RM=*/std::nullopt,
-      /*CM=*/std::nullopt, multi_llvm::CodeGenOptLevel::Aggressive,
+      TripleStr, CPU, Features, Options, /*RM=*/std::nullopt, CM,
+      multi_llvm::CodeGenOptLevel::Aggressive,
       /*JIT=*/true);
 }
 
@@ -372,8 +380,7 @@ compiler::Result HostTarget::initWithBuiltins(
         Features += '-' + FeatureName.str();
       }
     }
-    target_machine.reset(
-        createTargetMachine(CPUName, triple.str(), Features, ABI));
+    target_machine.reset(createTargetMachine(triple, CPUName, Features, ABI));
   }
 
   return compiler::Result::SUCCESS;
--- a/modules/loader/source/relocations.cpp
+++ b/modules/loader/source/relocations.cpp
@@ -15,11 +15,16 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 #include <cargo/endian.h>
+#include <cargo/error.h>
+#include <cargo/string_view.h>
+#include <inttypes.h>
+#include <loader/elf.h>
 #include <loader/relocation_types.h>
 #include <loader/relocations.h>
 
 #include <algorithm>
 #include <cstdint>
+#include <cstdlib>
 #include <optional>
 
 // There are many relocation types, but LLVM only emits a few, only those
@@ -80,10 +85,14 @@ inline Integer setBitRange(Integer value, Integer subvalue, int first,
 
 // Returns a triple of common variables used in relocation calculation:
 // * Writable position in for relocations (host memory)
-// * P - position of the relocation (target memory)
-// * S - value of the symbol in the symbol table (target memory)
+// * P:    : position of the relocation (in target memory)
+// * S + A : value of the symbol in the symbol table plus any relocation addend
+//           (in target memory)
+// See the link below for an explanation of the symbols - described in the
+// RISC-V ELF psABI document, but relevant for other architectures:
+// https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#calculation-symbols
 template <typename T>
-std::optional<std::tuple<uint8_t *, T, T>> decomposeRelocation(
+cargo::optional<std::tuple<uint8_t *, T, T>> decomposeRelocation(
     const loader::Relocation &r, loader::ElfMap &map) {
   static_assert(std::is_same_v<T, uint32_t> || std::is_same_v<T, uint64_t>,
                 "Invalid relocation size");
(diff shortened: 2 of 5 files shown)


modules/compiler/targets/host/source/target.cpp
```diff
@@ -73,6 +73,14 @@ static llvm::TargetMachine *createTargetMachine(llvm::StringRef CPU,
     Options.MCOptions.ABIName = ABI;
   }
 
+  std::optional<llvm::CodeModel::Model> CM;
+  // TODO: RISC-V (currently) needs the 'medium' code model as we don't ensure
```
coldav, line 77: what's the TODO saying - that we should switch to small and support it correctly or just investigate if this is correct?

modules/loader/source/relocations.cpp
```diff
@@ -645,88 +650,261 @@ bool resolveAArch64(const loader::Relocation &r, loader::ElfMap &map,
...
+  cargo::read_little_endian(&insn, insn_addr);
+  insn = setBitRange(insn, imm, 12, 20);
+  cargo::write_little_endian(insn, insn_addr);
+}
+
+uint32_t getLo12(uint64_t addr) {
+  return static_cast<uint32_t>(getBitRange(addr, 0, 12));
+}
+
+std::optional<uint32_t> getHi20(int64_t addr) {
```
coldav, line 674: What does this do, why do we sign extend an addr - is it a relative offset rather than an addr?

modules/loader/source/relocations.cpp
```diff
@@ -645,88 +650,261 @@ bool resolveAArch64(const loader::Relocation &r, loader::ElfMap &map,
...
+  auto signExtend32 = [](int64_t val) {
+    return static_cast<int64_t>(val << 32) >> 32;
+  };
+  addr += 0x800;
```
coldav, line 678: what's this bit for?

modules/loader/source/relocations.cpp
```diff
@@ -645,88 +650,261 @@ bool resolveAArch64(const loader::Relocation &r, loader::ElfMap &map,
...
+bool resolveRISCV(const loader::Relocation &r, loader::ElfMap &map,
+                  const std::vector<loader::Relocation> &relocations) {
+  const auto relocation_data = decomposeRelocation<uint64_t>(r, map);
+  if (!relocation_data) {
+    return false;
+  }
+  auto [relocation_address, relocation_target_address, symbol_target_address] =
+      *relocation_data;
+
+  // A PC-relative offset (S + A - P)
```
coldav, line 698: I think it would be clearer to say what S,A and P are first.

modules/compiler/targets/host/source/target.cpp
frasercrmck: Maybe we can remove the `TODO` as it's not clear (to me, at least) whether this is a bug that needs to be fixed or an acceptable convention we are okay with in the RISC-V host target.

modules/loader/source/relocations.cpp
frasercrmck: No, this is generic enough for either absolute (`HI20`) or relative (`PCREL_HI20`), so `addr` is an address that we're relocating. However, a `HI20` or `PCREL_HI20` relocation is always a 32-bit address, so we sign extend the 64-bit address we have allocated to assert loudly we're not incorrectly resolving the relocation to a symbol that's too big (absolute) or too far away (relative) for the instruction using it.

modules/loader/source/relocations.cpp
frasercrmck: That's just how HI20 relocations are resolved. Think of it because you're taking it and adding a sign-extended LO12 value to it later.

If you had to resolve `0xDEADBEEF` and took the pieces apart naively:

HI20 = `0xDEADB` -> `0xDEADB000`
LO12 = `0xEEF` -> `0xFFFFFEEF`

If you added them up you'd get `0xDEADB000` + `0xFFFFFEEF` = `0xDEADAEEF` which isn't correct.

So instead you take:

HI20 = `0xDEADB` + `0x1` (skipping the addition of `0x800` then shifting right by 12) -> `0xDEADC000`
LO12 = `0xEEF` -> `0xFFFFFFEEF`

Adding those together gets the correct 32-bit address: `0xDEADC000` + `0xFFFFEEF` = `0xDEADBEEF`.

I didn't document it explicitly as it's just taken as a given of how you do it in the ELF documentation

https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#absolute-addresses
https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#pc-relative-symbol-addresses

modules/loader/source/relocations.cpp
frasercrmck: I document them in `decomposeRelocation` as those are the return values of that function.

modules/compiler/targets/host/source/target.cpp
coldav: that makes sense

modules/loader/source/relocations.cpp
coldav: okay that makes sense, maybe just link to the documentation.

modules/loader/source/relocations.cpp
coldav: if it's absolute can't the top bit be set as an unsigned value?

modules/loader/source/relocations.cpp
```diff
@@ -80,8 +85,8 @@ inline Integer setBitRange(Integer value, Integer subvalue, int first,
 
 // Returns a triple of common variables used in relocation calculation:
 // * Writable position in for relocations (host memory)
-// * P - position of the relocation (target memory)
-// * S - value of the symbol in the symbol table (target memory)
+// * P:    : position of the relocation (target memory)
+// * S + A : value of the symbol in the symbol table (target memory)
```
coldav, line 89: why is there two values S and A, what's the difference?

modules/loader/source/relocations.cpp
coldav: fair enough I just find it hard to relate to these other named values in this actual function, without keeping looking back.

coldav: if I understand this correctly it's (relocation_target_address(S) + symbol_target_address(A) - relocation_address(P)

coldav, line 698: I'm not sure what the comment refers to , is it just to explain L699 or all the relative offsets later on?

modules/loader/source/relocations.cpp
```diff
@@ -645,88 +650,261 @@ bool resolveAArch64(const loader::Relocation &r, loader::ElfMap &map,
...
+      (void)fprintf(stderr, "Missing symbol: %.*s\n", (int)symbol_name.size(),
+                    symbol_name.data());
+      (void)fprintf(stderr, "   with relocation type: %u\n", r.type);
+#endif
+      CARGO_ASSERT(0, "Unsupported relocation type.");
+      return false;
+    }
+    case R_RISCV_64:
+      // 64-bit relocation: S + A
+      cargo::write_little_endian(symbol_target_address & 0xFFFFFFFF,
```
coldav, line 733: Is symbol_target_address  not a 64 bit value - does it not need to be cast to 32 bit to pick the right little_endian template?

modules/loader/source/relocations.cpp
frasercrmck: Both, I guess. It's to explain the variable `relative_value` and everywhere it's used

frasercrmck: They're just the calculation symbols used in the official RISC-V ELF documentation: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#calculation-symbols. I thought it would be clearer for people to reference those against the spec this code is implementing.

modules/loader/source/relocations.cpp
frasercrmck: Yes you're right, surprised that worked locally.

modules/loader/source/relocations.cpp
frasercrmck: Hmm yeah that sounds like I've got a bug after I started refactoring things together, thanks.

modules/loader/source/relocations.cpp
frasercrmck: done

modules/loader/source/relocations.cpp
frasercrmck: should be fixed now

modules/loader/source/relocations.cpp
coldav: I get you are using known elf symbols, but I don't think it does any harm to say what these values are

coldav (approved):

modules/loader/source/relocations.cpp
coldav: I think it works locally because of little endian, and you right the low half first, it's just that scribbles past on the second write.

modules/loader/source/relocations.cpp
frasercrmck: not sure what you are referring to exactly, but if you mean a PC-relative value in relocations such as `PCREL_LO12_I` which are just `S - P`, you're right, but that's because the addend `A` must be zero for those relocations. Other PC-relative values like `PCREL_HI20` must include the addend `A`.

Since `symbol_target_address` is `S + A` I think this documentation is consistent.

modules/loader/source/relocations.cpp
frasercrmck: I've expanded the comment to say it returns `S + A`, "the value of the symbol in the symbol table plus any relocation addend" which I think covers it

modules/loader/source/relocations.cpp
frasercrmck: On a second look I don't think this is a bug because all of the `HI20` relocations use the `LUI` or `AUIPC` instructions which sign-extend the immediate up to 32 or 64 bits depending on the RISC-V architecture. So a 32-bit absolute address with the top bit set on `riscv64` would necessarily be sign-extended to the incorrect address. I think we are correct to throw an error on that (it appears that `lld` does the same thing, fwiw)

PR #387, opened 2024-03-04
[cl] Permit size 0 in clEnqueue(Read|Write|Copy|Fill)Buffer

frasercrmck:
This is not explicitly forbidden by the current unified OpenCL specification for `clEnqueueReadBuffer`, `clEnqueueWriteBuffer`, or `clEnqueueCopyBuffer`; though it *was* forbidden up until OpenCL 2.0.

Therefore, make the checks for size 0 dependent on the OpenCL version being built.

A size of zero for `clEnqueueFillBuffer` has always been valid, so explicitly test that. The `muxCommandFillBuffer` function forbids a size of 0, so it was a bug to call that function when size was 0.


source/cl/test/UnitCL/source/clEnqueueReadBuffer.cpp
```diff
@@ -210,10 +210,16 @@ TEST_F(clEnqueueReadBufferTest, InvalidBuffer) {
                           0, nullptr, nullptr));
 }
 
-TEST_F(clEnqueueReadBufferTest, InvalidSize) {
-  ASSERT_EQ_ERRCODE(CL_INVALID_VALUE,
-                    clEnqueueReadBuffer(command_queue, inMem, CL_TRUE, 0, 0,
-                                        inBuffer, 0, nullptr, nullptr));
+TEST_F(clEnqueueReadBufferTest, SizeZero) {
+  // An error when size == 0 was removed starting with OpenCL 2.1.
+  if (UCL::isDeviceVersionAtLeast({2, 1})) {
```
hvdijk, line 215: Do we test OpenCL <3.0 anywhere?

examples/refsi/hal_refsi/source/refsi_hal_g1.cpp
```diff
@@ -273,12 +273,14 @@ bool refsi_g1_hal_device::mem_copy(hal::hal_addr_t dst, hal::hal_addr_t src,
             dst, src, size);
   }
 
-  std::vector<uint8_t> temp(size);
-  if (!mem_read(temp.data(), src, size, locker)) {
-    return false;
-  }
-  if (!mem_write(dst, temp.data(), size, locker)) {
-    return false;
+  if (size) {
```
hvdijk, line 276: I'm assuming this is because `mem_read` and `mem_write` still require `size` to be greater than zero? Are they used anywhere directly where a size of zero should be allowed?

coldav:
Hi @frasercrmck, do you think we should just close this one now?

frasercrmck:
> Hi @frasercrmck, do you think we should just close this one now?

Blast from the past. I've personally no drive to get this over the line - I can't even remember why it was required/desired? I'm happy if you want to close it. Presumably it can't be that important if you hadn't needed it in over a year.

PR #388, opened 2024-03-04
Round up memory_size.

hvdijk:
# Overview

Round up memory_size.

# Reason for change

We calculate memory_size as os_memory_bounded_size() / 4, but on 32-bit platforms with more than 4GiB of memory, os_memory_bounded_size() is 0xffffffff, so we end up calculating memory_size as 0x3fffffff. Given that we are calculating memory_size this way specifically for OpenCL CTS which assumes that it sees a multiple of various pattern sizes it uses, make it more suitable for OpenCL CTS by rounding up.

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/modules/mux/targets/host/source/device.cpp
+++ b/modules/mux/targets/host/source/device.cpp
@@ -432,10 +432,10 @@ device_info_s::device_info_s(host::arch arch, host::os os, bool native,
   this->clock_frequency = native ? os_cpu_frequency() : 0;
   this->compute_units = native ? os_num_cpus() : 0;
   this->buffer_alignment = sizeof(uint64_t) * 16;
-  // TODO redmine(8386): Reported memory size is quartered in order to pass the
+  // TODO Reported memory size is quartered (rounded up) in order to pass the
   // OpenCL CTS however this probably should be an OCL specific fix and not in
   // the host target.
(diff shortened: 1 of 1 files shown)


modules/mux/targets/host/source/device.cpp
```diff
@@ -432,10 +432,10 @@ device_info_s::device_info_s(host::arch arch, host::os os, bool native,
   this->clock_frequency = native ? os_cpu_frequency() : 0;
   this->compute_units = native ? os_num_cpus() : 0;
   this->buffer_alignment = sizeof(uint64_t) * 16;
-  // TODO redmine(8386): Reported memory size is quartered in order to pass the
-  // OpenCL CTS however this probably should be an OCL specific fix and not in
-  // the host target.
-  this->memory_size = native ? os_memory_bounded_size() / 4 : 0;
+  // TODO redmine(8386): Reported memory size is quartered (rounded up) in order
```
coldav, line 435: might be worth checking the redmine to see if this still has relevance.

coldav (approved):

modules/mux/targets/host/source/device.cpp
hvdijk: Checked, it still has relevance, but it's in a place where we would lose track of it so I created a new internal ticket so that we do keep track of it.

PR #394, opened 2024-03-06
Fix Riscv test run command to use the correct compiler

MaryaSharf:


hvdijk:
Is the change to run this on PRs involving `.github/workflows/run_ock_demo.yml` intended to stay in, or is that just so that you can test it in this PR? It's fine to me either way, but if it stays in, can you also include it in the commit message?

hvdijk (approved):

MaryaSharf:
No it takes quite long to run thats why dont want to keep it in, added just for testing purposes.

PR #396, opened 2024-03-06
[loader] Fix ELF64_shdr field size

frasercrmck:
The "sh_flags" field is meant to be Elf64_Xword, i.e., 64 bits in size.

Also sprinkle in a few static assertions to ensure the data structures
are the correct size.

### DIFF
--- a/modules/loader/include/loader/elf.h
+++ b/modules/loader/include/loader/elf.h
@@ -209,6 +209,8 @@ class ElfFile {
   };
   using Header32 = Header<uint32_t>;
   using Header64 = Header<uint64_t>;
+  static_assert(sizeof(Header32) == 0x34, "Wrong Header size");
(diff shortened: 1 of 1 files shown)


coldav (approved):

modules/loader/include/loader/elf.h
```diff
@@ -266,6 +271,7 @@ class ElfFile {
     addr_t value;
     addr_t size;
   };
+  static_assert(sizeof(Symbol64) == 0x18, "Wrong Symbol size");
```
hvdijk, line 274: Not necessary to make it part of this PR, but would it make sense to have a `Symbol<addr_t>` template class?

hvdijk: Oh wait, hadn't clocked that the fields are in a different order, it wouldn't work, never mind.

PR #398, opened 2024-03-08
Handle barriers with no successor

PietroGhg:
Currently code similar to 
```
define internal spir_kernel void @__vecz_v8__ZTSZZN8orderingclEvENKUlRN4sycl3_V17handlerEE_clES3_EUlT_E_(ptr addrspace(3) noundef align 1 %_arg_arr_acc, ptr addrspace(3) noundef align 4 %_arg_local_acc,
ptr addrspace(1) noundef align 1 %_arg_res_acc, ptr noundef align 8 %_arg_res_acc9) local_unnamed_addr #7 !srcloc !10 !kernel_arg_buffer_location !11 !kernel_arg_runtime_aligned !12 !kernel_arg_exclusive
_ptr !12 !sycl_fixed_targets !13 !sycl_kernel_omit_args !14 !codeplay_ca_vecz.derived !21 {
entry:
  %mux_call7 = call i64 @__mux_get_num_groups(i32 0) #12
  %mux_call9 = call i64 @__mux_get_local_size(i32 0) #12
  %mux_call11 = call i64 @__mux_get_local_size(i32 0) #12
  %mux_call8 = call i64 @__mux_get_num_groups(i32 0) #12
  %mux_call13 = call i64 @__mux_get_group_id(i32 0) #12
  %mux_call12 = call i64 @__mux_get_group_id(i32 0) #12
  %mux_call10 = call i64 @__mux_get_local_size(i32 0) #12
  %mux_call5 = call i64 @__mux_get_local_id(i32 0) #12
  %mux_call = call i64 @__mux_get_global_offset(i32 0) #12
  %mux_call6 = call i64 @__mux_get_local_id(i32 0) #12
  %.splatinsert1 = insertelement <8 x i64> poison, i64 %mux_call6, i64 0
  %.splat2 = shufflevector <8 x i64> %.splatinsert1, <8 x i64> poison, <8 x i32> zeroinitializer
  %0 = add <8 x i64> %.splat2, <i64 0, i64 1, i64 2, i64 3, i64 4, i64 5, i64 6, i64 7>
  %mux_call4 = call i64 @__mux_get_global_offset(i32 0) #12
  %arrayidx.i = getelementptr inbounds i8, ptr addrspace(3) %_arg_arr_acc, i64 %mux_call6
  %arrayidx.ascast.i = addrspacecast ptr addrspace(3) %arrayidx.i to ptr
  store <8 x i8> zeroinitializer, ptr %arrayidx.ascast.i, align 1
  br label %barrier

barrier:                                          ; preds = %entry
  call void @__mux_work_group_barrier(i32 1, i32 2, i32 912) #12
  store <8 x i8> <i8 1, i8 1, i8 1, i8 1, i8 1, i8 1, i8 1, i8 1>, ptr %arrayidx.ascast.i, align 1
  %cmp.i3 = icmp ult <8 x i64> %0, <i64 2147483648, i64 2147483648, i64 2147483648, i64 2147483648, i64 2147483648, i64 2147483648, i64 2147483648, i64 2147483648>
  %1 = extractelement <8 x i1> %cmp.i3, i64 0
  %2 = extractelement <8 x i1> %cmp.i3, i64 1
  %3 = extractelement <8 x i1> %cmp.i3, i64 2
  %4 = extractelement <8 x i1> %cmp.i3, i64 3
  %5 = extractelement <8 x i1> %cmp.i3, i64 4
  %6 = extractelement <8 x i1> %cmp.i3, i64 5
  %7 = extractelement <8 x i1> %cmp.i3, i64 6
  %8 = extractelement <8 x i1> %cmp.i3, i64 7
  call void @llvm.assume(i1 %1)
  call void @llvm.assume(i1 %2)
  call void @llvm.assume(i1 %3)
  call void @llvm.assume(i1 %4)
  call void @llvm.assume(i1 %5)
  call void @llvm.assume(i1 %6)
  call void @llvm.assume(i1 %7)
  call void @llvm.assume(i1 %8)
  %conv.i = trunc i64 %mux_call6 to i32
  store atomic volatile i32 %conv.i, ptr addrspace(3) %_arg_local_acc unordered, align 4
  tail call void @llvm.trap()
  unreachable
}
```
Leads to a crash in `makeWrapperFunction`. This PR tries to address this by adding a call to `llvm.trap` in the basic block that terminates the wrapper function for the barrier region.

### DIFF
--- a/modules/compiler/utils/source/work_item_loops_pass.cpp
+++ b/modules/compiler/utils/source/work_item_loops_pass.cpp
@@ -1664,6 +1664,7 @@ Function *compiler::utils::WorkItemLoopsPass::makeWrapperFunction(
 
       const auto &successors = barrierMain.getSuccessorIds(i);
       const auto num_succ = successors.size();
+
       if (num_succ == 1) {
         // If there is only one successor, we can branch directly to it
         exitIR.CreateBr(bbs[successors.front()]);
@@ -1680,6 +1681,22 @@ Function *compiler::utils::WorkItemLoopsPass::makeWrapperFunction(
                            br_block);
 
         exitIR.CreateBr(br_block);
+      } else if (num_succ == 0) {
+        // If a barrier region has no successor, we just emit a call to
+        // llvm.trap and unreachable. A barrier region can have zero successors
+        // if all its terminators end in unreachable. Since there are no
+        // successors, it is not possible to continue and therefore we emit an
+        // unreachable here.
+
+        // TODO: we should be flagging up unreachables sooner, so that we avoid
+        // wrapping barrier regions with no successors with work item loops,
+        // and we should also make sure that the barrier region has no
+        // successors because of all its terminators ending in unreachable.
+        // If it's not the case we may want to handle that differently.
+        auto trap =
+            M.getOrInsertFunction("llvm.trap", Type::getVoidTy(context));
+        exitIR.CreateCall(trap);
+        exitIR.CreateUnreachable();
       } else {
         // Make a basic block with a switch to jump to the next subkernel
         auto *const switch_body =
--- a/modules/compiler/test/lit/passes/work-item-loop-no-successor.ll
+++ b/modules/compiler/test/lit/passes/work-item-loop-no-successor.ll
@@ -1,0 +1,44 @@
+; Copyright (C) Codeplay Software Limited
+;
+; Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+; Exceptions; you may not use this file except in compliance with the License.
+; You may obtain a copy of the License at
+;
+;     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+;
+; Unless required by applicable law or agreed to in writing, software
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; RUN: muxc --passes work-item-loops,verify < %s | FileCheck %s
+target triple = "spir64-unknown-unknown"
+target datalayout = "e-i64:64-v16:16-v24:32-v32:32-v48:64-v96:128-v192:256-v256:256-v512:512-v1024:1024"
+
+define internal spir_kernel void @foo(ptr %an_arg) #0 {
+entry:
+  %mux_call = call i64 @__mux_get_local_id(i32 0) #12
+  br label %barrier
+
+barrier:                                          ; preds = %entry
+  call void @__mux_work_group_barrier(i32 1, i32 2, i32 912) #12
+  tail call void @llvm.trap()
+  unreachable
+}
+
+declare i64 @__mux_get_local_id(i32)
+
+declare void @__mux_work_group_barrier(i32, i32, i32)
+
+declare void @llvm.trap()
+
+declare void @llvm.assume(i1)
+
+attributes #0 = { "mux-kernel"="entry-point" }
+
+; CHECK: exitIR12:                                         ; preds = %exitIR11
+; CHECK:   call void @llvm.trap()
+; CHECK:   unreachable


modules/compiler/utils/source/work_item_loops_pass.cpp
```diff
@@ -1611,7 +1645,14 @@ Function *compiler::utils::WorkItemLoopsPass::makeWrapperFunction(
...
       }
 
+      const auto &successors = barrierMain.getSuccessorIds(i);
+      const auto num_succ = successors.size();
+      const bool barrierInTrapBB = isBarrierInTrapBB(barrierMain, i);
+
       auto *const exitBlock = [&]() {
+        if (num_succ == 0 && barrierInTrapBB) {
+          return schedule.makeRunOneWorkItem(block, i, true);
+        }
```
coldav, line 1655: maybe an assert if we get into this position with no barrierInTrapBB since this means something has gone wrong.

modules/compiler/utils/source/work_item_loops_pass.cpp
```diff
@@ -1343,6 +1356,27 @@ void setUpLiveVarsAlloca(compiler::utils::BarrierWithLiveVars &barrier,
...
+                       unsigned id) {
+  if (auto *const CI = barrierMain.getBarrierCall(id)) {
+    BasicBlock *originalBB = CI->getParent();
+    Instruction *terminator = originalBB->getTerminator();
+    if (isa<UnreachableInst>(terminator) &&
+        originalBB->sizeWithoutDebug() >= 2) {
+      Instruction &pre_term = *(++originalBB->rbegin());
+      if (auto *ci = dyn_cast<CallInst>(&pre_term)) {
+        auto called = ci->getCalledFunction();
+        if (called->getName() == "llvm.trap") {
```
coldav, line 1371: Does this work if there are multiple blocks or is there guaranteed to be only one here?

I'm wondering whether we need to check for the trap since if we hit an unreachable as our only successor we are unable to continue anyway.

modules/compiler/utils/source/work_item_loops_pass.cpp
PietroGhg: Yeah we should hit the `llvm_unreachable` [here](https://github.com/PietroGhg/oneapi-construction-kit/blob/111af696544a721ca1cb07e42586f6b5012fca30/modules/compiler/utils/source/work_item_loops_pass.cpp#L1698) if `num_succ == 0` without `barrierInTrapBB`, but I can add an assert here too

modules/compiler/utils/source/work_item_loops_pass.cpp
PietroGhg: If there are multiple blocks then the barrier region will have more than zero successors I think, so it should go in the code path that handles that case, I'll add more test cases

coldav: is that true? 
If you have block 1 branch unconditional to  block 2(terminate), isn't that just single block terminate?

Or if we have more than one unreachable terminates they wouldn't be in the first block.

I don't know this code that well, so I could just be misunderstanding - it should basically follow the same logic as num_sucessors when it gets built

coldav (approved):

modules/compiler/utils/source/work_item_loops_pass.cpp
```diff
@@ -1680,6 +1681,15 @@ Function *compiler::utils::WorkItemLoopsPass::makeWrapperFunction(
                            br_block);
 
         exitIR.CreateBr(br_block);
+      } else if (num_succ == 0) {
+        // If a barrier region has no successor, we just emit a call to
+        // llvm.trap
```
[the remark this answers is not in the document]
PietroGhg: I've updated the comment, thank you

modules/compiler/utils/source/work_item_loops_pass.cpp
```diff
@@ -1680,6 +1681,21 @@ Function *compiler::utils::WorkItemLoopsPass::makeWrapperFunction(
                            br_block);
 
         exitIR.CreateBr(br_block);
+      } else if (num_succ == 0) {
+        // If a barrier region has no successor, we just emit a call to
+        // llvm.trap. A barrier region can have zero successors if all its
```
coldav, line 1686:
```suggestion
        // llvm.trap and unreachable. A barrier region can have zero successors if all its
```

coldav (approved):

PR #399, opened 2024-03-13
Added support for WG mode cpu HAL and fixed some OCK only issues

coldav:
# Overview

Got cpu hal working with OCK in work group mode

# Reason for change

We wish to use this as the basis for a remote OCK. It also works as a simple demonstration.

# Description of change

Enabled c++17 for access to std::aligned for clik, so we could align the argument pack.
Added HAL_CPU_WG_MODE for cpu hal and support two paths in cpu hal.

Switched to common arg packing for cpu hal. Updated common arg packing to support aligned allocation and work item mode.

Switched library output to be in the directory expected for hal loading.

Added CA_HAL_DEBUG env variable to assist debugging

Fixed fill bug.


clik/external/hal_cpu/source/CMakeLists.txt
```diff
@@ -21,6 +21,12 @@ add_library(hal_cpu SHARED
 
 target_compile_definitions(hal_cpu PUBLIC -DBUILD_HAL_DLL)
 
+if(HAL_CPU_WG_MODE)
+  target_compile_definitions(hal_cpu PUBLIC -DHAL_CPU_WG_MODE=1)
+else()
+  target_compile_definitions(hal_cpu PUBLIC -DHAL_CPU_WI_MODE=1)
+endif()
```
hvdijk, line 28: Can we change the macro setup to make it more obvious that these are mutually exclusive? Something like `-DHAL_CPU_MODE=HAL_CPU_MODE_WG` vs. `-DHAL_CPU_MODE=HAL_CPU_MODE_WI`?

clik/external/hal_cpu/source/cpu_hal.cpp
```diff
@@ -32,12 +35,36 @@
 
 #include "device/device_if.h"
 
+// http://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
+static uint64_t round_up_pot(uint64_t v) {
```
hvdijk, line 39: Given that the link explains that this does not always round up to the next power of two, can you document what this does do, what assumptions it makes? Or is there perhaps an LLVM function whose name we could align with, in which case we could just document it as behaving like the corresponding LLVM function?

clik/external/hal_cpu/source/cpu_hal.cpp
```diff
@@ -32,12 +35,36 @@
...
+      debug = true;
+    }
+  }
 }
 
-cpu_hal::~cpu_hal() { free(local_mem); }
+cpu_hal::~cpu_hal() {
+#if defined(HAL_CPU_WI_MODE)
+  std::free(local_mem);
+#endif
```
hvdijk, line 66: In non-WI mode, it would be nice if the destructor could be defaulted. Would that also make it trivial?

clik/external/hal_cpu/source/cpu_hal.cpp
```diff
@@ -147,10 +174,40 @@ void cpu_barrier::wait(int num_threads) {
...
+  uint32_t group_x_next = std::min(group_x_start + num_groups_per_thread,
+                                   (uint32_t)(exec->wg.num_groups[0]));
+
+  for (uint32_t wg_z = 0; wg_z < exec->wg.num_groups[2]; wg_z++) {
+    exec->wg.group_id[2] = wg_z;
+    for (uint32_t wg_y = 0; wg_y < exec->wg.num_groups[1]; wg_y++) {
+      exec->wg.group_id[1] = wg_y;
+      for (uint32_t wg_x = group_x_start; wg_x < group_x_next; wg_x++) {
+        exec->wg.group_id[0] = wg_x;
+        kernel((void *)exec->packed_args, exec);
```
hvdijk, line 205: I know this was already there for the WI mode, but what is the cast doing here?

clik/external/hal_cpu/source/cpu_hal.cpp
```diff
@@ -147,10 +174,40 @@ void cpu_barrier::wait(int num_threads) {
...
 void cpu_hal::kernel_entry(exec_state *exec) {
   direct_kernel_fn kernel = (direct_kernel_fn)exec->kernel_entry;
   kernel((void *)exec->packed_args, exec);
 }
+#else
+void cpu_hal::kernel_entry(exec_state *exec) {
+  direct_kernel_fn kernel = (direct_kernel_fn)exec->kernel_entry;
+
+  // Simple split across x axis by dividing the num groups by the threads and
+  // rounding up. Some threads may get zero.
```
hvdijk, line 187: What are typical use cases here? Should we take into account the possibility that exec->wg.num_groups[0] may be smaller than the number of threads?

clik/external/hal_cpu/source/cpu_hal.cpp
coldav: it could be, it was virtual before, but I don't see that we plan to derive from it, so I think it makes sense.

clik/external/hal_cpu/source/cpu_hal.cpp
coldav: I've removed this from here, as it's not needed but added a comment in arg_pack.cpp where we still use the same function.

clik/external/hal_cpu/source/cpu_hal.cpp
coldav: I've removed it.

clik/external/hal_cpu/source/cpu_hal.cpp
```diff
@@ -32,12 +35,36 @@
...
+  v |= v >> 32;
+  v++;
+  return v;
+}
+
 cpu_hal::cpu_hal(hal::hal_device_info_t *info, std::mutex &hal_lock)
     : hal::hal_device_t(info), hal_lock(hal_lock) {
-  local_mem = (uint8_t *)malloc(local_mem_size);
+#if defined(HAL_CPU_WI_MODE)
+  local_mem = (uint8_t *)std::aligned_alloc(128, local_mem_size);
```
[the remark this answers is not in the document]
coldav: I've left a comment to explain

clik/external/hal_cpu/source/cpu_hal.cpp
coldav: we do - that's why some threads get zero, in this case wg_group_x will be >= to the total threads and won't execute. It's a fairly simplistic algorithm and I think it could perhaps be improved on, but I chose this to match "host". It guarantees that no thread has more than one column different to any other.

Where this is weak is when the number in the first dimension is quite small, but I felt this was good enough for now (and in part this is intended to be a demonstration that is easy to follow).

I've tried to add a bit more to the comments.

clik/external/hal_cpu/source/CMakeLists.txt
coldav: done

clik/external/hal_cpu/include/cpu_hal.h
```diff
@@ -41,7 +44,9 @@ struct cpu_barrier {
 class cpu_hal : public hal::hal_device_t {
  public:
   cpu_hal(hal::hal_device_info_t *info, std::mutex &hal_lock);
-  virtual ~cpu_hal();
+#if HAL_CPU_MODE == HAL_CPU_WI_MODE
+  ~cpu_hal();
```
hvdijk, line 48: I could be misremembering but I think we do need to either keep in the `virtual` (for both modes) or mark `cpu_hal` `final`, as otherwise some compilers will warn.

clik/external/hal_cpu/source/cpu_hal.cpp
hvdijk: Thanks, if this matches what we do elsewhere that's a good enough reason for me.

clik/external/CMakeLists.txt
```diff
@@ -25,6 +25,9 @@ if(CLIK_HAL_DIR STREQUAL "")
   endif()
 endif()
 
+# Clik currently only works with hal cpu in WI mode so disable WG mode
+option(HAL_CPU_WG_MODE "Build Hal CPU in workgroup mode" OFF)
```
hvdijk, line 29: Beware that this only sets the default, it presents the user with a nice toggle that they can flip which will break things. Do we want to pre-emptively error out here if the user makes that change, or do we just leave it to break later on?

clik/external/hal_cpu/source/cpu_hal.cpp
```diff
@@ -190,20 +244,32 @@ bool cpu_hal::kernel_exec(hal::hal_program_t program, hal::hal_kernel_t kernel,
...
+
+  exec.packed_args = (kernel_args_ptr)pack.data();
 
   // Specialize the execution state struct for each thread.
+#if HAL_CPU_MODE == HAL_CPU_WG_MODE
+  size_t num_threads = std::thread::hardware_concurrency();
+#elif HAL_CPU_MODE == HAL_CPU_WI_MODE
   size_t num_threads = work_group_size;
+#else
+#error Either HAL_CPU_MODE_WG or HAL_CPU_MODE_WI needs to be defined.
```
hvdijk, line 270: Outdated error message.

clik/external/hal_cpu/source/hal_main.cpp
```diff
@@ -62,16 +62,23 @@ class cpu_hal_platform : public hal::hal_t {
     const uint64_t local_ram_size = 8 << 20;
     hal_device_info.type = hal::hal_device_type_riscv;
     hal_device_info.word_size = sizeof(uintptr_t) * 8;
-    hal_device_info.target_name = "CPU";
+    hal_device_info.target_name = "ock cpu";
```
hvdijk, line 65: Is this change related to the PR?

clik/external/hal_cpu/source/hal_main.cpp
```diff
@@ -62,16 +62,23 @@ class cpu_hal_platform : public hal::hal_t {
...
     hal_device_info.global_memory_avail =
         global_ram_size - global_mem_max_over_allocation;
     hal_device_info.shared_local_memory_size = local_ram_size;
     hal_device_info.should_link = true;
     hal_device_info.should_vectorize = false;
-    hal_device_info.preferred_vector_width = 0;
+    // TODO: This is slightly arbitrary and based on the "host" target
+    hal_device_info.preferred_vector_width = 128 / (8 * sizeof(uint8_t));
```
hvdijk, line 72: `sizeof(uint8_t)` cannot ever be anything other than 1. It's not clear to me what you're trying to do here.

clik/external/hal_cpu/source/hal_main.cpp
```diff
@@ -62,16 +62,23 @@ class cpu_hal_platform : public hal::hal_t {
...
     hal_device_info.supports_fp16 = false;
     hal_device_info.supports_doubles = true;
+#if HAL_CPU_MODE == HAL_CPU_WG_MODE
+    hal_device_info.max_workgroup_size = 1024;
+#elif HAL_CPU_MODE == HAL_CPU_WI_MODE    
     hal_device_info.max_workgroup_size = 16;
+#else
+#error Either HAL_CPU_MODE should be HAL_CPU_WG_MODE or HAL_CPU_WI_MODE.
```
hvdijk, line 80: Same as above.

clik/external/CMakeLists.txt
coldav: I think it's fine if they want to break it, we should assume they've written the extra code to make it work. (although you are right we could set  a message if they do.

hal/source/arg_pack.cpp
```diff
@@ -30,6 +31,10 @@ namespace util {
 /// @{
 
 namespace {
+// http://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2
+// This will round up to power of 2 for all values except 0, which will
+// result in 0. This is good enough for our purposes where it is based on
+// a non-zero size value.
```
hvdijk, line 37: If a value of `0x8000000000000001` or greater is passed in, the result will be `0`.

clik/external/hal_cpu/source/hal_main.cpp
coldav: it's not absolutely necessary no, but I found it a bit clearer.

hal/source/arg_pack.cpp
coldav: fair enough I can reflect in the comment.

hal/source/arg_pack.cpp
```diff
@@ -135,6 +166,42 @@ bool hal_argpack_t::build(const hal_arg_t *args, uint32_t num_args) {
...
+
+bool hal_argpack_t::expand(size_t num_bytes) {
+  constexpr uint32_t alignment = 128;
+  constexpr uint32_t chunk_size = 128;
+
+  if (!pack || (write_point + num_bytes > pack_alloc_size)) {
+    // Allocate in chunks so we don't continually reallocate
+    pack_alloc_size =
+        ((write_point + num_bytes) + chunk_size - 1) / chunk_size * chunk_size;
+    uint8_t *replacement_pack = reinterpret_cast<uint8_t *>(
```
hvdijk, line 188: Here, `static_cast` and `reinterpret_cast` are equivalent, in which case `static_cast` is probably better as it provides more checks that the cast makes sense.

hal/source/arg_pack.cpp
```diff
@@ -135,6 +166,42 @@ bool hal_argpack_t::build(const hal_arg_t *args, uint32_t num_args) {
...
+        std::aligned_alloc(alignment, pack_alloc_size));
+    if (!replacement_pack) {
+      return false;
+    }
+    if (pack) {
+      if (write_point) {
+        // copy from old to new and free old one
+        memcpy(replacement_pack, pack, write_point);
```
hvdijk, line 196: We seem to be flipping between global-namespace and `std`-namespace C standard library functions, we should set a policy. No need to update all existing code, but nice to keep it consistent in new code.

clik/external/hal_cpu/source/hal_main.cpp
coldav: you are right, but it does reflect what the host target does. The comment there says   // default to 128 bit (16 bytes), which I could add but I'm not sure if that is more help.

modules/compiler/cookie/{{cookiecutter.target_name}}/source/info.cpp
```diff
@@ -38,7 +38,7 @@ namespace {{cookiecutter.target_name}} {
   dma_optimizable = true;
   supports_deferred_compilation = false;
   scalable_vector_support = {{cookiecutter.scalable_vector}};
-  kernel_debug = false;
+  kernel_debug = true;
```
hvdijk, line 41: Is this meant to stay in?

clik/external/hal_cpu/source/cpu_hal.cpp
```diff
@@ -190,20 +244,32 @@ bool cpu_hal::kernel_exec(hal::hal_program_t program, hal::hal_kernel_t kernel,
   }
   exec.kernel_entry = (entry_point_fn)kernel;
   exec.flags = flags;
+
+  hal::util::hal_argpack_t pack(get_word_size() * 8);
```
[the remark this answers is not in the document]
coldav: I can change get_word_size to get_word_size_in_bits() here

modules/compiler/cookie/{{cookiecutter.target_name}}/source/info.cpp
coldav: yes, it fixes the part of check-ComputeAorta that tests with '-g'

hal/source/arg_pack.cpp
coldav: fair I can use std::memcpy here

clik/external/hal_cpu/source/hal_main.cpp
hvdijk: Oh, I hadn't even realised that this `8` was again due to bits vs bytes. That comment would have helped me, thanks.

clik/external/hal_cpu/source/cpu_hal.cpp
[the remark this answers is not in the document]
hvdijk: If `get_word_size()` returns the word size in bytes, it's `hal_argpack_t`'s constructor's parameter that should be renamed to `word_size_in_bits`, if I'm understanding correctly what's happening here (and I'm not entirely sure I am)

clik/external/CMakeLists.txt
hvdijk: Sure, that sounds reasonable to me (with or without the message).

modules/compiler/cookie/{{cookiecutter.target_name}}/source/info.cpp
hvdijk: Ah, thanks, this is a confusing member name but its comment (kernel supports debugging, which I hadn't seen) helps explain.

clik/external/hal_cpu/source/cpu_hal.cpp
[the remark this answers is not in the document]
coldav: sure - yes, the parameter and usages within argpack should change.

hal/source/arg_pack.cpp
hvdijk: `0x8000000000000000` is 2^63, not 2^32. Values from 2^32 to 2^63 should be handled correctly, I think.

coldav: you are right, fixed.

PietroGhg (approved):

clik/external/hal_cpu/include/cpu_hal.h
```diff
@@ -25,6 +25,9 @@
 
 #include "hal.h"
 
+#define HAL_CPU_WI_MODE 0
+#define HAL_CPU_WG_MODE 1
```
hvdijk, line 29: Due to the way the preprocessor works, if `HAL_CPU_MODE` is not defined, `#if HAL_CPU_MODE == HAL_CPU_WI_MODE` will be true and the `#error` that you added will never fire. Can you change the values from 0/1 to 1/2? (Pre-approving)

hvdijk (approved):

PR #406, opened 2024-03-18
Fix vkEnumerateInstanceExtensionProperties.

hvdijk:
# Overview

Fix vkEnumerateInstanceExtensionProperties.

# Reason for change

GCC 13 rightly complains about an out of bounds pointer: when *pPropertyCount > vk::instance_extensions.size(), merely computing vk::instance_extensions.begin() + *pPropertyCount has undefined behavior and trying to cover that up via std::min afterwards is too late and not guaranteed to do what is intended.

# Description of change

Check correctly.

While looking into this I also found that we did not match the spec in that we did not return VK_INCOMPLETE when the passed in size was too small, and did not set *pPropertyCount when the passed in size was too large.

# Checklist

### DIFF
--- a/source/vk/test/UnitVK/source/CreateInstance.cpp
+++ b/source/vk/test/UnitVK/source/CreateInstance.cpp
@@ -85,10 +85,22 @@ TEST_F(CreateInstance, DefaultExtension) {
                                    nullptr, &extensionCount, nullptr));
 
   if (0 < extensionCount) {
+    uint32_t storedExtensionCount;
     std::vector<VkExtensionProperties> extensionProperties(extensionCount);
-    ASSERT_EQ_RESULT(VK_SUCCESS,
-                     vkEnumerateInstanceExtensionProperties(
-                         nullptr, &extensionCount, extensionProperties.data()));
+
+    storedExtensionCount = 0;
+    ASSERT_EQ_RESULT(VK_INCOMPLETE, vkEnumerateInstanceExtensionProperties(
+                                        nullptr, &storedExtensionCount,
+                                        extensionProperties.data()));
+    ASSERT_EQ(0, storedExtensionCount);
+
+    // Test that we do not overflow, not even on 32-bit platforms, by
+    // effectively treating this as -1.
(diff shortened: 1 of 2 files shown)


source/vk/test/UnitVK/source/CreateInstance.cpp
```diff
@@ -85,10 +85,20 @@ TEST_F(CreateInstance, DefaultExtension) {
...
-                     vkEnumerateInstanceExtensionProperties(
-                         nullptr, &extensionCount, extensionProperties.data()));
+
+    storedExtensionCount = 0;
+    ASSERT_EQ_RESULT(VK_INCOMPLETE, vkEnumerateInstanceExtensionProperties(
+                                        nullptr, &storedExtensionCount,
+                                        extensionProperties.data()));
+    ASSERT_EQ(0, storedExtensionCount);
+
+    storedExtensionCount = 0xffffffff;
```
coldav, line 97: should this not be no bigger than the size of the vector? (I appreciate it shrinks to the number of extensions, but even so).

hvdijk: In non-test code, for writing sensible code, absolutely, this is silly. But it's technically valid, and as a test, this is important to cover one of the cases that failed: on 32-bit platforms, depending on the optimization level, `vk::instance_extensions.begin() + *pPropertyCount` would end up as `vk::instance_extensions.begin() - 1` (ignoring some details) and therefore do the wrong thing because `std::min()` would not consider it greater than `vk::instance_extensions.end()`.

coldav: could we make the std::vector that big then? If you prefer not to, then I'm fine as is.

coldav: fair enough, it makes sense -  maybe a comment explaining why 0xffffffff

hvdijk: Yep, a comment makes sense, will do.

PR #408, opened 2024-03-19
DRAFT: FOR COMMENTS Support verbosity in cmake output

coldav:
# Overview

For message cmake commands replace with ca_message and support a verbosity level which can be changed using CA_CMAKE_VERBOSE_LEVEL(0 none, 1 summary, 2 detail). Including from LLVM will by default set to 0.
# Reason for change

Including from LLVM was noisy

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,6 +59,27 @@ endif()
 # Add cmake modules directory to the path.
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 
+set(CA_CMAKE_VERBOSE_LEVEL ""  CACHE STRING "cmake verbosity level 2 detailed (default), 1 summary, 0 none")
+set(OCK_CMAKE_VERBOSE_LEVEL "2")
+
+# Check if we are in an LLVM Tree.
+# This is done by checking for the LLVMCore target which does not exist
+# until we import LLVM.
+# This only happens of we have not imported or are fetched by LLVM via
+# FetchContent or as an external project. In this case we set a cmake
+# variable OCK_IN_LLVM_TREE to TRUE.
+set(OCK_IN_LLVM_TREE FALSE)
+if (TARGET LLVMCore)
+  set(OCK_IN_LLVM_TREE TRUE)
+  # We always reduce verbosity level if included in LLVM tree
+  set(OCK_CMAKE_VERBOSE_LEVEL "0")
+endif()
+
+# Even if in LLVM tree we support overriding of the verbosity level
+if (NOT CA_CMAKE_VERBOSE_LEVEL STREQUAL "")
+  set(OCK_CMAKE_VERBOSE_LEVEL ${CA_CMAKE_VERBOSE_LEVEL})
+endif()
+
 # Bring in our ca_option macro for declaring options.
 include(CAOption)
 
@@ -156,17 +177,6 @@ foreach(OPTIONS_MODULE ${OPTIONS_MODULES})
     include(${OPTIONS_MODULE})
   endif()
 endforeach()
-
-# Check if we are in an LLVM Tree.
-# This is done by checking for the LLVMCore target which does not exist
-# until we import LLVM.
-# This only happens of we have not imported or are fetched by LLVM via
-# FetchContent or as an external project. In this case we set a cmake
-# variable OCK_IN_LLVM_TREE to TRUE.
-set(OCK_IN_LLVM_TREE FALSE)
-if (TARGET LLVMCore)
-  set(OCK_IN_LLVM_TREE TRUE)
-endif()
 
 if(CA_RUNTIME_COMPILER_ENABLED AND NOT OCK_IN_LLVM_TREE)
   include(ImportLLVM)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -59,6 +59,27 @@ endif()
...
+if (TARGET LLVMCore)
+  set(OCK_IN_LLVM_TREE TRUE)
+  # We always reduce verbosity level if included in LLVM tree
+  set(OCK_CMAKE_VERBOSE_LEVEL "0")
+endif()
+
+# Even if in LLVM tree we support overriding of the verbosity level
+if (NOT CA_CMAKE_VERBOSE_LEVEL STREQUAL "")
+  set(OCK_CMAKE_VERBOSE_LEVEL ${CA_CMAKE_VERBOSE_LEVEL})
+endif()
```
hvdijk, line 81: I think this should be a cached variable so that if CMake decides it needs to re-run itself, the previously set verbosity level gets picked up. It also means we only need a single variable because cache variables by default (if `FORCE` is not used) do not override user settings.

CMakeLists.txt
```diff
@@ -59,6 +59,27 @@ endif()
...
+# until we import LLVM.
+# This only happens of we have not imported or are fetched by LLVM via
+# FetchContent or as an external project. In this case we set a cmake
+# variable OCK_IN_LLVM_TREE to TRUE.
+set(OCK_IN_LLVM_TREE FALSE)
+if (TARGET LLVMCore)
+  set(OCK_IN_LLVM_TREE TRUE)
+  # We always reduce verbosity level if included in LLVM tree
+  set(OCK_CMAKE_VERBOSE_LEVEL "0")
+endif()
```
hvdijk, line 76: We could just keep the default as always verbose, and override it from within DPC++.

cmake/CAOption.cmake
```diff
@@ -71,11 +72,11 @@ macro(ca_option name type description default)
   # Setup the main option of this macro.
   if(${type} STREQUAL "BOOL")
     option(${name} ${description} ${default})
-    message(STATUS "oneAPI Construction Kit ${description}: ${${name}}")
+    ca_message(STATUS CA_VERBOSE_DETAIL "oneAPI Construction Kit ${description}: ${${name}}")
```
hvdijk, line 75: `STATUS` already indicates the verbosity, and CMake has ways of automatically suppressing noise based on the requested verbosity level (`CMAKE_MESSAGE_LOG_LEVEL`). We could use that.

cmake/CAUtils.cmake
```diff
@@ -0,0 +1,30 @@
...
+function(ca_message mode verbosity_level ...)
+    set(CA_MESSAGE_VERBOSITY_LEVEL 0)
+    if (${verbosity_level} STREQUAL  "CA_VERBOSE_SUMMARY")
+        set(CA_MESSAGE_VERBOSITY_LEVEL 1)
+    elseif(${verbosity_level} STREQUAL  "CA_VERBOSE_DETAIL")
+        set(CA_MESSAGE_VERBOSITY_LEVEL 2)
+    endif()
+    list(REMOVE_AT ARGV 0)
+    list(REMOVE_AT ARGV 0)
+    if (${CA_MESSAGE_VERBOSITY_LEVEL} LESS_EQUAL ${OCK_CMAKE_VERBOSE_LEVEL})
```
hvdijk, line 27: CMake's `if` command is weird and in most cases, variables should not be expanded explicitly, because `if` will already try to expand them and you run the risk of double expansion. `if(CA_MESSAGE_VERBOSITY_LEVEL LESS_EQUAL OCK_CMAKE_VERBOSE_LEVEL)` should work.

cmake/CAOption.cmake
coldav: I thought about that, but it felt that we should output all warnings by default, and STATUS could have two levels i.e. summary information and detail information. The only one I felt has real meaning for me is the LLVM version when i run it. However I am open-minded we could just use STATUS etc as you suggest.

CMakeLists.txt
coldav: I did think about that, I was worried that we end up adding loads and loads of cmake options before including it and maybe this was another way.

CMakeLists.txt
coldav: I did initially go that way, but it seemed to require the fetcher to set the variable, rather than doing it by default in "inside llvm".

CMakeLists.txt
hvdijk: If DPC++ native CPU wants loads of things different from what we want as the default, there's probably a bigger problem that we should discuss. I would expect that a list of just those things where DPC++ wants something different to be relatively short, shorter than repeated `OCK_IN_LLVM_TREE` checks. Is that not the case?

cmake/CAOption.cmake
hvdijk: The ones that you mark as verbose here do have meaning for me, I check them almost every time I need to re-run CMake to make sure I didn't leave some extension off that I will be trying to test.

coldav: CMAKE_MESSAGE_LOG_LEVEL would be across LLVM too, whereas I think they want to only suppress OCK messages

hvdijk: CMake variables have directory scope. If set in a subdirectory, it should only affect that subdirectory without affecting the rest of LLVM.

coldav: I did try that, but I'll try again.

PR #411, opened 2024-03-21
Fix windows failure due to std::aligned_alloc

coldav:
# Overview

Windows does not support std::aligned_alloc, so use _aligned_malloc instead.

# Reason for change

Windows does not support std::aligned_alloc, so use _aligned_malloc instead.


hal/source/arg_pack.cpp
```diff
@@ -195,7 +201,11 @@ bool hal_argpack_t::expand(size_t num_bytes) {
         // copy from old to new and free old one
         std::memcpy(replacement_pack, pack, write_point);
       }
+#if defined(_MSC_VER)
+      free(pack);
```
hvdijk, line 205: If I understand correctly, `_aligned_malloc` needs to be matched with `_aligned_free`.

coldav: done

hvdijk (approved):

PR #413, opened 2024-03-22
Address clang-tidy issues.

hvdijk:
# Overview

Address clang-tidy issues.

# Reason for change

In cmake/AddCA.cmake, in add_ca_tidy, we try to add clang-tidy targets for all source files, but for source files specified with relative paths, we would resolve them using the wrong base directory.

# Description of change

This change uses file(REAL_PATH) with an explicit BASE_DIRECTORY to convert to an absolute path and add the intended tidy rule.

Many files require updates to address the warnings we now see from clang-tidy.

# Checklist


coldav (approved):

coldav (approved):

source/cl/test/UnitCL/source/ktst_regression_03.cpp
```diff
@@ -609,10 +609,18 @@ struct StrideMisaligned {
...
+    // representations because of its global_size member: clang has an open
+    // FIXME for supporting vectors. The static_assert below should cover any
+    // scenario where the compiler still adds padding inside the struct, so we
+    // can suppress the clang-tidy warning.
+    return memcmp  // NOLINT(bugprone-suspicious-memory-comparison,cert-exp42-c,cert-flp37-c)
+           (this, &rhs, sizeof(StrideMisaligned)) == 0;
   }
 };
 
+static_assert(sizeof(StrideMisaligned) == sizeof(cl_ulong4) + sizeof(cl_uint));
```
coldav, line 622: if sizeof(StrideMisaligned) is wrong because sizeof(cl_ulong4) was wrong we may still see issues, I wonder if it's better to do sizeof(cl_ulong)*4 + sizeof(cl_uint).
I'm fine with this though if you want to leave it.

hvdijk: I think you're right, that's probably better. Will update.

hvdijk: Done, https://github.com/codeplaysoftware/oneapi-construction-kit/compare/bcd8d5649c6daf6cf11645929b6813162255e3de..524de2d5147c33a57b0e9132b5414c8f57d072a4. As I expect to now merge this as is I have also done the squashing, but included the diff link here for easier review.

PR #424, opened 2024-04-01
Update riscv ignore file.

hvdijk:
# Overview

Update riscv ignore file.

# Reason for change

The test_select and test_bruteforce tests are the only tests that take too long to run.

# Description of change

points, etc.*

This change allows us to use cts-3.0-online-ignore-linux-riscv.csv with the OpenCL CTS CSV files included there.

# Checklist


source/cl/scripts/cts-3.0-online-ignore-linux-riscv.csv
```diff
@@ -1,1310 +1,3 @@
...
-SPIRV new,spirv_new/test_spirv_new ext_cl_khr_spirv_no_integer_wrap_decoration_fmul_uint
-SPIRV new,spirv_new/test_spirv_new ext_cl_khr_spirv_no_integer_wrap_decoration_fshiftleft_uint
-SPIRV new,spirv_new/test_spirv_new op_vector_times_scalar_double
-
-SPIR (cl_khr_spir),spir/test_spir conversions_double no-unzip
-Mipmaps (Kernel),images/kernel_read_write/test_image_streams test_mipmaps CL_FILTER_NEAREST
-Mipmaps (clCopyImage),images/clCopyImage/test_cl_copy_images
-Mipmaps (clReadWriteImage),images/clReadWriteImage/test_cl_read_write_images test_mipmaps
+Select,select/test_select
+Math,math_brute_force/test_bruteforce -w
```
coldav, line 3: do we really not want to run bruteforce at all now? Is it not feasible to split this up now? same goes for test_select.

hvdijk: If test_select and test_bruteforce take options that let us exclude the subtests that take too long, we could use that, and make our cityrunner to accept multiple CSV files so that we can say "use upstream quick.csv, except ignore-riscv.csv, but add custom-riscv.csv". I think that would need a cityrunner update in a separate PR, I don't think we should include it in here, but if we decide to do that, we could merge that first and avoid any window where we ignore these tests.

coldav: fair enough, I was just worried that we miss important accuracy fails. Maybe a jira to look into whether we should restore some of what we've lost

hvdijk: Actually, I had a thought about this now, I think we can cover these with minimal effort. Will update.

hvdijk: Have updated, hope to verify today that the changes to city_runner work. Can pull those to a separate PR if we prefer.

source/cl/scripts/cts-3.0-online-extra-linux-riscv.csv
```diff
@@ -0,0 +1,2 @@
+Select,select/test_select -w
+Math,math_brute_force/test_bruteforce -w\[512]
```
coldav, line 2: So we are replacing test_select with `test_select -w`?
what does the \[512] bit do and how does this make it better than "test_bruteforce -w"?

hvdijk: It increases the Wimpy Reduction Factor, basically making it even wimpier. Hopefully by enough.

coldav: why does it need \[ but not \]?

hvdijk: I assume GitHub ate your backslash there. Shell quoting rules, `-w[512]` looks for files named `-w5`, `-w1`, `-w2` in the current directory. We don't want that, we want to just pass the literal argument `-w[512]`.

coldav (approved):

PR #428, opened 2024-04-08
Minor changes to cpu hal to enable wrapping for remote hal access

coldav:
# Overview

Minor changes to cpu hal to enable wrapping for remote hal access

# Reason for change

Wrapping of the cpu hal with a server and remote access from a client requires some changes to make it work. 

# Description of change

The creation of the device info being a function that can be called externally through a static function. This allows us to use clc without having to use the client/server relation to get the device info.

Making the linker script (which is empty) defined within the static function rather than baking it. Technically we could bake it but we wanted to simplify the server and build the cpu hal files directly rather than having to build it as a shared library and complicate the dependencies.

### DIFF
--- a/clik/external/hal_cpu/include/cpu_hal.h
+++ b/clik/external/hal_cpu/include/cpu_hal.h
@@ -48,6 +48,10 @@ class cpu_hal final : public hal::hal_device_t {
   ~cpu_hal();
 #endif
 
+  // Set up the hal device info - this is done as a static so that other classes such
+  // as hal_client can set up the desired information directly
+  static hal::hal_device_info_t setup_cpu_hal_device_info();
+
   size_t get_word_size() const { return sizeof(uintptr_t); }
 
   // find a specific kernel function in a compiled program
--- a/clik/external/hal_cpu/source/CMakeLists.txt
+++ b/clik/external/hal_cpu/source/CMakeLists.txt
@@ -34,11 +34,6 @@ target_include_directories(hal_cpu PRIVATE
 
 target_link_libraries(hal_cpu hal_common dl pthread)
 
-hal_add_baked_data(hal_cpu
-    hal_cpu_linker_script
-    linker_script.h
-    ${HAL_CPU_SOURCE_DIR}/include/device/program.lds)
-
 set_target_properties(hal_cpu PROPERTIES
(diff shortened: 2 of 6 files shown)


clik/external/hal_cpu/source/CMakeLists.txt
```diff
@@ -34,11 +34,6 @@ target_include_directories(hal_cpu PRIVATE
 
 target_link_libraries(hal_cpu hal_common dl pthread)
 
-hal_add_baked_data(hal_cpu
-    hal_cpu_linker_script
-    linker_script.h
-    ${HAL_CPU_SOURCE_DIR}/include/device/program.lds)
```
hvdijk, line 40: This makes it so that `program.lds` is unused, right? Should it be removed?

clik/external/hal_cpu/include/cpu_hal.h
```diff
@@ -48,6 +48,10 @@ class cpu_hal final : public hal::hal_device_t {
   ~cpu_hal();
 #endif
 
+  // Set up the hal device info - this is done as a static so that other classes such
+  // as hal_client can set up the desired information directly
+  static void setup_cpu_hal_device_info(hal::hal_device_info_t &hal_device_info);
```
hvdijk, line 53: Agreed that using a static helper function for this is a good idea, but can it return `hal_device_info_t` by value so it is clear it does not rely on the caller set up anything in it in advance?

clik/external/hal_cpu/source/CMakeLists.txt
hvdijk: Oh, I hadn't seen that, `git grep program.lds clik` doesn't find any other references in there. Is it the things in `doc/overview/tutorials/` that you're referring to?

clik/external/hal_cpu/include/cpu_hal.h
coldav: sure

clik/external/hal_cpu/source/CMakeLists.txt
coldav: you are right, I think it may have used it in the past, but not now

hvdijk (approved):

PR #429, opened 2024-04-09
Return correct clEnqueueCommandBufferKHR errors

EwanC:
# Overview

We aren't returning the correct errors as specified by `clEnqueueCommandBufferKHR`.

Issues we're found based on testing of CTS PR https://github.com/KhronosGroup/OpenCL-CTS/pull/1931 that verifies these errors

# Reason for change

Non-conformant implementation of  `clEnqueueCommandBufferKHR`

# Description of change

1. Update some existing `clEnqueueCommandBufferKHR` return codes to the correct value.
2. Use helper function in `clEnqueueCommandBufferKHR` to verify event wait-list.
3. Add check for `CL_INVALID_COMMAND_QUEUE` to `clEnqueueCommandBufferKHR`.

- Updated UnitCL tests to check against the specification defined value.
- Modified these checks to use gtest `EXPECT` rather than `ASSERT` because if the `ASSERT` fails and returns immediately then we won't release the command-queue created by the test, resulting in a leak.
- Still one fail left when testing against CTS PR, for a spec related reason i've [commented on](https://github.com/KhronosGroup/OpenCL-CTS/pull/1931#discussion_r1557531816)

# Checklist

### DIFF
--- a/source/cl/source/extension/source/khr_command_buffer.cpp
+++ b/source/cl/source/extension/source/khr_command_buffer.cpp
@@ -1251,14 +1251,23 @@ CL_API_ENTRY cl_int CL_API_CALL clEnqueueCommandBufferKHR(
   const tracer::TraceGuard<tracer::OpenCL> guard("clEnqueueCommandBufferKHR");
 
   // Verify queue arguments.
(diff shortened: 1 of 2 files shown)


hvdijk (approved):
Thanks, looks good!

PR #430, opened 2024-04-09
[CMake] Fix check for LLVM_ENABLE_ZLIB, add _ZSTD.

hvdijk:
# Overview

[CMake] Fix check for LLVM_ENABLE_ZLIB, add _ZSTD.

# Reason for change

In cross builds, we check for a LLVM_ENABLE_ZLIB mismatch between host LLVM and target LLVM because it affects bitcode compatibility. This mismatch was not fully reliable: LLVM sometimes sets LLVM_ENABLE_ZLIB to other values than the values we check for, in particular it may set it to nothing.

In cross builds, LLVM_ENABLE_ZSTD has the same problem as LLVM_ENABLE_ZLIB: if the host LLVM has that enabled, the target LLVM needs it too.

# Description of change

This change allows any arbitrary value to be picked up and does the same check for LLVM_ENABLE_ZSTD.

# Checklist

### DIFF
--- a/doc/developer-guide.md
+++ b/doc/developer-guide.md
@@ -755,6 +755,7 @@ cmake . -GNinja \
   -DLLVM_HOST_TRIPLE=arm-unknown-linux-gnu \
   -DLLVM_DEFAULT_TARGET_TRIPLE=arm-unknown-linux-gnu \
   -DLLVM_ENABLE_ZLIB=OFF \
+  -DLLVM_ENABLE_ZSTD=OFF \
   -DLLVM_TABLEGEN=$LLVMNativeInstall/bin/llvm-tblgen \
   -DCLANG_TABLEGEN=$LLVMNativeBuild/bin/clang-tblgen
 ```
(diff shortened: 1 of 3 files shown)


doc/developer-guide.md
```diff
@@ -755,6 +755,7 @@ cmake . -GNinja \
   -DLLVM_HOST_TRIPLE=arm-unknown-linux-gnu \
```
coldav, line 755: Possibly worth adding to the changelog, but feel free to ignore.

hvdijk: Good to think about, but in `CHANGELOG.md`, we document the things that users need to take into account when switching to a newer version. This isn't like that, users where zstd was not detected aren't affected and can continue to not specify `LLVM_ENABLE_ZSTD`, users where zstd was detected would already have had to work around it by manually specifying `LLVM_ENABLE_ZSTD=OFF` (or `FALSE` or otherwise false) so this tells them they would already have known and don't need to specifically take into account again now. If you're okay with leaving it out, I'll leave it out :)

PR #434, opened 2024-04-11
Add hal transmitter and socket support for remote hal

coldav:
# Overview

Add hal transmitter and socket support for remote hal

# Reason for change

part of the support needed for communications for remote hal support

# Description of change

Add base class hal_transmitter which can be used for simple send/ receive operations of data. This can be used by client and servers for different models of communication.

Also added a socket version which will be the default use case.

### DIFF
--- a/hal/hal_remote/source/hal_socket_transmitter.cpp
+++ b/hal/hal_remote/source/hal_socket_transmitter.cpp
@@ -1,0 +1,163 @@
+// Copyright (C) Codeplay Software Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+// Exceptions; you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+// WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+// License for the specific language governing permissions and limitations
+// under the License.
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <hal_remote/hal_socket_transmitter.h>
+#include <hal_remote/hal_transmitter.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <cstdio>
+
+namespace hal {
+hal_socket_transmitter::~hal_socket_transmitter() {
+  if (sock) {
+    close(sock);
+  }
+  if (fd_to_use && fd_to_use != sock) {
+    close(fd_to_use);
+  }
+}
+
+hal_socket_transmitter::error_code hal_socket_transmitter::start_server(
+    bool print_port) {
+  if (!setup_connection_done) {
+    hal_socket_transmitter::error_code res = setup_connection(true);
+    if (res != 0) {
+      return res;
+    }
+    setup_connection_done = true;
+  }
+  if (listen() != 0) {
+    last_error = hal_socket_transmitter::listen_failed;
+    return last_error;
+  }
+  if (print_port) {
+    printf("Listening on port %d\n", get_port());
+    fflush(stdout);
+  }
+  if (accept() == -1) {
+    last_error = hal_socket_transmitter::accept_failed;
+    return last_error;
+  }
+  last_error = hal_socket_transmitter::success;
+  return last_error;
+}
+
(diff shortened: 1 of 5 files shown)


hal/hal_remote/source/hal_socket_transmitter.cpp
```diff
@@ -0,0 +1,163 @@
...
+  }
+}
+
+hal_socket_transmitter::error_code hal_socket_transmitter::setup_connection(
+    bool server) {
+  // @return 0 if success, otherwise -1 and errno set
+  sock = socket(AF_INET, SOCK_STREAM, IPPROTO_TCP);
+  server_address.sin_family = AF_INET;
+  server_address.sin_port = htons(port_requested);
+  server_address.sin_addr.s_addr = INADDR_ANY;
```
hvdijk, line 128: Always binding to all interfaces, with no way to override, seems like a possible security concern.

hal/hal_remote/source/hal_socket_transmitter.cpp
```diff
@@ -0,0 +1,163 @@
...
+    is_connected = true;
+    return hal_socket_transmitter::success;
+  } else {
+    return hal_socket_transmitter::connect_failed;
+  }
+}
+
+hal_socket_transmitter::error_code hal_socket_transmitter::setup_connection(
+    bool server) {
+  // @return 0 if success, otherwise -1 and errno set
```
hvdijk, line 124: Does a `@return` comment work here?

hal/hal_remote/source/hal_socket_transmitter.cpp
coldav: what would you prefer here? I'm fairly new to sockets.

hvdijk: See e.g. Apache's configuration, https://httpd.apache.org/docs/2.4/bind.html, where users can specify the IP address to listen to in addition to the port. How this should look in code I'm not entirely sure of, it depends on how this PR later ends up used, which I have not looked at yet.

coldav: I'm happy to extend this to an optional address and default to this (and document that) - would you be happy with that?

hvdijk: That seems okay, though keep in mind that depending on the address, we may also need `AF_INET6` instead of `AF_INET`.

Question: since the purpose of the HAL process is to take untrusted code from a remote client and execute it, what is the overall security going to look like?

coldav: it does add risk, but the user should be aware that this does remotely run processes, so your suggestion of limiting it to a particular IP address sounds good. Also it's not got any superuser priviledges.
For that matter, you can override the libhal as a shared library and grab things that way.

hal/hal_remote/source/hal_socket_transmitter.cpp
```diff
@@ -0,0 +1,220 @@
...
+    if (!server) {
+      current_port = port_requested;
+    }
+  }
+  setup_connection_done = true;
+  return hal_socket_transmitter::success;
+}
+
+int hal_socket_transmitter::listen() {
+  if (const int res = ::listen(sock, 1) != 0) {
```
hvdijk, line 196: This means `res` is initialised to `::listen(sock, 1) != 0`, i.e. it will always be zero or one, rather than returning the result of `::listen(sock, 1)`. Also, the `if (res != 0) return res; else return 0;` logic looks overly complicated, that amounts to just `return res;`.

But given that the API here is "1 is failure, 0 is success", whereas below for `accept()`, it's "-1 is failure, anything else, possibly zero, is the file descriptor that must not be used", I wonder if it's worth updating both to just return `bool`?

hvdijk: This pattern of `const int res = ... != 0` appears in a few other places where that does not look intentional.

hal/hal_remote/source/hal_socket_transmitter.cpp
```diff
@@ -0,0 +1,220 @@
...
+    return hal_socket_transmitter::connect_failed;
+  }
+}
+
+hal_socket_transmitter::error_code hal_socket_transmitter::setup_connection(
+    bool server) {
+  // don't support port 0
+  if (port_requested == 0) {
+    if (debug_enabled()) {
+      (void)fprintf(stderr, "port 0 requested. This is disallowed\n");
```
hvdijk, line 126: Given that port 0 is the default, this may not actually have been requested, would an error message along the lines of "The port to listen on must be specified" work?

PR #435, opened 2024-04-15
Enable cross compilation when building OCK in-tree.

PietroGhg:
# Overview

CMake changes that allow cross compiling OCK when building with LLVM in-tree.

# Reason for change

If `CMAKE_CROSSCOMPILING` is set and we are in-tree, it means that OCK is being cross compiled as part of LLVM, and so we can use `LLVM_NATIVE_TOOL_DIR` to find the tools required to compile the builtins. Moreover, since when building in tree we don't use `ImpotLLVM`, the `LLVM_PACKAGE_VERSION` variable was not defined, which lead to a configure error when checking the version string for the builtin tools.

# Description of change

Uses `LLVM_NATIVE_TOOL_DIR` to find builtin tools, sets `LLVM_PACKAGE_VERSION` when cross compiling in-tree.

### DIFF
--- a/modules/compiler/builtins/CMakeLists.txt
+++ b/modules/compiler/builtins/CMakeLists.txt
@@ -35,9 +35,13 @@ include(cmake/FindBuiltinsTools.cmake)
 if(NOT CA_EXTERNAL_BUILTINS AND CA_RUNTIME_COMPILER_ENABLED)
   if(NOT DEFINED CA_BUILTINS_TOOLS_DIR)
     if(CMAKE_CROSSCOMPILING)
-      message(FATAL_ERROR
-        "Cross compiled builds requires CA_EXTERNAL_BUILTINS or "
-        "CA_BUILTINS_TOOLS_DIR to be set!")
+      if(EXISTS ${LLVM_NATIVE_TOOL_DIR})
+        set(CA_BUILTINS_TOOLS_DIR ${LLVM_NATIVE_TOOL_DIR})
+      else()
+        message(FATAL_ERROR
+          "Cross compiled builds requires CA_EXTERNAL_BUILTINS or "
+          "CA_BUILTINS_TOOLS_DIR to be set!")
+      endif()
     # We might be in-tree and not be importing an LLVM directory, in which case
     # don't assume any builtins directory - we'll take the tools required to
     # build the builtins directly from the corresponding LLVM targets.
--- a/modules/compiler/builtins/cmake/FindBuiltinsTools.cmake
+++ b/modules/compiler/builtins/cmake/FindBuiltinsTools.cmake
@@ -60,7 +60,7 @@ function(find_builtins_tools tools_dir)
       )
       set(BUILTINS_LLVM_CMAKE ${BUILTINS_LLVM_CMAKE}/LLVMConfig.cmake)
     endif()
-    if(NOT EXISTS BUILTINS_LLVM_CMAKE)
+    if(NOT EXISTS ${BUILTINS_LLVM_CMAKE})
       message(
         WARNING
         "builtins tools installation does not have necessary cmake modules; "
@@ -96,12 +96,13 @@ function(find_builtins_tools tools_dir)
       IMPORTED_LOCATION ${BUILTINS_LINKER})
   endif()
 
+  set(llvm_version "${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}")
+
   # Check that the version of the compiler building the builtins matches the
   # version of the compiler linked into the runtime.
   execute_process(COMMAND ${BUILTINS_COMPILER} --version
     OUTPUT_VARIABLE version_string RESULT_VARIABLE result)
   if(result EQUAL 0)
-    string(REPLACE "svn" "" llvm_version ${LLVM_PACKAGE_VERSION})
     string(REGEX MATCH "clang version [0-9]+\\.[0-9]+\\.[0-9]"
       version_string ${version_string})
     string(REPLACE "clang version " "" version_string ${version_string})


modules/compiler/builtins/cmake/FindBuiltinsTools.cmake
```diff
@@ -60,7 +67,7 @@ function(find_builtins_tools tools_dir)
       )
       set(BUILTINS_LLVM_CMAKE ${BUILTINS_LLVM_CMAKE}/LLVMConfig.cmake)
     endif()
-    if(NOT EXISTS BUILTINS_LLVM_CMAKE)
+    if(NOT EXISTS ${BUILTINS_LLVM_CMAKE})
```
hvdijk, line 70: This change looks suspicious. The `if` command is special and takes variable names, and will implicitly check their values. Why is `${`...`}` needed?

modules/compiler/builtins/cmake/FindBuiltinsTools.cmake
```diff
@@ -29,6 +29,13 @@ include(CAPlatform)  # Bring in CA_HOST_EXECUTABLE_SUFFIX.
 # look for the builtins tools, if it is an empty string, the function will
 # attempt to use eventual in tree targets of the builtins tools.
 function(find_builtins_tools tools_dir)
+  if (CMAKE_CROSSCOMPILING AND OCK_IN_LLVM_TREE)
+    # If we are building OCK in LLVM TREE, ImportLLVM was not included and
+    # so LLVM_PACKAGE_VERSION is not defined.
+    set(LLVM_PACKAGE_VERSION
```
coldav, line 35: Is this just for the check at the bottom? Maybe move it there or add explanation of how the version comes into it.

modules/compiler/builtins/CMakeLists.txt
```diff
@@ -35,9 +35,14 @@ include(cmake/FindBuiltinsTools.cmake)
...
-      message(FATAL_ERROR
-        "Cross compiled builds requires CA_EXTERNAL_BUILTINS or "
-        "CA_BUILTINS_TOOLS_DIR to be set!")
+      if(EXISTS ${LLVM_NATIVE_TOOL_DIR})
+        set(CA_BUILTINS_TOOLS_DIR ${LLVM_NATIVE_TOOL_DIR})
+      else()
+        message(FATAL_ERROR
+          "Cross compiled builds requires CA_EXTERNAL_BUILTINS or "
+          "CA_BUILTINS_TOOLS_DIR to be set, or set LLVM_NATIVE_TOOL_DIR"
+          "if you are building in-tree.")
```
hvdijk, line 44: `LLVM_NATIVE_TOOL_DIR` is a DPC++-specific variable and I don't think we fully support in-tree builds done by users just yet, so I think it's fine to use the variable if it is set, but we probably should not include it in the message we tell users (and it might be good to include a CMake comment that this is specifically for DPC++)

modules/compiler/builtins/cmake/FindBuiltinsTools.cmake
PietroGhg: I was confused by this as well, without `${...}` I get the warning `builtins tools installation does not...`, I tried printing out the content of `BUILTINS_LLVM_CMAKE` on my side, and it is a valid absolute path. Adding the `${...}` removed the warning, but this change is not strictly necessary for this PR, so I've fine with reverting it and looking more into this.

hvdijk: Ignore my comment, this change is correct, please keep it in. The special rule for `if` does not apply to `if([NOT] EXISTS)`.

modules/compiler/builtins/cmake/FindBuiltinsTools.cmake
PietroGhg: I've moved the check at the bottom

modules/compiler/builtins/CMakeLists.txt
PietroGhg: Yeah you are right, I've removed the mention of `LLVM_NATIVE_TOOL_DIR`, thank you

modules/compiler/builtins/cmake/FindBuiltinsTools.cmake
```diff
@@ -96,6 +96,13 @@ function(find_builtins_tools tools_dir)
       IMPORTED_LOCATION ${BUILTINS_LINKER})
   endif()
 
+  if (CMAKE_CROSSCOMPILING AND OCK_IN_LLVM_TREE)
+    # If we are building OCK in LLVM TREE, ImportLLVM was not included and
+    # so LLVM_PACKAGE_VERSION is not defined.
+    set(LLVM_PACKAGE_VERSION
+    "${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}${LLVM_VERSION_SUFFIX}")
+  endif()
+
```
hvdijk, line 105: Ah, that shows what it's for... We could instead perhaps just use `${LLVM_VERSION_MAJOR}.${LLVM_VERSION_MINOR}.${LLVM_VERSION_PATCH}` (note: no `LLVM_VERSION_SUFFIX`) below instead of `LLVM_PACKAGE_VERSION`? We can probably drop the `string(REPLACE "svn"` as well.

PietroGhg: I've done it, thank you

coldav (approved):

modules/compiler/builtins/CMakeLists.txt
```diff
@@ -35,9 +35,13 @@ include(cmake/FindBuiltinsTools.cmake)
 if(NOT CA_EXTERNAL_BUILTINS AND CA_RUNTIME_COMPILER_ENABLED)
   if(NOT DEFINED CA_BUILTINS_TOOLS_DIR)
     if(CMAKE_CROSSCOMPILING)
-      message(FATAL_ERROR
-        "Cross compiled builds requires CA_EXTERNAL_BUILTINS or "
-        "CA_BUILTINS_TOOLS_DIR to be set!")
+      if(EXISTS ${LLVM_NATIVE_TOOL_DIR})
```
hvdijk, line 38: Just making sure, what does this do when `LLVM_NATIVE_TOOL_DIR` is not defined? Is `if(EXISTS )` well-defined, or do we need something like `if(DEFINED LLVM_NATIVE_TOOL_DIR AND EXISTS "${LLVM_NATIVE_TOOL_DIR}")` ?

PietroGhg: The [documentation](https://cmake.org/cmake/help/latest/command/if.html#exists) states that `if(EXISTS ` is false if it's given an empty string, and I think an undefined variable evaluates to an empty string, so this looks ok to me, but I can add the check if you think it's useful

hvdijk: That's fine then, thanks for checking.

hvdijk (approved):

PR #437, opened 2024-04-18
add support for a hal remote library

coldav:
# Overview

This adds support for generic client and server hal support to allow
sending commands as binary requests over a `hal_transmitter`.
    
# Reason for change

Required to support customer needs of talking to a remote cpu.

# Description of change

Added encoders and decoders for converting hal commands to binary format
    
Add hal_server base class which uses the encoders and decoders and a
hal_transmitter to receive hal commands remotely and act on the given
hal device.
    
Add a hal_client and hal_device_client base class which can be used with
a transmitter to send to a remote server binary commands.
    
Added a hal socket cpu client.
Added hal cpu socket based server program.

This currently only supports single device hals

### DIFF
--- a/examples/hal_cpu_remote_server/CMakeLists.txt
+++ b/examples/hal_cpu_remote_server/CMakeLists.txt
@@ -1,0 +1,43 @@
+# Copyright (C) Codeplay Software Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+# Exceptions; you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+cmake_minimum_required(VERSION 3.16 FATAL_ERROR)
+
+project(hal_cpu_remote_server VERSION 1.0.0 LANGUAGES C CXX ASM)
+
+set(HAL_CPU_REMOTE_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../clik/external/hal_cpu)
+
+add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../../hal ${CMAKE_CURRENT_BINARY_DIR}/hal)
+
+# Add the files in directly, rather than drag in as a shared library or take
+# all the extra clik code
+add_executable(hal_cpu_server_bin 
+  hal_cpu_executable.cpp
+  ${HAL_CPU_REMOTE_SOURCE_DIR}/source/cpu_hal.cpp
+  ${HAL_CPU_REMOTE_SOURCE_DIR}/source/hal_main.cpp
+)
+
+target_compile_definitions(hal_cpu_server_bin PUBLIC -DHAL_CPU_MODE=HAL_CPU_WG_MODE)
+target_include_directories(hal_cpu_server_bin PRIVATE
+    ${CMAKE_CURRENT_BINARY_DIR}
+    ${HAL_CPU_REMOTE_SOURCE_DIR}/include
+)
+
+target_link_libraries(hal_cpu_server_bin hal_remote hal_common dl pthread)
+
+
+install(TARGETS hal_cpu_server_bin
+        LIBRARY DESTINATION bin
+)
--- a/examples/hal_cpu_remote_server/README.md
+++ b/examples/hal_cpu_remote_server/README.md
@@ -1,0 +1,127 @@
+# Hal server
+
+## Introduction
+
+This supports a remote CPU based HAL which uses a socket connection to
+communicate with a remote client. This can be cross-compiled as needed. It is a
+small program which can be modified to use different ways of
+transmission or different underlying HAL interfaces.
+
+It should be paired with a built `oneapi-construction-kit` OpenCL interface using
+the HAL socket client, which resides under `examples/hal_cpu_client`. Details of
+how to build this target will be given below.
+
+## Security
+
+We recommend not running this program as root to reduce any likelihood of a
+malicious actor, as this gives a protocol for running programs. The customer
+should consider a specific user account just for running the server that has
+limited access or using dockers to limit access.
+
+We also recommend using ssh port forwarding. Note that we do not add any
+additional encryption of the executables that are sent.
+
+## Building the server
+
+This can be done as follows:
+
+```
+  cmake -GNinja <path_to_ock>/examples/hal_cpu_remote_server -Bbuild_server
+  ninja
+```
+
+If cross-compilation (for example RISC-V) is needed the following additional arguments are needed:
+
+```
+   -DCMAKE_TOOLCHAIN_FILE=<path_to_ock>/platform/riscv64-linux/riscv64-gcc-toolchain.cmake -DCMAKE_CROSSCOMPILING=TRUE
+```
+
+Note this requires the RISC-V toolchain to be installed as follows on Ubuntu:
+
+```
+  sudo apt-get install --yes gcc-9-riscv64-linux-gnu g++-9-riscv64-linux-gnu
+```
+
+If you wish to run on a non RISC-V machine, then `qemu` and the RISC-V toolchain will need to be installed.
+
+## Building the client
+
+We are looking to simplify this process, but currently we need to create an out
+of tree target. This is done as follows (for RISC-V):
+
+```
+  python <path_to_ock>/scripts/create_target.py <path_to_ock> \
+    <path_to_ock>/scripts/new_target_templates/cpu_client_riscv.json \
+    --external-dir <new_target_dir>
+
+  cd <new_target_dir>
+
+  cmake -Bbuild_client -GNinja \
+    -DCA_MUX_TARGETS_TO_ENABLE="cpu_client" \
+    -DCA_CPU_CLIENT_ENABLED=ON \
+    -DCA_ENABLE_API=cl \
+    -DCA_LLVM_INSTALL_DIR=<path_to_llvm_install> \
+    -DCA_CL_ENABLE_ICD_LOADER=ON  \
+    -DCA_EXTERNAL_ONEAPI_CON_KIT_DIR=<path_to_ock> 
+    -DCA_EXTERNAL_CPU_CLIENT_HAL_DIR=<path_to_ock>/examples/hals/hal_cpu_client
+   ninja -Cbuild_client
+```
+
+This assumes building the ICD, but it is possible to build without this.
+
+Note if we just wish a quick test then we could target `UnitCL` with the ninja
+line. Also if the server is running on a different architecture the LLVM build
+will need to enable that architecture.
+
+# Running the server
+
+The server requires a port which will be listened on. This can be any free user
+port. It also will only accept connections from specified nodes. This can be an
+ip address or host name e.g. "127.0.0.1" or "localhost". This defaults to
+"127.0.0.1", assuming that ports will be forwarded if the client is not run on
+the same machine.
+
+The server can be run as follows:
+
+```
+   cd build_server
+   ./hal_cpu_server_bin <port>
+```
+The node can also set with `-n`. It supports `-h` for info.
+
+The server will accept one connection, which when completed will end the
+program. This is to avoid programs putting the cpu into a bad state.
+
+The server can easily be run in a loop though or repeatedly started through a service manager e.g.
+
+```
+ while qemu-riscv64 -L/usr/riscv64-linux-gnu ./hal_cpu_server_bin <port>; do :; done
+```
+
+The environment variable `HAL_DEBUG_SERVER` can be set to 1 to give more debug output.
+
+We recommend this is never run as root as it runs executables sent over tcp/ip.
+
+# Running the client
+
+The client currently supports a socket connection only to the current node and
+it expects the server to be running on the same machine. However, if the
+customer desires to be able to run the server on a different machine (for
+example where we have a RISC-V device running Ubuntu), then the customer may
+consider using ssh port forwarding as suggested above. The recommended way of
+using port forwarding would be as follows:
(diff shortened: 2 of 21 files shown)


examples/hals/hal_cpu_client/source/hal_cpu_executable.cpp
```diff
@@ -0,0 +1,52 @@
...
+  if (const char *env = getenv("HAL_REMOTE_PORT")) {
+    port = atoi(env);
+  }
+
+  hal::hal_socket_transmitter transmitter(port);
+  hal::hal_server::error_code last_error = hal::hal_server::status_success;
+  do {
+    auto res = transmitter.start_server(true);
+    if (res != hal::hal_socket_transmitter::success) {
+      printf("Unable to start server on requested port %d\n", port);
```
hvdijk, line 35: This should probably go to `stderr`.

examples/hals/hal_cpu_client/source/hal_cpu_executable.cpp
```diff
@@ -0,0 +1,52 @@
...
+    // Check if the failure was due the transmitter, if so check if connection
+    // closed, otherwise we have a real error.
+    if (last_error == hal::hal_server::status_transmitter_failed &&
+        transmitter.get_last_error() !=
+            hal::hal_socket_transmitter::connection_closed) {
+      fprintf(stderr, "Error with tcp/ip connection\n");
+      exit(1);
+    }
+
+  } while (last_error == hal::hal_server::status_transmitter_failed);
```
hvdijk, line 51: Are we sure that after `status_transmitter_failed`, the code is in a state where the server can be meaningfully restarted? I would personally prefer to have a service manager restart the whole process automatically (several service managers exist) but I'm okay with this approach if we know the process is in a good enough state.

examples/hals/hal_cpu_client/source/hal_main.cpp
```diff
@@ -0,0 +1,61 @@
...
+    uint32_t port = 0;
+    if (const char *env = getenv("HAL_REMOTE_PORT")) {
+      port = atoi(env);
+    }
+    set_port(port);
+  }
+
+  // return generic platform information
+  const hal::hal_info_t &get_info() override {
+    std::lock_guard<std::mutex> locker(lock);
```
hvdijk, line 41: This lock doesn't do anything, does it?

hal/hal_remote/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+  ${CMAKE_CURRENT_SOURCE_DIR}/include/hal_remote/hal_binary_decoder.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/include/hal_remote/hal_client.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/include/hal_remote/hal_device_client.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/include/hal_remote/hal_server.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/include/hal_remote/hal_transmitter.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/source/hal_device_client.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/source/hal_server.cpp
+)
+
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
hvdijk, line 28: On non-Linux systems, what is the result? Do we get a `hal_remote` library with no object files?

hal/hal_remote/include/hal_remote/hal_binary_decoder.h
```diff
@@ -0,0 +1,388 @@
...
+      case hal_binary_encoder::COMMAND::MEM_WRITE_REPLY:
+      case hal_binary_encoder::COMMAND::MEM_FILL_REPLY:
+      case hal_binary_encoder::COMMAND::MEM_FREE_REPLY:
+      case hal_binary_encoder::COMMAND::MEM_COPY_REPLY:
+      case hal_binary_encoder::COMMAND::PROGRAM_FREE_REPLY:
+      case hal_binary_encoder::COMMAND::KERNEL_EXEC_REPLY:
+      case hal_binary_encoder::COMMAND::DEVICE_CREATE_REPLY:
+      case hal_binary_encoder::COMMAND::DEVICE_DELETE_REPLY:
+        return sizeof(uint32_t);  // bool as 32 bit
+        break;
```
hvdijk, line 95: This is unreachable.

hal/hal_remote/include/hal_remote/hal_binary_decoder.h
```diff
@@ -0,0 +1,388 @@
...
+      default:
+        printf("Unknown command : %d\n", (int)command);
+        assert(0 && "Command cannot be decoded");
```
hvdijk, line 98: Since this is data from an untrusted client, we probably can't do `assert(0)`, we probably have to take that out and always do the `return`?

hal/hal_remote/include/hal_remote/hal_client.h
```diff
@@ -0,0 +1,115 @@
...
+      made_connection = true;
+    }
+
+    hal::hal_binary_encoder encoder(index);
+    hal::hal_binary_decoder decoder;
+    hal::hal_binary_encoder::COMMAND command;
+    encoder.encode_device_create();
+    get_transmitter().send(encoder.data(), encoder.size(), true);
+    bool ok = get_transmitter().receive(
+        (uint8_t *)&command, sizeof(hal::hal_binary_encoder::COMMAND));
```
hvdijk, line 65: This is wrong if server and client disagree on endianness. We only have little endian targets at the moment, can we add an assert for that?

hal/hal_remote/include/hal_remote/hal_binary_decoder.h
```diff
@@ -0,0 +1,388 @@
...
+    return data_required_unknown;
+  }
+
+  /// @brief  decode a command given data of a particular sise
+  /// @param command
+  /// @param data
+  /// @param size length of data in bytes
+  /// @return true if succeeded otherwise false.
+  bool decode(hal_binary_encoder::COMMAND command, const uint8_t *data,
+              uint32_t size) {
```
hvdijk, line 109: Can we use `cargo::result` here to make it clearer from the API what the meaning of the result is?

hvdijk: Also, I cannot really follow the logic of when `uint8_t *` is used vs. `void *`, I would have expected `void *` here but it's not a problem if there is a reason for it.

examples/hals/hal_cpu_client/source/hal_cpu_executable.cpp
coldav: it's a bit convoluted but I check the connection_closed above so if it fails at this point it's because the connection is closed. However I think we agreed before that this could lead to all sorts of issues with allocs being left over etc, so I'll take out the loop and advise the customer to use a bash while loop as an option.

hal/hal_remote/CMakeLists.txt
coldav: when I had it initially as all headers, it didn't seem to complain, but it does have some .cpp files above anyway (L24).

hal/hal_remote/include/hal_remote/hal_binary_decoder.h
coldav: I don't want to drag in cargo, we've avoided it in hal so far in keeping it mostly standalone (and working with clik).
I think you are right about void * though, although we do tend to need to then cast it elsewhere to index into it.

hal/hal_remote/include/hal_remote/hal_client.h
coldav: I agree it depends on matching endianness. I'll have to have a think about best to check the endianness with an assert (in theory it's in the info).

examples/hals/hal_cpu_client/source/hal_main.cpp
coldav: nothing useful no. I will remove.

examples/hal_cpu_remote_server/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+cmake_minimum_required(VERSION 3.16 FATAL_ERROR)
+
+project(hal_cpu_remote_server VERSION 1.0.0 LANGUAGES C CXX ASM)
+
+set(HAL_CPU_REMOTE_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/../../clik/external/hal_cpu)
```
hvdijk, line 20: Can we move `examples/clik/external/hal_cpu` to `examples/external/hal_cpu` since it is no longer limited to `clik`?

examples/hal_cpu_remote_server/README.md
```diff
@@ -0,0 +1,127 @@
...
+```
+  sudo apt-get install --yes gcc-9-riscv64-linux-gnu g++-9-riscv64-linux-gnu
+```
+
+If you wish to run on a non RISC-V machine, then `qemu` and the RISC-V toolchain will need to be installed.
+
+## Building the client
+
+We are looking to simplify this process, but currently we need to create an out
+of tree target. This is done as follows (for RISC-V):
```
hvdijk, line 50: I could be misremembering, but didn't you want to change this process before merging?

examples/hal_cpu_remote_server/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+The server can easily be run in a loop though or repeatedly started through a service manager e.g.
+
+```
+ while qemu-riscv64 ./hal_cpu_server_bin; do :; done
+```
+
+The environment variable `HAL_DEBUG_SERVER` can be set to 1 to give more debug output.
+
+We recommend this is never run as root as it runs executables sent over tcp/ip.
```
hvdijk, line 103: We talked about mentioning the option of sandboxing in some way too here.

examples/hal_cpu_remote_server/hal_cpu_executable.cpp
```diff
@@ -0,0 +1,118 @@
...
+  std::string node = "127.0.0.1";
+  // Note this default will fail as we do not allow port 0
+  uint16_t port = 0;
+  struct sigaction action;
+  memset(&action, 0, sizeof(action));
+  action.sa_handler = handle_sig;
+  sigaction(SIGTERM, &action, NULL);
+
+  for (;;) {
+    switch (getopt(argc, argv, "n:h?")) {
```
hvdijk, line 56: `?` is not a good option for `getopt` because the return value of `'?'` for an option `-?` is indistinguishable from the return value of `'?'` for an unrecognised option, and the option `-?` is not something we should try to support since `?` is a shell metacharacter and so unless quoted, this looks for files in the current directory whose names are `-` followed by a single character.

examples/hal_cpu_remote_server/README.md
coldav: do you have any suggestions e.g. You may also consider running in a docker to sandbox it?

examples/hal_cpu_remote_server/CMakeLists.txt
coldav: That's a bigger change since the original idea of clik was that it was fairly standalone. I think it's a good idea, but I'd prefer to to do it in a separate PR.

examples/hal_cpu_remote_server/README.md
coldav: I think it's a big enough change, we should consider doing it separately and not hold up this PR.

hal/include/hal_types.h
```diff
@@ -169,7 +169,7 @@ struct hal_device_info_t {
   uint32_t max_workgroup_size;
 
   /// @brief true if little endian.
-  bool is_little_endian;
+  bool is_little_endian = true;
```
hvdijk, line 172: We should be detecting this rather than assuming, I think.

examples/hal_cpu_remote_server/README.md
hvdijk: Could do that, or even just running under a user account that does not have access to anything.

coldav: I already say that above, I've also added dockers as a mention.

hal/include/hal_types.h
coldav: This is a default in the base Hal, it's up to the writer of each HAL to dictate this value (also because clc uses this an offline way, we can't easily detect other than the host) - I have added some additional checks on the host side though to see that it has the same reported endianness.

hvdijk: That makes sense, being unable to detect it. But it still shouldn't be necessary to set it here. I see it's already getting set as it should be in `cpu_hal::setup_cpu_hal_device_info`.

coldav: This is more about avoiding uninitialized variables than it being the defacto value.

hvdijk: But it gets initialized to `true` rather than `{}`, and other fields are not initialized similarly.

coldav: fair enough, I'll remove it, I've just seen issues with not setting things to default values here, especially when we add fields, but I'll think on it and remove it from this change.

coldav: removed.

examples/hal_cpu_remote_server/hal_cpu_executable.cpp
```diff
@@ -0,0 +1,118 @@
...
+      case 'n':
+        node = optarg;
+        continue;
+
+      case '?':
+      case 'h':
+      default:
+        print_usage(argv[0]);
+        exit(0);
```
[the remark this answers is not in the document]
hvdijk: This does print to `stderr` on bad options now, but still exits with exit code 0.

examples/hal_cpu_remote_server/hal_cpu_executable.cpp
```diff
@@ -0,0 +1,120 @@
...
+  for (;;) {
+    switch (auto opt = getopt(argc, argv, "n:h")) {
+      case 'n':
+        node = optarg;
+        continue;
+
+      case 'h':
+        print_usage(stdout, argv[0]);
+        exit(0);
+        break;
```
hvdijk, line 64: Nit: code after `exit(0)` isn't going to execute :)

hvdijk (approved):

PR #446, opened 2024-05-01
Remove outdated partial Unified Runtime adapter

kbenzie:
The recommended integration into oneAPI through the construction kit is via the OpenCL adapter. This patch removes the outdated, partial, implementation of the Unified Runtime specification on top of ComputeMux.


coldav (approved):

hvdijk:
No issue with this, but note that we need to make sure we remove this from our CI testing before we remove the code.

kbenzie:
> No issue with this, but note that we need to make sure we remove this from our CI testing before we remove the code.

Is that in GitHub Actions?

coldav:
> > No issue with this, but note that we need to make sure we remove this from our CI testing before we remove the code.
> 
> Is that in GitHub Actions?

sadly not but I will merge this once it's removed there if that is okay.

kbenzie:
> sadly not but I will merge this once it's removed there if that is okay.

No worries, was just wondering if I need to do something in this PR.

PR #449, opened 2024-05-06
LLVM 19 fixes

PietroGhg:
# Overview

Both `muxc` and `veczc` were not using the LLVM passes to out assembly/bitcode, this PR addressed that. 
This PR also makes sure that a module is created with the old debug info format in `spirv-ll`.

# Reason for change

Not using the upstream passes means that we were missing out on any kind of logic that is added upstream when outputting IR, e.g. LLVM 19 is changing the debug info format, and `PrintModulePass` handles printing IR in a way that doesn't break `CHECK` directives for debug intrinsics.

# Description of change

Uses `PrintModulePass` and `BitcodeWriterPass` in `muxc` and `veczc`, sets old type of debug info format in `spirv-ll`.

# Checklist

### DIFF
--- a/modules/compiler/spirv-ll/source/module.cpp
+++ b/modules/compiler/spirv-ll/source/module.cpp
@@ -73,7 +73,11 @@ spirv_ll::Module::Module(
       WorkgroupSize({{1, 1, 1}}),
       BufferSizeArray(nullptr),
       deferredSpecConstantOps(),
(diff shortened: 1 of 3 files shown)


hvdijk (approved):
LGTM, and personally, since the remaining problems are not OCK problems, but LLVM problems, will be fixed there, and this PR doesn't make things any worse, I'd be okay with merging this already.

PR #450, opened 2024-05-06
[host] Remove invalid caching of object code.

hvdijk:
# Overview

[host] Remove invalid caching of object code.

# Reason for change

As seen in clBuildProgramTwiceTest.RedefineMacro, we may reuse a HostModule to create a binary from a different LLVM module, or with different LLVM build options. We do not have sufficient information to determine whether we can re-use a prior result, and can reasonably assume that if createBinary is called a second time, it will be called a second time specifically because the caller does not want the prior result.

# Description of change

Remove the loading of the prior value of `object_code`.

# Checklist

### DIFF
--- a/modules/compiler/targets/host/source/module.cpp
+++ b/modules/compiler/targets/host/source/module.cpp
@@ -177,11 +177,6 @@ compiler::Result HostModule::createBinary(
     return compiler::Result::FINALIZE_PROGRAM_FAILURE;
   }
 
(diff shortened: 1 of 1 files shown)


coldav (approved):

hvdijk:
> Can we check the benchmarking on this before merge please?

Yep, benchmarks showed no impact on how we normally run.

PR #452, opened 2024-05-08
Switch to llvm versions 17/18 as supported versions

coldav:
# Overview

Switch min/max versions for OCK to be 17/18

# Reason for change

Require to support only two LLVM versions, and 18 is stable

# Description of change

Updated cmake checks and documentation
updated ci usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ Upgrade guidance:
 * The `compiler::utils::createLoop` API has moved its list of `IVs` parameter
   into its `compiler::utils::CreateLoopOpts` parameter. It can now also set the
   IV names via a second `CreateLoopOpts` field.
-* Support for LLVM versions is now limited to LLVM 16 and LLVM 17. Support for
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -39,7 +39,7 @@ Upgrade guidance:
   `CA_HOST_TARGET_AARCH64_CPU` and `CA_HOST_TARGET_RISCV64_CPU`. The environment variable
   `CA_HOST_TARGET_CPU` remains the same name. Note that `CA_HOST_TARGET_CPU_NATIVE` is no longer
   supported but can be achieved by using `native` as a value for the variants.
-
+* LLVM 16 is no longer supported. 17 and 18 are the only supported versions.
```
hvdijk, line 42: There is an earlier entry "Support for LLVM versions is now limited to LLVM 16 and LLVM 17. Support for earlier LLVM versions has been removed." This should probably be an edit of that instead, rather than a new entry.

coldav: good spot

hvdijk (approved):

PR #453, opened 2024-05-08
Fix replace_local_module_scope_variables mempcy bug

coldav:
# Overview

Rewrite replace_local_module_scope_variables to only add parameters where needed.

# Reason for change

replace_local_module_scope_variables was incorrectly trying to clone and delete the original of memcpy which was referenced in a llvm.compiler.used line to ensure it continued to exist.

# Description of change

This fixes this by looking at any globals and working back through the user chain until we find an Instruction and only cloning functions that actually require it, as well as all kernels. It also moves and modifies the `addParamToAllFunctions` to the ReplaceLocalModuleScopeVariablesPass as it is only used in there and we can make more appropriate changes.

This change puts it more in line with other similar add parameter pass_functions such as the scheduling parameters and is more efficient.

### DIFF
--- a/modules/compiler/compiler_pipeline/source/replace_local_module_scope_variables_pass.cpp
+++ b/modules/compiler/compiler_pipeline/source/replace_local_module_scope_variables_pass.cpp
@@ -19,6 +19,8 @@
 #include <compiler/utils/metadata.h>
 #include <compiler/utils/pass_functions.h>
 #include <compiler/utils/replace_local_module_scope_variables_pass.h>
+#include <llvm/ADT/PriorityWorklist.h>
+#include <llvm/ADT/SmallPtrSet.h>
 #include <llvm/IR/Constants.h>
 #include <llvm/IR/DIBuilder.h>
 #include <llvm/IR/IRBuilder.h>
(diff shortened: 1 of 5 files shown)


modules/compiler/compiler_pipeline/source/replace_local_module_scope_variables_pass.cpp
```diff
@@ -117,6 +121,106 @@ struct GlobalVarDebugInfoWrapper final {
...
+  // currently takes a std::function so we search the created vector of
+  // functions.
+  return compiler::utils::cloneFunctionsAddArg(
+      module,
+      [newParamType, newParamAttrs](llvm::Module &) {
+        return compiler::utils::ParamTypeAttrsPair{newParamType, newParamAttrs};
+      },
+      [&FuncsToClone](const llvm::Function &func, bool &ClonedWithBody,
+                      bool &ClonedNoBody) {
+        ClonedWithBody = FuncsToClone.end() != find(FuncsToClone, &func);
```
hvdijk, line 218: This is `llvm::find` being used here, right? `llvm::is_contained` might be a better fit.

coldav: sounds good, done.

PR #454, opened 2024-05-09
Remove all code specific to LLVM 16

coldav:
# Overview

We have now switched to a minimum of LLVM 17, so there is no need for specific code to handle LLVM 16.


modules/compiler/multi_llvm/include/multi_llvm/triple.h
```diff
@@ -16,12 +16,7 @@
...
 #define MULTI_LLVM_TRIPLE_H_INCLUDED
 
-#include <multi_llvm/llvm_version.h>
-
-#if LLVM_VERSION_MAJOR >= 17
 #include <llvm/TargetParser/Triple.h>
-#else
-#include <llvm/ADT/Triple.h>
-#endif
+#include <multi_llvm/llvm_version.h>
```
hvdijk, line 20: This whole header was only so that users didn't need the `#if` for how to include `Triple.h`. Now that this is no longer needed, can we update users of `<multi_llvm/triple.h>` to include `<llvm/TargetParser/Triple.h>` directly?

coldav: agreed. Done.

hvdijk (approved):

PR #455, opened 2024-05-14
Reduce memory consumption.

hvdijk:
# Overview

Reduce memory consumption.

# Reason for change

On 32-bit platforms, OpenCL CTS's `test_conversions` test would fail after running out of memory.

# Description of change

* When destroying a HostKernel, clean up the string pool. The string pool contains the names of every symbol used by the kernel now freed. We do not want these to persist if there is nothing else still using them.
* When resetting a command buffer, do not just clear the commands, shrink the commands too. Otherwise, if enough commands were present that dynamic allocations were needed, this memory would not be freed until the command buffer got destroyed. But because we re-use command buffers, this would keep the allocations until long after we no longer need them.
* When resetting a command buffer, do not just clear and shrink the commands, also clear and shrink the ndranges. These too are no longer needed when a command buffer is reused.

# Checklist

### DIFF
--- a/modules/mux/targets/host/source/command_buffer.cpp
+++ b/modules/mux/targets/host/source/command_buffer.cpp
@@ -1189,6 +1189,9 @@ mux_result_t hostResetCommandBuffer(mux_command_buffer_t command_buffer) {
   const std::lock_guard<std::mutex> lock(host->mutex);
 
   host->commands.clear();
+  host->commands.shrink_to_fit();
+  host->ndranges.clear();
+  host->ndranges.shrink_to_fit();
 
   return mux_success;
 }
--- a/modules/compiler/targets/host/CMakeLists.txt
+++ b/modules/compiler/targets/host/CMakeLists.txt
@@ -64,7 +64,8 @@ set(HOST_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/source/target.cpp
     )
 
-add_ca_library(compiler-host STATIC ${HOST_SOURCES})
+add_ca_library(compiler-host STATIC ${HOST_SOURCES}
+  DEPENDS mux-config abacus_generate)
 
 target_include_directories(compiler-host PUBLIC
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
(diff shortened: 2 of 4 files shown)


modules/mux/targets/host/source/command_buffer.cpp
```diff
@@ -1189,6 +1189,9 @@ mux_result_t hostResetCommandBuffer(mux_command_buffer_t command_buffer) {
   const std::lock_guard<std::mutex> lock(host->mutex);
 
   host->commands.clear();
+  host->commands.shrink_to_fit();
+  host->ndranges.clear();
+  host->ndranges.shrink_to_fit();
```
coldav, line 1194: Why does it need this clear/shrink/clear/shrink? Maybe add a comment as to why?

coldav (approved):

modules/mux/targets/host/source/command_buffer.cpp
hvdijk: `clear()` just sets the vector's size to zero while leaving its capacity as it was, `shrink_to_fit()` tries to reduce the vector's capacity (the number of elements the vector has reserved memory for) too. Although `host->commands` and `host->ndranges` are defined as cargo's `small_vector`, the API is the same as for `std::vector` ([`clear()`](https://en.cppreference.com/w/cpp/container/vector/clear), [`shrink_to_fit()`](https://en.cppreference.com/w/cpp/container/vector/shrink_to_fit)), so this should hopefully be familiar to people already.

I did expect this would be commented in the Doxygen comments for [`small_vector::clear()`](https://github.com/codeplaysoftware/oneapi-construction-kit/blob/bed85cff4d1f5133a814982e948e22d8dd558c9f/modules/cargo/include/cargo/small_vector.h#L471) and [`small_vector::shrink_to_fit()`](https://github.com/codeplaysoftware/oneapi-construction-kit/blob/bed85cff4d1f5133a814982e948e22d8dd558c9f/modules/cargo/include/cargo/small_vector.h#L459), but it isn't. I'm happy to add a bit more there.

coldav: It wasn't the api usage as such, but why you do it twice.

coldav: sorry I misread it the first time, it's fine.

PR #457, opened 2024-05-15
Fix memory leak.

hvdijk:
# Overview

Fix memory leak.

# Reason for change

In bf1b1eab5b, I cleared host->ndranges, but the result of that the loop in hostDestroyCommandBuffer to free the arguments no longer does anything useful: there is no longer any reference to the arguments.

# Description of change

Change packed_args to a mux::dynamic_array so that this is automatically taken care of for us.

# Checklist

### DIFF
--- a/modules/mux/targets/host/source/command_buffer.cpp
+++ b/modules/mux/targets/host/source/command_buffer.cpp
@@ -67,8 +67,9 @@ size_t calcPackedArgsAllocSize(
 // Iterates through the argument descriptors and for each argument sets the
 // location in the packed args location to the correct value.
 void populatePackedArgs(
-    uint8_t *packed_args_alloc,
+    mux::dynamic_array<uint8_t> &packed_args,
     const mux::dynamic_array<mux_descriptor_info_t> &descriptors) {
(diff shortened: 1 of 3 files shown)


modules/mux/targets/host/source/command_buffer.cpp
```diff
@@ -67,8 +67,9 @@ size_t calcPackedArgsAllocSize(
 // Iterates through the argument descriptors and for each argument sets the
 // location in the packed args location to the correct value.
 void populatePackedArgs(
-    uint8_t *packed_args_alloc,
+    mux::dynamic_array<uint8_t> &packed_args,
     const mux::dynamic_array<mux_descriptor_info_t> &descriptors) {
+  uint8_t *const packed_args_alloc = packed_args.begin();
```
coldav, line 72: maybe data() is clearer (appreciate it's the same thing here).

coldav (approved):

modules/mux/targets/host/source/command_buffer.cpp
hvdijk: Agreed, and likewise elsewhere, will update.

PR #461, opened 2024-05-20
Fix hal_remote to build for windows, fix other minor issues

coldav:
# Overview

This fixes the build for windows so that it does not discard building hal_remote library.
There is also an update for the README to resolve the incorrect description on the ssh forwarding.

# Reason for change

Run through showed some minor issues, we also want to have windows capable of building most of the code.

### DIFF
--- a/hal/hal_remote/CMakeLists.txt
+++ b/hal/hal_remote/CMakeLists.txt
@@ -25,7 +25,7 @@ set(HAL_SOURCE
   ${CMAKE_CURRENT_SOURCE_DIR}/source/hal_server.cpp
 )
 
-if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
   list (APPEND HAL_SOURCE ${CMAKE_CURRENT_SOURCE_DIR}/include/hal_remote/hal_socket_client.h)
   list (APPEND HAL_SOURCE ${CMAKE_CURRENT_SOURCE_DIR}/include/hal_remote/hal_socket_transmitter.h)
   list (APPEND HAL_SOURCE ${CMAKE_CURRENT_SOURCE_DIR}/source/hal_socket_transmitter.cpp)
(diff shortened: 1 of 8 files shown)


hal/hal_remote/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ set(HAL_SOURCE
   ${CMAKE_CURRENT_SOURCE_DIR}/source/hal_server.cpp
 )
 
-if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+if(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "Darwin")
```
hvdijk, line 28: I'd rather we didn't add macOS-specific logic until we're ready to actually support that, but you can achieve essentially the same effect with
```suggestion
if(NOT CMAKE_SYSTEM_NAME STREQUAL "Windows")
```
which I'd be fine with, if that works.

hal/hal_remote/include/hal_remote/hal_binary_decoder.h
```diff
@@ -93,7 +93,7 @@ class hal_binary_decoder {
       case hal_binary_encoder::COMMAND::DEVICE_DELETE_REPLY:
         return sizeof(uint32_t);  // bool as 32 bit
       default:
-        fprintf(stderr, "Unknown command : %d\n", (int)command);
+        (void)fprintf(stderr, "Unknown command : %d\n", (int)command);
```
hvdijk, line 96: Can we either use `std::cerr` everywhere, or use `fprintf` everywhere? There are some annoying subtleties when you mix the two. I *think* that won't affect us, but I'm not 100% sure and if we can avoid having to think about that, that will make things easier.

coldav: yes fair enough, will change

hal/hal_remote/CMakeLists.txt
coldav: we already have Darwin stuff in OCK - but I'll got with not "windows"

hvdijk (approved):

PR #463, opened 2024-05-21
[host] Fix hostQueryLocalSizeForSubGroupCount

hvdijk:
# Overview

[host] Fix hostQueryLocalSizeForSubGroupCount

# Reason for change

When we fixed HostKernel::queryLocalSizeForSubGroupCount last year, we missed that the JIT and non-JIT paths go through different code paths, and the same fix is also needed for hostQueryLocalSizeForSubGroupCount.

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/modules/mux/targets/host/source/kernel.cpp
+++ b/modules/mux/targets/host/source/kernel.cpp
@@ -259,13 +259,28 @@ mux_result_t hostQueryLocalSizeForSubGroupCount(mux_kernel_t kernel,
                                                 size_t *local_size_x,
                                                 size_t *local_size_y,
                                                 size_t *local_size_z) {
-  (void)kernel;
-  // FIXME: For a single sub-group, we know we can satisfy that with a
-  // work-group of 1,1,1. For any other sub-group count, we should ensure that
-  // the work-group size we report comes back through getKernelVariantForWGSize
-  // when it comes to run it. See CA-4784.
-  if (sub_group_count == 1) {
-    *local_size_x = 1;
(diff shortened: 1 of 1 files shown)


modules/mux/targets/host/source/kernel.cpp
```diff
@@ -259,13 +259,29 @@ mux_result_t hostQueryLocalSizeForSubGroupCount(mux_kernel_t kernel,
...
                                                 size_t *local_size_y,
                                                 size_t *local_size_z) {
-  (void)kernel;
-  // FIXME: For a single sub-group, we know we can satisfy that with a
-  // work-group of 1,1,1. For any other sub-group count, we should ensure that
-  // the work-group size we report comes back through getKernelVariantForWGSize
-  // when it comes to run it. See CA-4784.
-  if (sub_group_count == 1) {
-    *local_size_x = 1;
+  host::kernel_variant_s variant;
```
coldav, line 262: Is there any possibility of using some common code with the jit path here?

hvdijk: In theory yes, in practice I think it would require a much bigger redesign. Right now, we don't even agree on the API: this function returns through pointers, whereas [`HostKernel::queryLocalSizeForSubGroupCount`](https://github.com/codeplaysoftware/oneapi-construction-kit/blob/cca28347df75b15572bf071000ceb1431621a86c/modules/compiler/targets/host/source/kernel.cpp#L182-L216) returns by value.

PR #469, opened 2024-05-29
Return a hal_device_info_riscv_t from cpu hal/remote hal

coldav:
# Overview

Return a hal_device_info_riscv_t from cpu hal/remote hal

# Reason for change

Allow in-tree risc-v to support cpu hal/remote HAL

# Description of change

Change cpu hal to return a reference to the device info.
Fix assert when vlen = 0 in mux kernels.
Add link_shared to hal arguments.
Add the examples/hals directory to the list searched

### DIFF
--- a/modules/mux/targets/riscv/external/CMakeLists.txt
+++ b/modules/mux/targets/riscv/external/CMakeLists.txt
@@ -22,7 +22,8 @@ set(RISCV_TOOLCHAIN_DIR "${CA_LLVM_INSTALL_DIR}" CACHE PATH
   "Path to the clang RISC-V toolchain")
 
 # Iterate through all 'hal_*' sub-directories.
-file(GLOB HAL_DIR_NAMES FOLLOW_SYMLINKS LIST_DIRECTORIES true "hal_*")
+file(GLOB HAL_DIR_NAMES FOLLOW_SYMLINKS LIST_DIRECTORIES true "hal_*"
+     ${PROJECT_SOURCE_DIR}/examples/hals/hal_*)
 
 if (DEFINED CA_RISCV_EXTERNAL_HAL_DIR)
   list(APPEND HAL_DIR_NAMES ${CA_RISCV_EXTERNAL_HAL_DIR})
--- a/modules/mux/targets/riscv/source/kernel.cpp
+++ b/modules/mux/targets/riscv/source/kernel.cpp
@@ -35,11 +35,13 @@ cargo::expected<riscv::kernel_s *, mux_result_t> kernel_s::create(
     cargo::string_view name, mux::allocator allocator) {
   auto *hal_device_info = static_cast<const riscv::hal_device_info_riscv_t *>(
       device->hal_device->get_info());
-  // Note if VLEN is unknown this will return 0
-  // Assert non-zero as sub_group_size is unknowable without it
-  assert(hal_device_info->vlen &&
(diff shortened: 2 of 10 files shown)


modules/mux/targets/riscv/external/CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ set(RISCV_TOOLCHAIN_DIR "${CA_LLVM_INSTALL_DIR}" CACHE PATH
 
 # Iterate through all 'hal_*' sub-directories.
 file(GLOB HAL_DIR_NAMES FOLLOW_SYMLINKS LIST_DIRECTORIES true "hal_*")
+file(GLOB HAL_DIR_NAMES_EXAMPLES FOLLOW_SYMLINKS LIST_DIRECTORIES true ${PROJECT_SOURCE_DIR}/examples/hals/hal_*)
```
hvdijk, line 26: We talked about possibly doing
```suggestion
file(GLOB HAL_DIR_NAMES_EXAMPLES FOLLOW_SYMLINKS LIST_DIRECTORIES true hal_* ${PROJECT_SOURCE_DIR}/examples/hals/hal_*)
```

modules/mux/targets/riscv/source/kernel.cpp
```diff
@@ -50,10 +47,19 @@ cargo::expected<riscv::kernel_s *, mux_result_t> kernel_s::create(
...
     variant.sub_group_size = meta.sub_group_size.getKnownMinValue();
     if (meta.sub_group_size.isScalable()) {
+      // Assert non-zero as sub_group_size is unknowable without vlen != 0
+      assert(
+          hal_device_info->vlen &&
+          "vlen must be known at runtime to calculate scalable subgroup width");
       variant.sub_group_size *= real_vscale;
     }
     variant.min_work_width = meta.min_work_item_factor.getKnownMinValue();
+    variant.min_work_width = meta.min_work_item_factor.getKnownMinValue();
```
hvdijk, line 57: Accidental duplication?

modules/mux/targets/riscv/source/kernel.cpp
```diff
@@ -50,10 +47,19 @@ cargo::expected<riscv::kernel_s *, mux_result_t> kernel_s::create(
...
     if (meta.min_work_item_factor.isScalable()) {
+      // Assert non-zero as sub_group_size is unknowable without vlen != 0
+      assert(
+          hal_device_info->vlen &&
+          "vlen must be known at runtime to calculate scalable subgroup width");
```
hvdijk, line 62: There is another use of `real_vscale` below that also needs the same assert. (Is there perhaps a way we can write that assert only once? Maybe make `real_scale` a lambda that includes the assert?)

modules/mux/targets/riscv/external/CMakeLists.txt
coldav: good idea

modules/mux/targets/riscv/source/kernel.cpp
coldav: ta

modules/mux/targets/riscv/source/kernel.cpp
coldav: i've tried something similar

hvdijk (approved):

PR #481, opened 2024-06-25
Fix risc-v memmove creating memcpy instrinsic

coldav:
# Overview

Fix risc-v memmove creating memcpy instrinsic

# Reason for change

SYCL CTS test vector_load_store with host risc-v showed that memmove intrinsics could result in a linked builtin which then included memcpy instrinsics which could result in a linker error.

# Description of change

This resolves this by passing -fno-builtin in the build for the builtins. This only removes the additional llvm.memcpy and all other instrinsics remain.

### DIFF
--- a/modules/compiler/targets/riscv/test/lit/passes/link-builtins-memcpy.ll
+++ b/modules/compiler/targets/riscv/test/lit/passes/link-builtins-memcpy.ll
@@ -1,0 +1,51 @@
+; Copyright (C) Codeplay Software Limited
+;
+; Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+; Exceptions; you may not use this file except in compliance with the License.
+; You may obtain a copy of the License at
+;
+;     https://github.com/codeplaysoftware/oneapi-construction-kit/blob/main/LICENSE.txt
+;
+; Unless required by applicable law or agreed to in writing, software
+; distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+; WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+; License for the specific language governing permissions and limitations
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; RUN: muxc --device "%riscv_device" --passes link-builtins,verify -S %s | FileCheck %s
+
+target triple = "riscv64-unknown-unknown-elf"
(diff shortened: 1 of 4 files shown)


modules/compiler/targets/riscv/test/lit/passes/link-builtins-memmove.ll
```diff
@@ -0,0 +1,52 @@
...
+; under the License.
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; RUN: muxc --device "%riscv_device" --passes link-builtins,verify -S %s | FileCheck %s
+
+target triple = "riscv64-unknown-unknown-elf"
+target datalayout = "e-p:64:64:64-m:e-i64:64-f80:128-n8:16:32:64-S128"
+
+; CHECK-NOT: call void @llvm.memcpy
```
hvdijk, line 22: Ideally, we'd check there that *either* no call to `memcpy` is emitted, *or* a definition of `memcpy` is emitted. But that isn't something we can easily do in FileCheck, so this test is fine as is.

modules/compiler/targets/riscv/test/lit/passes/link-builtins-memcpy.ll
```diff
@@ -0,0 +1,52 @@
...
+;
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+; RUN: muxc --device "%riscv_device" --passes link-builtins,verify -S %s | FileCheck %s
+
+target triple = "riscv64-unknown-unknown-elf"
+target datalayout = "e-m:e-p:64:64-i64:64-i128:128-n32:64-S128"
+
+; CHECK-NOT: call void @llvm.memcpy
+; CHECK: define linkonce dso_local spir_func noundef ptr @memmove(
```
hvdijk, line 23: This looks like a wrong test, I think you wanted to check in this file that `memcpy` works, not that `memmove` works?

coldav: ah yes quite right

hvdijk (approved):

modules/compiler/targets/riscv/test/lit/passes/link-builtins-memmove.ll
coldav: agreed, nature of lit tests.

PR #485, opened 2024-06-27
Fix risc-v memmove creating memcpy instrinsic

coldav:
# Overview

Pass the attribute __attribute__((no_builtin)) to the memmove builtin and adds lit tests for memcpy,memmove and memset intrinsics.

# Reason for change

SYCL CTS test vector_load_store with host risc-v showed that memmove intrinsics could result in a linked builtin which then included memcpy instrinsics which could result in a linker error.

### DIFF
--- a/modules/compiler/builtins/source/builtins.cpp
+++ b/modules/compiler/builtins/source/builtins.cpp
@@ -658,12 +658,17 @@ void *memcpy(void *__restrict dst, const void *__restrict src, size_t num) {
   return dst;
 }
 
+// We don't want an llvm.memcpy intrinsic here as that complicates the handling
+// in the link-builtins pass
 void *memmove(void *dst, const void *src, size_t num) {
   if (reinterpret_cast<uintptr_t>(static_cast<char *>(dst) + num) <=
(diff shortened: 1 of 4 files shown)


modules/compiler/builtins/source/builtins.cpp
```diff
@@ -658,12 +658,17 @@ void *memcpy(void *__restrict dst, const void *__restrict src, size_t num) {
...
 void *memmove(void *dst, const void *src, size_t num) {
   if (reinterpret_cast<uintptr_t>(static_cast<char *>(dst) + num) <=
           reinterpret_cast<uintptr_t>(src) ||
       reinterpret_cast<uintptr_t>(static_cast<const char *>(src) + num) <=
           reinterpret_cast<uintptr_t>(dst)) {
-    return memcpy(dst, src, num);
+    return [&]() __attribute__((always_inline, no_builtin)) {
+      return memcpy(dst, src, num);
+    }
+    ();
```
hvdijk, line 671: Interesting that clang-format gets confused by this. Had not expected that, but that's fine.

hvdijk (approved):

PR #486, opened 2024-06-27
Add missing atomic functions.

hvdijk:
# Overview

Add missing atomic functions.

# Reason for change

We only implemented the *_explicit versions, and only the versions that take both memory_order and memory_scope parameters. The OpenCL specification tells us we also need overloads that only take memory_order, and non-explicit versions that do not take either.

# Description of change

points, etc.*

Tested by running OpenCL CTS's updated `test_generic_address_space` test, which previously failed, and now passes.

# Checklist

### DIFF
--- a/modules/compiler/builtins/scripts/generate_header_30.sh
+++ b/modules/compiler/builtins/scripts/generate_header_30.sh
@@ -391,6 +391,8 @@ function all_atomic()
     for i in int long uint ulong float double
     do
       double_support_begin $i
+      echo "void __CL_BUILTIN_ATTRIBUTES atomic_store(volatile $k atomic_$i *object, $i desired);"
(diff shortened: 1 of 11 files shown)


modules/compiler/builtins/scripts/generate_header_30.sh
```diff
@@ -391,6 +391,8 @@ function all_atomic()
     for i in int long uint ulong float double
     do
       double_support_begin $i
+      echo "void __CL_BUILTIN_ATTRIBUTES atomic_store(volatile $k atomic_$i *object, $i desired);"
```
coldav, line 394: I believe we test some of the others, is it possible to add a test for these ones just so it's at least covered?

hvdijk: Done, I changed the existing `atomic_store_explicit` tests to use any of the three versions so that they all get used.

coldav (approved):

PR #489, opened 2024-07-04
Fix test IncrementKernelTwiceDifferentQueues

coldav:
# Overview

Replaced the global value increment with an atomic_inc in UnitCL test IncrementKernelTwiceDifferentQueues

# Reason for change

This test started two queues at the same time and updated the same value. This could on occasion cause a race condition, exhibited on risc-v host.


hvdijk (approved):
`atomic_inc` is one of the legacy atomic functions, just making sure, but that is still required to be fully supported, right? If so, this looks good to me.

coldav:
> `atomic_inc` is one of the legacy atomic functions, just making sure, but that is still required to be fully supported, right? If so, this looks good to me.

I believe so, we test it elsewhere and it's in the 3.0 spec.

PR #490, opened 2024-07-09
Update OpenCL headers

hvdijk:
# Overview

Update OpenCL headers.

# Reason for change

# Description of change

points, etc.*

# Checklist


LICENSE.txt
```diff
@@ -225,8 +225,8 @@ Software.
 Software from third parties included in the oneAPI Construction Kit
 ===================================================================
 
-The source/cl/external/OpenCL-Headers folder (Khronos OpenCL headers), if
-present, is held under its own licensing terms.
+The OpenCL Headers (https://github.com/KhronosGroup/OpenCL-Headers) may be
+fetched under the build directory and is held under its own licensing terma.
```
coldav, line 229: terma -> terms

source/cl/test/UnitCL/source/Common.cpp
```diff
@@ -778,7 +778,7 @@ bool UCL::isImageFormatSupported(cl_context context,
       CASE(CL_RGx);
       CASE(CL_RGBx);
       CASE(CL_DEPTH);
-      CASE(CL_DEPTH_STENCIL);
+      // CASE(CL_DEPTH_STENCIL);
```
coldav, line 781: If we want to leave it commented out can we have a comment (or just delete it). Same below.

LICENSE.txt
hvdijk: Fixed, thanks.

source/cl/test/UnitCL/source/Common.cpp
hvdijk: This is consistent with how it is done in modules/compiler/builtins/libimg/source/host.cpp. But I don't mind deleting it entirely if we do it in both places.

external/CMakeLists.txt
```diff
@@ -14,6 +14,26 @@
...
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
+include(FetchContent)
+
+# Enable C and C++ language extensions which we otherwise disable, as
+# external projects are entitled to assume CMake defaults.
+set(CMAKE_C_EXTENSIONS ON)
+set(CMAKE_CXX_EXTENSIONS ON)
+
+# OpenCL Headers are used throughout the project, so make them available
```
coldav, line 24: I get why you've gone this way, but it would be good to get rid of the libimg dependency and keep it in the source/cl area at some point.
Could add a comment to say that's it's the libimg dependency that requires it rather than throughout.
Just a suggestion though - feel free to ignore.

coldav (approved):

source/cl/test/UnitCL/source/Common.cpp
coldav: I'm happy with deleting in both.

external/CMakeLists.txt
hvdijk: There's more than just that, there's also `modules/mux/targets/host/extension/example/main.c` which is outside of `source/cl`. But I'll edit the comment to focus mainly on libimg.

source/cl/test/UnitCL/source/Common.cpp
hvdijk: Alright, done.

external/CMakeLists.txt
hvdijk: Edited, does it look okay to you like this?

coldav: sure

PR #491, opened 2024-07-10
Minor fixes to guides, ready for release.

coldav:
# Overview

Minor fixes to guides, ready for release.

# Reason for change

Some things confused readers of the guides and we have added more testing for operating systems.


doc/developer-guide.md
```diff
@@ -418,6 +418,10 @@ cmake llvm -GNinja \
   -DLLVM_TARGETS_TO_BUILD='X86;ARM;AArch64'
 ```
 
+Note that `RISCV` will need to be added to `-DLLVM_TARGETS_TO_BUILD` for any
+RISC-V target, and `-DLLVM_ENABLE_PROJECTS=clang;lld` will be required for most
+non "host" targets e.g. `refsi`.
```
hvdijk, line 423: Given that the instructions are not meant to be optimal, merely meant to work, maybe we should just have the instructions say `-DLLVM_ENABLE_PROJECTS='clang;lld' -DLLVM_TARGETS_TO_BUILD='X86;ARM;AArch64;RISCV`. This should work in all cases. We can then add a note saying that lld is only needed in some cases, and it's not necessary to build all targets, but it still gives a command that people can just copy and paste.

doc/developer-guide.md
```diff
@@ -515,6 +519,9 @@ cmake . -GNinja \
   -DCA_LLVM_INSTALL_DIR=$LLVMInstall
 ```
 
+This will build a `host` target, which will run on the machine it is built on.
```
hvdijk, line 522:
```suggestion
This will build a `host` target, which will run device code on the machine that the oneAPI Construction Kit runs on.
```
This is not necessarily the machine that it is built on.

doc/getting-started.rst
```diff
@@ -7,6 +7,7 @@ Platform Dependencies
 Each supported platform has a set of dependencies which are required in order to
 build the oneAPI Construction Kit.
 
+* `Ubuntu 24.04`_
```
hvdijk, line 10: The order here does not match the order below.

doc/getting-started.rst
```diff
@@ -97,15 +98,25 @@ Install the AArch64 toolchain.
...
+some qemu RISC-V features such as the half extension which was added in version
+7.
 
 Windows 10
 ----------
 
 On an x86-64 Windows 10 host system these are the minimum requirements to build
-the oneAPI Construction Kit. These tools are required for any build configuration.
+the oneAPI Construction Kit. These tools are required for any build
+configuration. Windows 11 has not been tested but may work.
```
hvdijk, line 117: It seems odd to describe Windows 11 in a section on Windows 10. Ideally we would just test it though.

doc/getting-started.rst
```diff
@@ -97,15 +98,25 @@ Install the AArch64 toolchain.
...
 
-* `Visual Studio <https://www.visualstudio.com/>`_ 2017 or 2019
+* `Visual Studio <https://www.visualstudio.com/>`_ 2017 or 2019. 2022 may work but is not tested.
```
hvdijk, line 119: It turns out Visual Studio 2017 doesn't work. The Visual Studio CMake generators don't work with any version of Visual Studio, we assume the build configuration is known at CMake time which it isn't with the VS CMake generator, and with the Ninja CMake generator, we get an error because VS2017 comes with a too old version of Ninja. We should update the instructions to use Ninja everywhere, and VS2022 works fine (except in debug builds, but this is not specific to VS2022; I'll create a PR for that).

doc/getting-started.rst
coldav: Ideally we would, but it does not exist in CI (interestingly github uses windows server)

doc/developer-guide.md
coldav: I think that makes sense, yes

doc/getting-started.rst
coldav: I've gone with "should" work and put it in a different section.

hvdijk (approved):

PR #496, opened 2024-07-16
Add support for host target setting of support of features

coldav:
# Overview

Allow features to be added using env variable or cmake using standard  mechanism of "+v,+zfence" etc

# Reason for change

Allow risc-v host to support "+v" etc more easily.

# Description of change

Added cmake (CA_HOST_TARGET_${HOST_ARCH_UPPER}_FEATURES) and env variable (CA_HOST_TARGET_FEATURES) support  to host target.

### DIFF
--- a/doc/developer-guide.md
+++ b/doc/developer-guide.md
@@ -318,24 +318,28 @@ The builtin CMake options used when invoking CMake on the command line.
   use the installed OpenCL or Vulkan library. Do disable this behaviour set
   `-DCMAKE_SKIP_RPATH=ON` when configuring CMake in build directory.
 
-* `CA_HOST_TARGET_<arch>_CPU`: This option is used by the `host` target to
-  optimize for performance on a given CPU.`arch` should be a capitalized
-  version of the `host` target architecture e.g. `X86_64`, `RISCV64` or
-  `AARCH64`. If set to "native" host will optimize for the CPU being used to
-  compile it. Otherwise a CPU name can be provided, for example "skylake", but
-  be warned that this string will be passed directly to the llvm backend so make
-  sure it's a valid CPU name. Information about your host CPU can be found by
-  running`llc --version`, and a list of host CPUs supported by your installed
-  version of LLVM can be found by running`llc --march=[your-arch] --mcpu=help`.
+* `CA_HOST_TARGET_<arch>_CPU`, `CA_HOST_TARGET_<arch>_FEATURES`: These options
+  are used by the `host` target to optimize for performance on a given CPU.
+  `arch` should be a capitalized version of the `host` target architecture e.g.
+  `X86_64`, `RISCV64` or `AARCH64`.
+
+  `CPU` can be set to `native` to optimize for the CPU being used to compile.
+  Otherwise a CPU name can be provided, e.g. `skylake`. This string will be
+  passed directly to the LLVM backend; it has to be a valid CPU name. A list of
+  CPUs supported by LLVM can be found by running `clang -mcpu=help`.
+
+  `FEATURES` should be a comma-separated list of features preceded by either `+`
+  or `-` to enable or disable them, e.g. `+v,-zfencei`. The features are the
+  same as those supported by the `-mattr` option in LLVM tools such as `llc` and
+  `opt` and add to the features supported by default.
+
+  If no `CPU` or `FEATURES` are specified, kernels will be compiled to run on
+  any CPU that meets our minimal assumptions.
   
-  Be aware that if `host` is compiled with this option set, running it on a
-  different CPU from the one specified (or the one compiled with if "native" was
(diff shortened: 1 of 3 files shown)


doc/developer-guide.md
```diff
@@ -320,6 +320,13 @@ The builtin CMake options used when invoking CMake on the command line.
   caveats above apply, and this may result in an illegal instruction crash if
   your CPU doesn't support the generated instructions.
 
+- `CA_HOST_TARGET_<arch>_FEATURES`: This option is used by the `host`` target to
```
hvdijk, line 323: LLVM's naming is a bit inconsistent but in its options this is called "target-specific attributes" and controlled by the `-mattr=` option, despite being called "features" at other levels. I expect it will be clearer to the user to use that terminology and to say that it is processed the same as that option (which I believe you are doing).

modules/compiler/targets/host/source/target.cpp
```diff
@@ -102,6 +102,29 @@ static llvm::TargetMachine *createTargetMachine(llvm::Triple TT,
       /*JIT=*/true);
 }
 
+void UpdateFeatureMapFromString(llvm::StringMap<bool> &FeatureMap,
```
hvdijk, line 105: This looks like it duplicates some logic from LLVM's [`SubtargetFeatures`](https://llvm.org/doxygen/classllvm_1_1SubtargetFeatures.html) class that we could use.

modules/compiler/targets/host/source/target.cpp
```diff
@@ -314,6 +356,12 @@ compiler::Result HostTarget::initWithBuiltins(
   }
 #endif
 
+#ifndef NDEBUG
+  if (const char *E = getenv("CA_HOST_TARGET_FEATURES")) {
+    FeatureStr = E;
```
hvdijk, line 361: If both CMake's `CA_HOST_TARGET_*_FEATURES` and the environment's `CA_HOST_TARGET_FEATURES` are set, I would expect it to behave as if `-mattr=$CA_HOST_TARGET_*_FEATURES -mattr=$CA_HOST_TARGET_FEATURES`, just as setting both CMake's `CA_HOST_TARGET_*_CPU` and the environment's `CA_HOST_TARGET_CPU` are set it acts as `-mcpu=$CA_HOST_TARGET_*_CPU -mcpu=$CA_HOST_TARGET_CPU`. Multiple `-mcpu` options result in the last one winning out, but multiple `-mattr` options don't behave that same way, they result in the options being merged, so I think we should also do that here.

modules/compiler/targets/host/source/target.cpp
coldav: thanks, I've updated to use that.

modules/compiler/targets/host/source/target.cpp
coldav: I think I've done that now.

modules/compiler/targets/host/source/target.cpp
hvdijk: Ha, that's possibly a creative slight misuse of a very easy to accidentally misuse class? The constructor seems to assume that the features already have a `+`/`-` flag. When they don't, I'm not sure LLVM will keep the current behaviour of keeping subtly invalid feature strings, I'm not sure that's something we should rely on. Let me have a better look at this, maybe I'm worried over nothing, it does look as if the effect is exactly what we would want.

coldav: I checked that it "worked" correctly without the +/- flag, and was able to test with hasFlag(), so i think it's okay.

doc/developer-guide.md
```diff
@@ -319,23 +319,28 @@ The builtin CMake options used when invoking CMake on the command line.
...
   different CPU from the one specified (or the one compiled with if "native" was
-  specified) isn't supported and bad things may happen. When the oneAPI
-  Construction Kit is built in debug mode, the environment variable
-  `CA_HOST_TARGET_CPU` is also respected across all `host` targets, which can
-  help track down codegen differences among different machine targets. The
-  caveats above apply, and this may result in an illegal instruction crash if
-  your CPU doesn't support the generated instructions.
+  specified) isn't supported and bad things may happen.
+
+* `CA_HOST_TARGET_<arch>_FEATURES`: This option is used by the `host` target to
```
coldav, line 338: I think it's worth saying that each architecture will have default features which is extended unless the CPU has been overriden and the env variable will extend any features defined in this cmake.

doc/developer-guide.md
```diff
@@ -319,23 +319,28 @@ The builtin CMake options used when invoking CMake on the command line.
...
+  enable features on a given CPU. `arch` should be a capitalized version of the
```
coldav, line 339: maybe enable or disable

PR #501, opened 2024-07-19
Update clUpdateMutableCommandsKHR to match new API

EwanC:
# Overview

Update cl_khr_command_buffer_mutable_dispatch implementation to reflect specification changes merged in https://github.com/KhronosGroup/OpenCL-Docs/issues/1041

Reflects changes from upstream in:

* https://github.com/KhronosGroup/OpenCL-Headers/pull/245
* https://github.com/KhronosGroup/OpenCL-CTS/pull/1984

# Description of change

Updates our implementation from spec revision 0.9.1 of cl_khr_command_buffer_mutable_dispatch to revision 0.9.2 which made a breaking API change to `clUpdateMutableCommandsKHR()`.

# Checklist

### DIFF
--- a/source/cl/examples/MutableDispatchKHR/source/main.cpp
+++ b/source/cl/examples/MutableDispatchKHR/source/main.cpp
@@ -209,9 +209,12 @@ int main(const int argc, const char **argv) {
     // If not executing the first frame
     if (i != 0) {
       // Configure the mutable configuration to update the kernel arguments
-      cl_mutable_dispatch_arg_khr arg_0{0, sizeof(cl_mem), &input_A_buffer};
-      cl_mutable_dispatch_arg_khr arg_1{1, sizeof(cl_mem), &input_B_buffer};
-      cl_mutable_dispatch_arg_khr arg_2{2, sizeof(cl_mem), &output_buffer};
+      const cl_mutable_dispatch_arg_khr arg_0{0, sizeof(cl_mem),
+                                              &input_A_buffer};
+      const cl_mutable_dispatch_arg_khr arg_1{1, sizeof(cl_mem),
+                                              &input_B_buffer};
+      const cl_mutable_dispatch_arg_khr arg_2{2, sizeof(cl_mem),
+                                              &output_buffer};
       cl_mutable_dispatch_arg_khr args[] = {arg_0, arg_1, arg_2};
       cl_mutable_dispatch_config_khr dispatch_config{
           command_handle,
@@ -227,7 +230,7 @@ int main(const int argc, const char **argv) {
           nullptr /* local_work_size */};
 
       // Update the command buffer with the mutable configuration
-      cl_uint num_configs = 1;
+      const cl_uint num_configs = 1;
       cl_command_buffer_update_type_khr config_types[1] = {
           CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR};
       const void *configs[1] = {&dispatch_config};
--- a/source/cl/examples/MutableDispatchKHR/source/main.cpp
+++ b/source/cl/examples/MutableDispatchKHR/source/main.cpp
@@ -215,8 +215,8 @@ int main(const int argc, const char **argv) {
                                               &input_B_buffer};
       const cl_mutable_dispatch_arg_khr arg_2{2, sizeof(cl_mem),
                                               &output_buffer};
-      cl_mutable_dispatch_arg_khr args[] = {arg_0, arg_1, arg_2};
-      cl_mutable_dispatch_config_khr dispatch_config{
+      const cl_mutable_dispatch_arg_khr args[] = {arg_0, arg_1, arg_2};
+      const cl_mutable_dispatch_config_khr dispatch_config{
           command_handle,
           3 /* num_args */,
           0 /* num_svm_arg */,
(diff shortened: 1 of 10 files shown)


coldav (approved):

source/cl/examples/MutableDispatchKHR/source/main.cpp
```diff
@@ -215,10 +215,8 @@ int main(const int argc, const char **argv) {
                                               &input_B_buffer};
       const cl_mutable_dispatch_arg_khr arg_2{2, sizeof(cl_mem),
                                               &output_buffer};
-      const cl_mutable_dispatch_arg_khr args[] = {arg_0, arg_1, arg_2};
-      const cl_mutable_dispatch_config_khr dispatch_config{
-          CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR,
-          nullptr,
+      cl_mutable_dispatch_arg_khr args[] = {arg_0, arg_1, arg_2};
+      cl_mutable_dispatch_config_khr dispatch_config{
```
hvdijk, line 219: With the `clang-tidy` reports on the earlier version for variables that could be `const`, I am surprised that none are raised here. Could you check whether this really needs to be non-`const`, or whether the source file was skipped in the check?

source/cl/source/extension/source/khr_command_buffer.cpp
```diff
@@ -1072,19 +1072,15 @@ cargo::expected<mux_descriptor_info_s, cl_int> createArgumentDescriptor(
...
-                               mutable_config.num_mutable_dispatch);
-
   // Verify struct configures kernel arguments and return error if malformed
-  for (const auto &config : mutable_dispatch_configs) {
-    OCL_CHECK(config.type != CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR,
-              return CL_INVALID_VALUE);
-
+  for (const auto config_ptr : mutable_dispatch_configs) {
+    OCL_CHECK(config_ptr == nullptr, return CL_INVALID_VALUE);
+    const cl_mutable_dispatch_config_khr config = *config_ptr;
```
hvdijk, line 1083: Originally, `config` was a reference, it should probably remain a reference?

source/cl/examples/CMakeLists.txt
```diff
@@ -88,7 +88,7 @@ endif()
 if(${OCL_EXTENSION_cl_khr_command_buffer_mutable_dispatch})
   # Intentionally misnamed. Windows guesses that executables with "patch"
   # in the file name are installers and need admin privileges.
-  add_ca_cl_example(clMutableDispathcKHR
+  add_ca_cl_example(clMutableDispatchKHR
```
hvdijk, line 91: This was not a typo, the comment explains why it had that name.

source/cl/source/extension/source/khr_command_buffer_mutable_dispatch.cpp
```diff
@@ -102,20 +103,24 @@ clUpdateMutableCommandsKHR(cl_command_buffer_khr command_buffer,
...
-                !mutable_config->num_mutable_dispatch,
-            return CL_INVALID_VALUE);
+  for (size_t i = 0; i < num_configs; i++) {
+    if (config_types[i] != CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR) {
+      return CL_INVALID_VALUE;
+    }
+  }
+
+  const cl_mutable_dispatch_config_khr **casted_configs =
+      reinterpret_cast<const cl_mutable_dispatch_config_khr **>(configs);
```
hvdijk, line 119: I don't think this (when dereferencing the result) is well-defined in standard C++, and I do not recall whether current compilers optimize on the assumption that we do not do this. Do you know?

source/cl/examples/MutableDispatchKHR/source/main.cpp
EwanC: `clang-tidy` was being run (I manually run the `tidy-clMutableDispathcKHR` target to check) but the tool didn't seem to mandate `const` on these, but I went through and added `const` anyway

source/cl/source/extension/source/khr_command_buffer.cpp
EwanC: I've made this `config` variable a reference and the loop iteration variable `config_ptr`

source/cl/examples/CMakeLists.txt
EwanC: Reverted this, thanks

source/cl/source/extension/source/khr_command_buffer_mutable_dispatch.cpp
EwanC: So we're looking at the same thing, do you have a reference to where this is UB? I'm looking at https://en.cppreference.com/w/cpp/language/reinterpret_cast and guessing it's point 5) then "Type Accessibility" doesn't seem to cover this case as well defined.


I'm not sure how type based alias analysis works with `void**` that we're only reading from after the cast,  could put this through compiler explorer to double check what some popular compilers are doing. An alternative could be to use a C-style case, but looking at https://en.cppreference.com/w/cpp/language/explicit_cast it seems like this will fallback to `reinterpret_cast`  since `static_cast` isn't valid. The most conservative code is maybe defining a separate variable and memcpying to it.

hvdijk: Yes, it's the type accessibility thing. There is no need to `memcpy`, we can use `cargo::array_view<const void *>` and use a `static_cast` on its elements instead.

EwanC:
I'll just close this, and the other one I have open.

PR #504, opened 2024-07-19
Added some relative relocations for RISC-V

coldav:
# Overview

Added RISCV_BRANCH and RISCV_RVC_JUMP relocations. 

# Reason for change

This is useful for debugging where we assemble the output assembly for testing purposes, but could in theory happen regardless.
# Description of change

Added RISCV_BRANCH and RISCV_RVC_JUMP relocations, as well as skipping RISCV_ALIGN relocations.

### DIFF
--- a/modules/loader/source/relocations.cpp
+++ b/modules/loader/source/relocations.cpp
@@ -684,14 +684,15 @@ uint32_t getLo12(uint64_t addr) {
   return static_cast<uint32_t>(getBitRange(addr, 0, 12));
 }
 
+// Sign extends a number in the bottom 'b' bits of a 64-bit number back up to
+// a 64-bit number.
+int64_t signExtendN(uint64_t val, unsigned b) {
+  CARGO_ASSERT(b > 0, "bit-width cannot be zero");
+  CARGO_ASSERT(b <= 64, "bit-width out of range");
+  return static_cast<int64_t>(val << (64 - b)) >> (64 - b);
+}
+
 cargo::optional<uint32_t> getHi20(int64_t addr) {
-  // Sign extends a number in the bottom 'b' bits of a 64-bit number back up to
-  // a 64-bit number.
-  auto signExtendN = [](uint64_t val, unsigned b) {
-    CARGO_ASSERT(b > 0, "bit-width cannot be zero");
-    CARGO_ASSERT(b <= 64, "bit-width out of range");
-    return static_cast<int64_t>(val << (64 - b)) >> (64 - b);
-  };
   // See documentation for why we add 0x800 here:
   // https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#absolute-addresses
   // https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#pc-relative-symbol-addresses
@@ -710,6 +711,14 @@ cargo::optional<uint32_t> getHi20(int64_t addr) {
 // assumes little-endian RISCV
 bool resolveRISCV(const loader::Relocation &r, loader::ElfMap &map,
                   const std::vector<loader::Relocation> &relocations) {
+  // The ALIGN relocation is special and is related to keeping alignment with
+  // nops after linking. Currently skip this as we believe this is not required.
+  // Additionally it has no symbol which decomposeRelocation does not handle, so
+  // this would need fixed if this is ever implemented.
+  if (r.type == loader::RelocationTypes::RISCV::R_RISCV_ALIGN) {
+    // TODO: Check that we fit the alignment constraints as is.
+    return true;
+  }
   const auto relocation_data = decomposeRelocation<uint64_t>(r, map);
   if (!relocation_data) {
     return false;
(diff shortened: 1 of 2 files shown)


modules/loader/source/relocations.cpp
```diff
@@ -832,6 +839,37 @@ bool resolveRISCV(const loader::Relocation &r, loader::ElfMap &map,
       writeITypeImm(getLo12(relative_value), relocation_address + 4);
       break;
     }
+    case R_RISCV_BRANCH: {
+      // R_RISCV_BRANCH immediate is for B-Type instructions and is split
+      // across 12 bits spread across the instruction and skips the bottom bit
+      // see conditional branches in the isa spec
+      uint32_t value;
+      const uint32_t trunc_value = static_cast<uint32_t>(relative_value);
```
coldav, line 847: This is actually signed so I need to convert to int32_t before shifting.

modules/loader/source/relocations.cpp
```diff
@@ -832,6 +839,37 @@ bool resolveRISCV(const loader::Relocation &r, loader::ElfMap &map,
...
+      break;
+    }
+
+    case R_RISCV_RVC_JUMP: {
+      // R_RISCV_RVC_JUMP is for compressed instructions and fits in bits 2-12
+      // of a 16 bit value and skips the bottom bit. See CJ format in the isa
+      // spec.
+      uint16_t value;
+      cargo::read_little_endian(&value, relocation_address);
+      const uint16_t offset_16 = static_cast<uint16_t>(relative_value) >> 1;
```
coldav, line 867: This is signed so need to convert to int16_t before casting.

hvdijk:
We're not exactly consistent with what checks we perform on the relocations we handle already, some we assert that the values are in range, others we silently discard high bits. For these completely new ones, could we add the assertions that the new relocations we're handling with this are as we expect -- that the instruction following an `ALIGN` relocation is aligned, that the values for `BRANCH` and `RVC_JUMP` relocations are in range?

modules/loader/source/relocations.cpp
```diff
@@ -832,6 +839,51 @@ bool resolveRISCV(const loader::Relocation &r, loader::ElfMap &map,
...
       break;
     }
+    case R_RISCV_BRANCH: {
+      // R_RISCV_BRANCH immediate is for B-Type instructions and is split
+      // across 12 bits spread across the instruction and skips the bottom bit
+      // see conditional branches in the isa spec
+      int64_t signed_relative_value = static_cast<int64_t>(relative_value);
+      uint64_t abs_val =
+          signed_relative_value < 0 ? -signed_relative_value : relative_value;
+      if (abs_val & (~0x8FFE)) {
```
hvdijk, line 849: This check does not look right. The constant used here doesn't have consecutive bits set, and by checking absolute value, the largest valid negative value is rejected.

We might be able to fix this very simply:  in `getHi20`, we do
```
  // Sign extends a number in the bottom 'b' bits of a 64-bit number back up to
  // a 64-bit number.
  auto signExtendN = [](uint64_t val, unsigned b) {
    CARGO_ASSERT(b > 0, "bit-width cannot be zero");
    CARGO_ASSERT(b <= 64, "bit-width out of range");
    return static_cast<int64_t>(val << (64 - b)) >> (64 - b);
  };
  ...
  // Check that the value, when sign-extended back up to a 64-bit number (e.g.,
  // through LUI) will not lose bits.
  if ((addr >> 12) != signExtendN(addr >> 12, 20)) {
```
If we make `signExtendN` a function rather than just a helper lambda, we could do the check that way here too.

hvdijk: This looks good, thanks. There's still the previous version of the code commented out which I assume was an accident.

To make sure it doesn't get missed, commenting here as well: could you add a check that the instruction after an `ALIGN` relocation is suitably aligned?

coldav: Sure yes. Will tidy up before committing

modules/loader/source/relocations.cpp
```diff
@@ -832,6 +841,49 @@ bool resolveRISCV(const loader::Relocation &r, loader::ElfMap &map,
...
+      cargo::write_little_endian(value, relocation_address);
+      break;
+    }
+
+    case R_RISCV_RVC_JUMP: {
+      // R_RISCV_RVC_JUMP is for compressed instructions and fits in bits 2-12
+      // of a 16 bit value and skips the bottom bit. See CJ format in the isa
+      // spec.
+      if (signExtendN(relative_value, 13) !=
+          static_cast<int64_t>(relative_value & (~1))) {
```
hvdijk, line 874: This check does not match what it says it checks. This checks that the vaue fits in 13 bits (and that the bottom bit is zero) but the below assert says it needs to fit in 12 bits. I *think* the text is correct and this should be
```suggestion
      if (signExtendN(relative_value, 12) !=
          static_cast<int64_t>(relative_value & (~1))) {
```
but please double-check, I could be wrong.

coldav: It's 13 bits because the bottom bit gets dropped. I could also have shifted it by one  but then couldn't sneak in the ~1 bit.

coldav: I'll add some comments to make this clear

coldav: yes you are right, this one is different.

modules/loader/source/relocations.cpp
```diff
@@ -832,6 +841,54 @@ bool resolveRISCV(const loader::Relocation &r, loader::ElfMap &map,
...
+    case R_RISCV_BRANCH: {
+      // R_RISCV_BRANCH immediate is for B-Type instructions and is split
+      // across 12 bits spread across the instruction and skips the bottom bit
+      // of the input relative offset. See conditional branches in the isa spec.
+
+      // Check that the value, when sign-extended back up to a 64-bit number
+      // will not lose bits.
+      if (signExtendN(relative_value, 13) !=
+          static_cast<int64_t>(relative_value & (~1))) {
+        CARGO_ASSERT(0, "R_RISCV_BRANCH branch > 13 bits or not even.");
```
hvdijk, line 853: `CARGO_ASSERT` means that in debug builds, a relocation we cannot support will cause the whole program to abort, but the `return false` was in order to actually try to handle it properly by reporting a failure to the caller, right? In that case, I think we just want to print a message in debug builds and keep running? And likewise for `R_RISCV_RVC_JUMP`.

coldav: I'm not sure it's really recoverable if we get this error tbh, since we can't do the relocation, but I'll do it if you prefer (note we do a CARGO_ASSERT - note we're not entirely consistent on this.

hvdijk: We're definitely not consistent, agreed on that. If we return false, we ultimately end up at
```
  if (!loader::resolveRelocations(*elf_file, elf_map)) {
    return mux_error_internal;
  }
```
which is *almost* already what I think we would want long term: long term I think we would still want to return an error there, but have it be `mux_error_invalid_binary` which should get translated to `CL_INVALID_BINARY`. That still won't work, of course, but it's a standard error code that allows a program to report the error in a user-friendly way.

hvdijk (approved):

PR #507, opened 2024-07-24
[cargo, host] Handle macOS.

hvdijk:
# Overview

[cargo, host] Handle macOS.

# Reason for change

# Description of change

* pthread_setname_np only takes a single argument and can only set the thread name of the current thread. Where cargo optionally uses pthread_setname_np, it needs the two-argument version, so change the CMake check to reject the macOS version.
* For macOS/arm64, we need the same workaround as for macOS/x86, we need to generate ELF since Mach-O has restrictions on names that means our IR cannot be compiled for it. This makes things easier for us anyway as we already have an ELF loader.

# Checklist

### DIFF
--- a/modules/compiler/targets/host/source/target.cpp
+++ b/modules/compiler/targets/host/source/target.cpp
@@ -186,9 +186,24 @@ compiler::Result HostTarget::initWithBuiltins(
       triple = llvm::Triple("armv7-unknown-linux-gnueabihf-elf");
       break;
     case host::arch::AARCH64:
-      assert(host::os::LINUX == host_device_info.os &&
-             "AArch64 cross-compile only supports Linux");
-      triple = llvm::Triple("aarch64-linux-gnu-elf");
+      switch (host_device_info.os) {
+        case host::os::LINUX:
+          triple = llvm::Triple("aarch64-linux-gnu-elf");
+          break;
+        case host::os::MACOS:
+          assert(host_device_info.native &&
+                 "macOS cross-compile not supported");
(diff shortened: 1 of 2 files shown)


modules/compiler/targets/host/source/target.cpp
```diff
@@ -186,9 +186,21 @@ compiler::Result HostTarget::initWithBuiltins(
...
-             "AArch64 cross-compile only supports Linux");
-      triple = llvm::Triple("aarch64-linux-gnu-elf");
+      switch (host_device_info.os) {
+        case host::os::LINUX:
+          triple = llvm::Triple("aarch64-linux-gnu-elf");
+          break;
+        case host::os::MACOS:
+          assert(host_device_info.native &&
+                 "macOS cross-compile not supported");
+          // Our generated code does not meet Mach-O restrictions. Force ELF
```
coldav, line 196: I think it would be helpful to say what the restriction is like in the PR description.

coldav (approved):

modules/compiler/targets/host/source/target.cpp
hvdijk: Agreed. During a simple build, we get during the execution of `clc`:
```
LLVM ERROR: Global variable 'notes_global' has an invalid section specifier 'notes': mach-o section specifier requires a segment and section separated by a comma.
```
I'll include this somewhere. As I recall there were other restrictions as well but I don't recall what they were and they don't show up in just a build.

coldav (approved):

PR #510, opened 2024-07-25
Fix x86 compiler, disable test for macOS.

hvdijk:
# Overview

Fix x86 compiler, disable test for macOS.

# Reason for change

# Description of change

* We were setting up the x86 compiler for the host OS, rather than for Linux, but we only support cross-compilation for Linux. For AArch64 and RISC-V, we were already forcing Linux, do the same for x86 so that we do not attempt to cross-compile from macOS/arm64 to macOS/x86.
* Disable Precision_01_Pow_Func on macOS. This test only passes when a sufficiently large long double type is available. On macOS/arm64, long double is the same as double.

# Checklist

### DIFF
--- a/modules/compiler/targets/host/source/info.cpp
+++ b/modules/compiler/targets/host/source/info.cpp
@@ -130,11 +130,11 @@ void HostInfo::get(compiler::AddCompilerFn add_compiler) {
       HOST_CROSS_DEVICE_NAME_AARCH64));
 #endif
 #ifdef HOST_CROSS_X86
-  add_compiler(getCrossCompilerInfo<host::arch::X86, HOST_OS>(
+  add_compiler(getCrossCompilerInfo<host::arch::X86, host::os::LINUX>(
       HOST_CROSS_DEVICE_NAME_X86));
 #endif
 #ifdef HOST_CROSS_X86_64
-  add_compiler(getCrossCompilerInfo<host::arch::X86_64, HOST_OS>(
+  add_compiler(getCrossCompilerInfo<host::arch::X86_64, host::os::LINUX>(
       HOST_CROSS_DEVICE_NAME_X86_64));
 #endif
 #ifdef HOST_CROSS_RISCV32
--- a/source/cl/test/UnitCL/source/ktst_precision.cpp
+++ b/source/cl/test/UnitCL/source/ktst_precision.cpp
@@ -107,11 +107,9 @@ T Remquo7BitRef(T x, T y, cl_int &quo_out) {
 }
 }  // namespace
 
-#ifdef __arm__
-// TODO: CA-2730
-TEST_P(Execution, DISABLED_Precision_01_Pow_Func) {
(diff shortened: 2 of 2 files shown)


modules/compiler/targets/host/source/info.cpp
```diff
@@ -130,11 +130,11 @@ void HostInfo::get(compiler::AddCompilerFn add_compiler) {
       HOST_CROSS_DEVICE_NAME_AARCH64));
 #endif
 #ifdef HOST_CROSS_X86
-  add_compiler(getCrossCompilerInfo<host::arch::X86, HOST_OS>(
+  add_compiler(getCrossCompilerInfo<host::arch::X86, host::os::LINUX>(
```
coldav, line 133: The title is a bit misleading given this change as well.

coldav (approved):

source/cl/test/UnitCL/source/ktst_precision.cpp
```diff
@@ -107,11 +107,9 @@ T Remquo7BitRef(T x, T y, cl_int &quo_out) {
 }
 }  // namespace
 
-#ifdef __arm__
-// TODO: CA-2730
-TEST_P(Execution, DISABLED_Precision_01_Pow_Func) {
-#elif defined(_WIN32)
-// TODO: CA-3492
+#if defined(__arm__) || defined(_WIN32) || defined(__APPLE__)
```
coldav, line 110: totally understand why you've done this, but it feels a shame to not be able to run this test in some way on these architectures.

hvdijk: Until we can fix the test to work properly, we can force it to run, if wanted, using the `--gtest_also_run_disabled_tests` option. It will require a rename of the kernel file (`precision.01_pow_func.cl` -> `disabled_precision.01_pow_func.cl`).

modules/compiler/targets/host/source/info.cpp
hvdijk: The goal here is to disable a test but agreed it doesn't exactly disable a test. Without this, we get

```console
% bin/UnitCompiler
[...]
[----------] 3 tests from CreateBinaryTest
[ RUN      ] CreateBinaryTest.CreateBinary/ComputeAorta_AArch64
[       OK ] CreateBinaryTest.CreateBinary/ComputeAorta_AArch64 (24 ms)
[ RUN      ] CreateBinaryTest.CreateBinary/ComputeAorta_x86_64
Assertion failed: (host_device_info.native && "macOS cross-compile not supported"), function initWithBuiltins, file target.cpp, line 244.
zsh: abort      bin/UnitCompiler
```

I'll try to reword.

hvdijk: Couldn't come up with wording that covered both, so just edited the title to separate the two changes.

coldav (approved):

PR #523, opened 2024-09-03
github workflow cross compiling for aarch64

coldav:
# Overview

Add github workflow cross-compilation of aarch64

# Reason for change

Allow aarch64 testing of OCK

# Description of change

Modified workflow to support extra flags, and cache naming to include arch.

### DIFF
--- a/.github/workflows/create_llvm.yml
+++ b/.github/workflows/create_llvm.yml
@@ -6,6 +6,9 @@ on:
       - main
     paths:
       - '.github/workflows/create_llvm.yml'
+  pull_request:
+    paths:
+      - '.github/workflows/create_llvm.yml'         
   workflow_dispatch:
 
 jobs:
@@ -15,18 +18,33 @@ jobs:
         version: [17, 18]
         os: [ubuntu-22.04]
         build_type: [Release, RelAssert]
+        arch : [x86_64, aarch64]
+        exclude:
+          # For now just do latest llvm with aarch to reduce cache usage.
+          - arch: aarch64
+            version: 17
+          - arch: aarch64
+            build_type: Release
         include:
           # We want to set flags related to particular matrix dimensions. To do this
           # we need to create default values first, and then against particular matrix
           # dimensions.
           # Note that we need to use RelAssert as the cache key matching can match Release against ReleaseAssert
           - os_flags:
+          - arch_flags:
           - build_type_flags:
-          - build_type: RelAssert
-            build_type_flags: -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON
+          - build_type_flags: -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON
+            build_type: RelAssert
+          - arch_flags: -DCMAKE_TOOLCHAIN_FILE="$GITHUB_WORKSPACE/ock/platform/arm-linux/aarch64-toolchain.cmake"
+                        -DLLVM_TARGET_ARCH=AArch64
+                        -DLLVM_HOST_TRIPLE=aarch64-unknown-linux-gnu
+                        -DLLVM_DEFAULT_TARGET_TRIPLE=aarch64-unknown-linux-gnu
+                        -DLLVM_TABLEGEN="$GITHUB_WORKSPACE/build_native/bin/llvm-tblgen"
+                        -DCLANG_TABLEGEN="$GITHUB_WORKSPACE/build_native/bin/clang-tblgen"
+            arch: aarch64
           - build_type: Release
             build_type_flags: -DCMAKE_BUILD_TYPE=Release
-    
+
     runs-on: ${{ matrix.os }}
     steps:
       - name: Cache llvm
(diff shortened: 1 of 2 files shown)


.github/workflows/create_llvm.yml
```diff
@@ -35,23 +53,61 @@ jobs:
...
+      - name: install aarch64 build tools
+        if: ${{ steps.cache.outputs.cache-hit != 'true' && matrix.arch == 'aarch64' }}
+        run:
+          sudo apt-get install --yes gcc-11-aarch64-linux-gnu g++-11-aarch64-linux-gnu
+
       - name: Flags checker
         if: steps.cache.outputs.cache-hit != 'true'
         run:
-          echo Building on "${{ matrix.os }}" with os_flags "${{ matrix.os_flags}}" extra flags "${{ matrix.build_type_flags}}" and  build_type "${{matrix.build_type}}"
+          echo Building on "${{ matrix.os }}" with os_flags "${{ matrix.os_flags}}" aarch "${{ matrix.arch }}"extra flags "${{ matrix.build_type_flags}}" and  build_type "${{matrix.build_type}}"
```
hvdijk, line 90:
```suggestion
          echo Building on "${{ matrix.os }}" with os_flags "${{ matrix.os_flags}}" arch "${{ matrix.arch }}" extra flags "${{ matrix.build_type_flags}}" and  build_type "${{matrix.build_type}}"
```

.github/workflows/create_llvm.yml
```diff
@@ -35,23 +53,61 @@ jobs:
...
+          path: ock          
+
       - name: Install Ninja
         if: steps.cache.outputs.cache-hit != 'true'
         uses: llvm/actions/install-ninja@main
 
+      - name: install aarch64 build tools
+        if: ${{ steps.cache.outputs.cache-hit != 'true' && matrix.arch == 'aarch64' }}
+        run:
+          sudo apt-get install --yes gcc-11-aarch64-linux-gnu g++-11-aarch64-linux-gnu
```
hvdijk, line 85:
```suggestion
          sudo apt-get install --yes g++-11-aarch64-linux-gnu
```
? `gcc-11-aarch64-linux-gnu` should already be a dependency of that.

.github/workflows/create_llvm.yml
```diff
@@ -15,18 +18,33 @@ jobs:
...
           # Note that we need to use RelAssert as the cache key matching can match Release against ReleaseAssert
           - os_flags:
+          - arch_flags:
           - build_type_flags:
-          - build_type: RelAssert
-            build_type_flags: -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON
+          - build_type_flags: -DCMAKE_BUILD_TYPE=Release -DLLVM_ENABLE_ASSERTIONS=ON
+            build_type: RelAssert
+          - arch_flags: -DCMAKE_TOOLCHAIN_FILE="$GITHUB_WORKSPACE/ock/platform/arm-linux/aarch64-toolchain.cmake"
+                        -DLLVM_TARGET_ARCH=AArch64
```
hvdijk, line 39: `LLVM_TARGET_ARCH` should not be necessary, that's for when you build LLVM *as* a cross-compiler, rather than when you build LLVM *using* a cross-compiler.

.github/workflows/create_llvm.yml
```diff
@@ -15,18 +18,33 @@ jobs:
...
+                        -DLLVM_HOST_TRIPLE=aarch64-unknown-linux-gnu
+                        -DLLVM_DEFAULT_TARGET_TRIPLE=aarch64-unknown-linux-gnu
```
hvdijk, line 41: `LLVM_DEFAULT_TARGET_TRIPLE` should not be necessary, it defaults to `LLVM_HOST_TRIPLE`.

.github/workflows/create_llvm.yml
hvdijk: BTW, I'm not entirely sure whether installing packages during every run is fine, ideally we would cache this (https://github.com/marketplace/actions/cache-apt-packages) but I understand we are already hitting the cache limits.

.github/workflows/create_llvm.yml
```diff
@@ -35,23 +53,61 @@ jobs:
...
+        run:
+          sudo apt-get install --yes gcc-11-aarch64-linux-gnu g++-11-aarch64-linux-gnu
+
       - name: Flags checker
         if: steps.cache.outputs.cache-hit != 'true'
         run:
-          echo Building on "${{ matrix.os }}" with os_flags "${{ matrix.os_flags}}" extra flags "${{ matrix.build_type_flags}}" and  build_type "${{matrix.build_type}}"
+          echo Building on "${{ matrix.os }}" with os_flags "${{ matrix.os_flags}}" aarch "${{ matrix.arch }}"extra flags "${{ matrix.build_type_flags}}" and  build_type "${{matrix.build_type}}"
+
+      - name: cmake native for cross
```
hvdijk, line 92: This name confuses me.

.github/workflows/create_llvm.yml
coldav: I'll look into caching at some point, but at least this has consistency with what we do elsewhere. I'll try dropping one - I was just matching what we install in gitlab, but will try your suggestion.

.github/workflows/create_llvm.yml
coldav: I can drop it, but it's what we have in our documentation.

.github/workflows/create_llvm.yml
coldav: okay, but again it's in the documentation

.github/workflows/create_llvm.yml
coldav: okay, again i was following the documentation we have which says this is required.
Does it do it automatically?

.github/workflows/create_llvm.yml
hvdijk: Yeah, there's a bunch of things in the documentation and in our internal GitLab builds that are not necessary but also not harmful. I've just ignored them but if we're moving things over to GitHub now seems like a good time to clean that up.

.github/workflows/create_llvm.yml
hvdijk: I believe so, yes, this is conditional on `LLVM_USE_HOST_TOOLS` which is always set to `ON` when cross-compiling.

hvdijk (approved):

PR #534, opened 2024-09-24
Add windows build of llvm to testing

coldav:
# Overview

Add windows to list of builds for llvm

# Reason for change

Required for some our OCK testing.

### DIFF
--- a/.github/workflows/create_llvm.yml
+++ b/.github/workflows/create_llvm.yml
@@ -16,15 +16,22 @@ jobs:
     strategy:
       matrix:
         version: [17, 18]
-        os: [ubuntu-22.04]
+        os: [ubuntu-22.04, windows-2019]
         build_type: [Release, RelAssert]
         arch : [x86_64, aarch64]
         exclude:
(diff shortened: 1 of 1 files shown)


.github/workflows/create_llvm.yml
```diff
@@ -16,15 +16,21 @@ jobs:
     strategy:
       matrix:
         version: [17, 18]
-        os: [ubuntu-22.04]
+        os: [ubuntu-22.04, windows-2019]
         build_type: [Release, RelAssert]
         arch : [x86_64, aarch64]
         exclude:
-          # For now just do latest llvm with aarch to reduce cache usage.
+          # For now just do latest llvm with aarch/windows to reduce cache usage.
```
hvdijk, line 23: I get what you're saying here, but only by also comparing it to what you have down below. On its own it reads as a comment on what we do on Windows-on-ARM which (we don't support).

hvdijk (approved):

.github/workflows/create_llvm.yml
coldav: I'll tweak what I say.

PR #535, opened 2024-09-24
[NFC] Fix CMakeCache comments.

hvdijk:
# Overview

[NFC] Fix CMakeCache comments.

# Reason for change

We were using the wrong host in the comment for e.g. CA_HOST_TARGET_AARCH64_FEATURES so that depending on configuration, we might document that as controlling RISCV64 features, and giving RISCV64-specific features as an example.

# Description of change

- Leave off "e.g. '+v,+zfh'" when the target is not RISC-V.
- For cross compilers, use the correct host.

# Checklist

### DIFF
--- a/modules/compiler/targets/host/CMakeLists.txt
+++ b/modules/compiler/targets/host/CMakeLists.txt
@@ -188,8 +188,13 @@ if(CA_HOST_CROSS_COMPILERS)
     target_compile_definitions(compiler-host PRIVATE
     CA_HOST_TARGET_${HOST_ARCH_UPPER}_CPU="${CA_HOST_TARGET_${HOST_ARCH_UPPER}_CPU}")
   endif()
+  set(CA_HOST_TARGET_${HOST_ARCH_UPPER}_FEATURES_DESC
+    "Feature list that host ${HOST_ARCH_UPPER} should enable or disable as a comma separated + or - list")
+  if(HOST_ARCH_UPPER STREQUAL RISCV32 OR HOST_ARCH_UPPER STREQUAL RISCV64)
+    string(APPEND CA_HOST_TARGET_${HOST_ARCH_UPPER}_FEATURES_DESC " e.g. '+v,+zfh'")
+  endif()
   ca_option(CA_HOST_TARGET_${HOST_ARCH_UPPER}_FEATURES STRING
-    "Feature list that host ${HOST_ARCH_UPPER} should enable or disable as a comma separated + or - list e.g. '+v,+zfh'" "")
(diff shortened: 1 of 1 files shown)


PietroGhg (approved):

modules/compiler/targets/host/CMakeLists.txt
```diff
@@ -188,8 +188,13 @@ if(CA_HOST_CROSS_COMPILERS)
     target_compile_definitions(compiler-host PRIVATE
     CA_HOST_TARGET_${HOST_ARCH_UPPER}_CPU="${CA_HOST_TARGET_${HOST_ARCH_UPPER}_CPU}")
   endif()
+  set(CA_HOST_TARGET_${HOST_ARCH_UPPER}_FEATURES_DESC
+    "Feature list that host ${HOST_ARCH_UPPER} should enable or disable as a comma separated + or - list")
+  if(HOST_ARCH_UPPER STREQUAL RISCV32 OR HOST_ARCH_UPPER STREQUAL RISCV64)
+    string(APPEND CA_HOST_TARGET_${HOST_ARCH_UPPER}_FEATURES_DESC " e.g. '+v,+zfh'")
```
coldav, line 194: I get where you are coming from, but the intention was to show the form of it. However, if you found it confusing as is, then I'll accept your change.

hvdijk: If the arch name hadn't been wrong, I probably would not have found it nearly as confusing, but looking at what we end up with (other variables omitted):

```diff
-//Feature list that host RISCV64 should enable or disable as a
-// comma separated + or - list e.g. '+v,+zfh'
+//Feature list that host AARCH64 should enable or disable as a
+// comma separated + or - list
 CA_HOST_TARGET_AARCH64_FEATURES:STRING=

 //Feature list that host RISCV64 should enable or disable as a
 // comma separated + or - list e.g. '+v,+zfh'
 CA_HOST_TARGET_RISCV64_FEATURES:STRING=

-//Feature list that host RISCV64 should enable or disable as a
-// comma separated + or - list e.g. '+v,+zfh'
+//Feature list that host X86_64 should enable or disable as a comma
+// separated + or - list
 CA_HOST_TARGET_X86_64_FEATURES:STRING=
```

This does manage to still show off the form IMO, as these variables are all closely together. But I'd also be happy with a way of showing off the form on each variable if we can find one that avoids giving an example that is never valid for some of the variables.

coldav: I think it is okay as it's documented in the developer guide.

PR #542, opened 2024-10-01
Dependency pinning

carlewis:
# Overview

Pin external workflow actions to the commit hash. Also fix Scorecard badge.

# Reason for change

Pinning external workflow actions to the hashed version is a good practice to help against supply chain attacks. This change will raise the OpenSSF score too.

# Description of change

External workflow actions have been pinned to the commit hash. The only actions not pinned are `llvm/actions/install-ninja` and `llvm/actions/setup-windows` as they don't have releases.

Dependabot will open PRs in the future, whenever pinned actions need update. 

By default, PRs open by dependabot to update dependencies will be tagged to be reviewed by @codeplaysoftware/security-managers and @codeplaysoftware/ock-workflow-reviewers as the code owners. 

# Checklist

### DIFF
--- a/.github/actions/build_portBLAS_action/action.yml
+++ b/.github/actions/build_portBLAS_action/action.yml
@@ -46,7 +46,7 @@ runs:
         tar -cvzf portBLAS_build.tar.gz portBLAS_build_dir
 
     - name: Upload Artifacts
-      uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
+      uses: actions/upload-artifact@v4
       with:
         name: portBLAS_build
         path: portBLAS_build.tar.gz
--- a/.github/actions/build_portBLAS_action/action.yml
+++ b/.github/actions/build_portBLAS_action/action.yml
@@ -46,7 +46,7 @@ runs:
         tar -cvzf portBLAS_build.tar.gz portBLAS_build_dir
 
     - name: Upload Artifacts
-      uses: actions/upload-artifact@v4
+      uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
       with:
         name: portBLAS_build
         path: portBLAS_build.tar.gz
--- a/.github/actions/build_portDNN_action/action.yml
+++ b/.github/actions/build_portDNN_action/action.yml
@@ -47,7 +47,7 @@ runs:
         tar -cvzf portDNN_build.tar.gz portDNN_build_dir
 
     - name: Upload Artifacts
-      uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
+      uses: actions/upload-artifact@v4
       with:
         name: portDNN_build
         path: portDNN_build.tar.gz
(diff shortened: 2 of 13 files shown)


hvdijk:
Dependency pinning is mentioned at https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions for third-party actions and makes sense in that context, so I'd be happy with the change to `hendrikmuhs/ccache-action`, no question there. When we do it for GitHub's own actions though, the attack we are trying to protect ourselves against, isn't it GitHub itself being compromised? GitHub's documentation for GitHub's actions does not suggest doing this, so if we deviate from their suggestions, I'd like to be sure we're doing this for good reasons, since it means we no longer get any bugfixes either.

carlewis:
> Dependency pinning is mentioned at https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions for third-party actions and makes sense in that context, so I'd be happy with the change to `hendrikmuhs/ccache-action`, no question there. When we do it for GitHub's own actions though, the attack we are trying to protect ourselves against, isn't it GitHub itself being compromised? GitHub's documentation for GitHub's actions does not suggest doing this, so if we deviate from their suggestions, I'd like to be sure we're doing this for good reasons, since it means we no longer get any bugfixes either.

Other than GitHub's security guide, we also follow [OpenSSF scorecard](https://scorecard.dev/viewer/?uri=github.com/codeplaysoftware/oneapi-construction-kit) recommendations. Dependency pinning is a medium level check. It doesn't only apply to GitHub actions,  but to Python pip or other package managers dependencies, so this PR won't get us the maximum score, but will improve the value.

hvdijk:
> Other than GitHub's security guide, we also follow [OpenSSF scorecard](https://scorecard.dev/viewer/?uri=github.com/codeplaysoftware/oneapi-construction-kit) recommendations. Dependency pinning is a medium level check. It doesn't only apply to GitHub actions, but to Python pip or other package managers dependencies, so this PR won't get us the maximum score, but will improve the value.

I know this is in no way your fault, but I am not impressed with the referenced https://github.com/step-security/secure-repo?tab=readme-ov-file#3-pin-actions-to-a-full-length-commit-sha which specifically says (emphasis mine)

> GitHub's Security Hardening for GitHub Actions guide [recommends pinning actions to full length commit **for third party actions**](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

and then gives an example

```diff
-      uses: actions/checkout@v3
+      uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # tag: v3
```

If they want to deviate from GitHub's recommendations, they should not refer to GitHub's recommendations as a justification.

hvdijk:
They have an open issue for addressing that because they have received similar feedback before, https://github.com/step-security/secure-repo/issues/1606.

carlewis:
@hvdijk I just updated the branch focusing the version pinning on third party actions.

hvdijk:
Thanks, I see there were already GitHub actions pinned to specific commits in `.github/workflows/scorecard.yml` so it makes sense that you just kept that as is. Looks good to me, then, but will wait for @coldav to comment since this doesn't just pin actions, it also updates them, and he will know better whether there's any actions there where we need to check that that's okay.

coldav (approved):

PR #544, opened 2024-10-02
x86 updates to create_llvm

alan-forbes-cp:
# Overview

Add x86 support to create_llvm

# Reason for change

x86 support

# Description of change

Add x86 llvm to cache. Tested in my fork.

### DIFF
--- a/.github/workflows/create_llvm.yml
+++ b/.github/workflows/create_llvm.yml
@@ -20,11 +20,15 @@ jobs:
         build_type: [Release, RelAssert]
         arch : [x86, x86_64, aarch64]
         exclude:
-          # Reduce the versions of llvm for aarch64 architecture, as
+          # Reduce the versions of llvm for aarch64 & x86 architectures, as
           # well as any windows os variants to reduce cache usage.
           - arch: aarch64
             version: 17
           - arch: aarch64
+            build_type: Release
+          - arch: x86
+            version: 17
+          - arch: x86
             build_type: Release
           - os: windows-2019
             version: 17
--- a/.github/workflows/create_llvm.yml
+++ b/.github/workflows/create_llvm.yml
@@ -36,6 +36,8 @@ jobs:
             build_type: Release
           - os: windows-2019
             arch: aarch64
+          - os: windows-2019
+            arch: x86
         include:
           # We want to set flags related to particular matrix dimensions. To do this
           # we need to create default values first, and then against particular matrix
--- a/.github/workflows/create_llvm.yml
+++ b/.github/workflows/create_llvm.yml
@@ -52,6 +52,7 @@ jobs:
                         -DLLVM_HOST_TRIPLE=aarch64-unknown-linux-gnu
             arch: aarch64
           - arch_flags: -DLLVM_BUILD_32_BITS=ON -DLIBXML2_LIBRARIES=IGNORE -DLLVM_ENABLE_TERMINFO=OFF
+                        -DLLVM_HOST_TRIPLE=i686-unknown-linux-gnu
             arch: x86
           - build_type: Release
             build_type_flags: -DCMAKE_BUILD_TYPE=Release
--- a/.github/workflows/create_llvm.yml
+++ b/.github/workflows/create_llvm.yml
@@ -18,7 +18,7 @@ jobs:
         version: [17, 18]
         os: [ubuntu-22.04, windows-2019]
         build_type: [Release, RelAssert]
-        arch : [x86_64, aarch64]
+        arch : [x86, x86_64, aarch64]
         exclude:
           # Reduce the versions of llvm for aarch64 architecture, as
           # well as any windows os variants to reduce cache usage.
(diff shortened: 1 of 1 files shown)


.github/workflows/create_llvm.yml
```diff
@@ -45,6 +45,8 @@ jobs:
           - arch_flags: -DCMAKE_TOOLCHAIN_FILE="$GITHUB_WORKSPACE/ock/platform/arm-linux/aarch64-toolchain.cmake"
                         -DLLVM_HOST_TRIPLE=aarch64-unknown-linux-gnu
             arch: aarch64
+          - arch_flags: -DLLVM_BUILD_32_BITS=ON -DLIBXML2_LIBRARIES=IGNORE -DLLVM_ENABLE_TERMINFO=OFF
```
hvdijk, line 48: This matches what we do on our other CI, but the second and third options are workaround for `LLVM_BUILD_32_BITS` being implemented in a way that doesn't handle everything that's needed, and these wouldn't be needed if we cross-compiled for 32-bit x86 in the same way that we do for AArch64. Maybe this is a good time to switch to that?

Does `LLVM_HOST_TRIPLE` need to be set too, or is that done automatically by `LLVM_BUILD_32_BITS`?

hvdijk: `LLVM_HOST_TRIPLE` is set for cross builds because the value that would be used implicitly is wrong, but that applies to 32-bit vs. 64-bit builds as well. I checked our internal CI now, in our current build with `-DLLVM_BUILD_32_BITS=ON`, the log includes:
```
-- LLVM host triple: x86_64-unknown-linux-gnu
...
-- LLVM default target triple: x86_64-unknown-linux-gnu
```
We probably get away with it because we only use this build in limited ways, and those limited ways do always set the target triple, but it's still wrong, this is not the right host or target triple for 32-bit x86.

coldav: if it's easy to do it in a cross way then sure, that makes sense - not sure how much Alan understands about it though.

alan-forbes-cp: I assume the triple required is `x86-unknown-linux-gnu` - in which case I can add in the same way as is done for aarch64.

hvdijk: It would be `i386-unknown-linux-gnu` or `i686-unknown-linux-gnu` depending on whether the build environment defaults to assuming newer processors. I do not know what the build environment in GitHub is like, @coldav do you know?

coldav: I don't know.

hvdijk: Okay, I checked, my local system's `gcc -v` includes `--with-arch-32=i686`, that's the Ubuntu-supplied GCC, and I think we use the Ubuntu-supplied GCC in GitHub too? It should be `i686-unknown-linux-gnu`, then.

.github/workflows/create_llvm.yml
```diff
@@ -36,6 +36,8 @@ jobs:
             build_type: Release
           - os: windows-2019
             arch: aarch64
+          - os: windows-2019
```
coldav, line 39: you'll need to exclude more than that to get down to a single extra version

coldav: ah sorry I got confused

.github/workflows/create_llvm.yml
alan-forbes-cp: Confirmed for Github - thanks @hvdijk

coldav (approved):

.github/workflows/create_llvm.yml
```diff
@@ -88,10 +97,17 @@ jobs:
         run:
           sudo apt-get install --yes g++-11-aarch64-linux-gnu
 
+      - name: install x86 build tools
+        if: ${{ steps.cache.outputs.cache-hit != 'true' && matrix.arch == 'x86' }}
+        run: |
+          sudo dpkg --add-architecture i386
+          sudo apt-get update
+          sudo apt-get install --yes gcc-multilib g++-multilib libc6-dev:i386 lib32tinfo-dev
```
coldav, line 105: Your other change for setup_ubuntu installs this. So we may want to combine in some way at some point.

.github/workflows/create_llvm.yml
```diff
@@ -18,20 +18,26 @@ jobs:
         version: [17, 18]
         os: [ubuntu-22.04, windows-2019]
         build_type: [Release, RelAssert]
-        arch : [x86_64, aarch64]
+        arch : [x86, x86_64, aarch64]
         exclude:
-          # Reduce the versions of llvm for aarch64 architecture, as
+          # Reduce the versions of llvm for aarch64 & x86 architectures, as
```
coldav, line 23: windows too, not part of this change as such, but since you are changing the line.

coldav (approved):

.github/workflows/create_llvm.yml
alan-forbes-cp: Done

alan-forbes-cp:
@codeplaysoftware/ock-workflow-reviewers
Could I have this one merged please? Ta.

PR #547, opened 2024-10-02
Add ubuntu MR job install prereqs

alan-forbes-cp:
# Overview

Add prereqs for ubuntu MR jobs run to date (very likely there are more prereqs to come)

# Reason for change

Allow MR jobs to run.

# Description of change

Misc installs. Tested in my fork.

### DIFF
--- a/.github/actions/setup_ubuntu_build/action.yml
+++ b/.github/actions/setup_ubuntu_build/action.yml
@@ -24,20 +24,17 @@ runs:
   steps:
     - name: Install prerequisites
       shell: bash    
-      env:
-        UBUNTU_VER : '${{ inputs.ubuntu_version }}'
-        ARCH : '${{ inputs.arch }}'
       run: |
         pip install lit clang-format==17.0.6 virtualenv
-        if [ "$ARCH" = "x86" ]; then sudo dpkg --add-architecture i386 ; fi
+        if [ "${{ inputs.arch }}" = "x86" ]; then sudo dpkg --add-architecture i386 ; fi
         wget -qO - https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo apt-key add -
-        if [ "$UBUNTU_VER" = "20.04" ]; then sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.3.243-focal.list https://packages.lunarg.com/vulkan/1.3.243/lunarg-vulkan-1.3.243-focal.list; fi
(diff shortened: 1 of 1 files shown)


.github/actions/setup_ubuntu_build/action.yml
```diff
@@ -24,10 +24,20 @@ runs:
...
+        UBUNTU_VER : '${{ inputs.ubuntu_version }}'
+        ARCH : '${{ inputs.arch }}'
       run: |
-        sudo apt-get install -y spirv-tools
-        pip install lit
-
+        pip install lit clang-format==17.0.6 virtualenv
+        if [ "$ARCH" = "x86" ]; then sudo dpkg --add-architecture i386 ; fi
+        wget -qO - https://packages.lunarg.com/lunarg-signing-key-pub.asc | sudo apt-key add -
+        if [ "$UBUNTU_VER" = "20.04" ]; then sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.3.243-focal.list https://packages.lunarg.com/vulkan/1.3.243/lunarg-vulkan-1.3.243-focal.list; fi
```
coldav, line 34: can't we just use the ${{ inputs.ubuntu_version }}, I can't see the point in adding an env which gives no more information.

.github/actions/setup_ubuntu_build/action.yml
```diff
@@ -24,10 +24,20 @@ runs:
...
+        if [ "$UBUNTU_VER" = "22.04" ]; then sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.3.243-jammy.list https://packages.lunarg.com/vulkan/1.3.243/lunarg-vulkan-1.3.243-jammy.list; fi
+        if [ "$UBUNTU_VER" = "24.04" ]; then sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-noble.list https://packages.lunarg.com/vulkan/lunarg-vulkan-noble.list; fi
+        sudo apt-get update
+        sudo apt-get install --yes spirv-tools doxygen
```
coldav, line 38: I am pretty sure we don't need spirv-tools if we install vulkan.

alan-forbes-cp: You're right. Removed.

.github/actions/setup_ubuntu_build/action.yml
alan-forbes-cp: I think we can. I had syntax fails with an early version of this and found that there were known issues with the use of inputs in that context. Also some security hardening advice recommends using intermediate env vars rather than inputs (or any Githab vars) directly. However the issues look to be fixed and I have updated to use inputs.

coldav (approved):

PR #548, opened 2024-10-03
Enable CodeQL SAST

carlewis:
# Overview

Add a workflow to run CodeQL on main branch and pull requests. 

# Reason for change

CodeQL is a static application security testing (SAST) toolchain that allows to automate security checks and integrate them into our development workflows. 

# Description of change

Enabling CodeQL will run the scan process on main branch and pull requests. Results are accessible through the project `Security` tab. 

Workflow requires building the source code so it uses similar jobs to the ones defined in the `build_pr_cache.yml` workflow. 

# Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,99 @@
+name: CodeQL
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  schedule:
+    - cron: '19 9 * * 3'
+
+permissions: {}
+
+jobs:
+  analyze-host-x86_64-release:
+    name: Analyze host x86_64 release
+    permissions:
+      # required for all workflows
+      security-events: write
+      # required to fetch internal or private CodeQL packs
+      packages: read
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      # installs tools, ninja and installs llvm (default 17, RelAssert) and sets up cache
+      - name: setup-ubuntu
+        uses:  ./.github/actions/setup_ubuntu_build
+        with:
+          save: true
+          llvm_version: 18
+          llvm_build_type: RelAssert
+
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,99 @@
...
+  pull_request:
+    branches:
+      - main
+  schedule:
+    - cron: '19 9 * * 3'
+
+permissions: {}
+
+jobs:
+  analyze-host-x86_64-release:
```
coldav, line 16: I'm a little concerned this essentially duplicates jobs we have already. I already did that for the cache seeding and regret the overhead of maintaining it. It feels like there should be a better way e.g. same workflow with different inputs or workflows calling workflows, or conditionally setting up env variables.

carlewis: From the PoV of CodeQL we just need to build the source code between the `Initialize CodeQL` and `Perform CodeQL Analysis` jobs. I lack the context of the project so can't say with certainity, but it sounds as if we had different actions for each configuration, instead of using the same actions+parameters we may reuse those 'alias' actions. 

A second option could be having a single actions script with a configuration dictionary indexed with a descriptive key. So we always use the same script with a single parameter (the configuration dictionary key), so we can use it in a workflow matrix. 

```yaml
  strategy:
    matrix: 
      include:
      - configuration: build-host-x86_64-release
      - configuration: build-riscv-m1-release
      - configruation: ...
...
jobs:

  uses: ./.github/actions/build_and_run_ock
  with:
     configuration: ${{ matrix.configuration }}
```

Both options add complexity and need refactoring the workflows so maybe it's not worth it. One thing I tried was creating a matrix that changed the `uses: ` action, but it's not allowed.

coldav: I think we can allow this one through on the basis we will do the same thing in a different way - it sounds like your requirements are to just build it but not run it (which coincidentally is the same requirement that building the PR cache has).

I'll look into this problem.

coldav (approved):

PR #550, opened 2024-10-04
Allow setting of CPU HAL WG mode number of threads

coldav:
# Overview

Support a cmake option to set the number of threads used on cpu hal

# Reason for change

CPU HAL normally creates a number of threads for each kernel enqueue based on std::thread::hardware_concurrency. This is catastrophic on qemu where this causes memory to be continually increased. 

# Description of change

Add CA_HAL_CPU_WG_NUM_THREADS cmake variable

### DIFF
--- a/clik/external/hal_cpu/include/cpu_hal.h
+++ b/clik/external/hal_cpu/include/cpu_hal.h
@@ -48,8 +48,8 @@ class cpu_hal final : public hal::hal_device_t {
   ~cpu_hal();
 #endif
 
-  // Set up the hal device info - this is done as a static so that other classes such
-  // as hal_client can set up the desired information directly
+  // Set up the hal device info - this is done as a static so that other classes
+  // such as hal_client can set up the desired information directly
   static hal::hal_device_info_t &setup_cpu_hal_device_info();
 
   size_t get_word_size() const { return sizeof(uintptr_t); }
@@ -112,6 +112,9 @@ class cpu_hal final : public hal::hal_device_t {
   cpu_barrier barrier;
 #endif
   std::map<hal::hal_program_t, std::string> binary_files;
+
+  // Default number of threads for wg mode
+  unsigned int wg_num_threads;
 };
 
 #endif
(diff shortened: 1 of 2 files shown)


clik/external/hal_cpu/CMakeLists.txt
```diff
@@ -25,6 +25,10 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_C_FLAGS ${CMAKE_C_FLAGS} -fvisibility=hidden)
 set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} -fvisibility=hidden)
 
+# This is useful when running via qemu. Setting to 1 does not create a new thread at all
+# and help on memory.
+set(CA_HAL_CPU_WG_NUM_THREADS "" CACHE STRING "Limit number of threads in workgroup mode")
```
hvdijk, line 30: This comment doesn't quite match what it does: it can also be used to increase the number of threads to beyond `std::thread::hardware_concurrency()`. I am not sure whether a value other than empty string or "1" could be meaningful; if it couldn't be, maybe it would be better to just have a boolean option?

coldav: I thought about it and it seems the most generic, we could switch to 4 threads if for example too much memory was being used for other reasons. It's similar to CA_HOST_NUM_THREADS (although that is an environment variable). I think 1 is an artefact of using qemu here, so something generic feels better.

coldav: however I've changed the text a little to make it clear that's just one usage of it.

hvdijk: As a CMake option, setting `CA_HAL_CPU_WG_NUM_THREADS=4` seems questionable to me, since it results in a build that always uses 4 threads, even when there aren't 4 cores/vcpus/whatever you want to call them, with no way at that point to override but to rebuild OCK. As you say, `CA_HOST_NUM_THREADS` works as an environment variable, and that avoids that issue. Could we do that here too, or does that cause complications in how we want to use this in our CI (for now)?

coldav: We could go that way yes.

coldav: done

PR #552, opened 2024-10-04
add single pr job

alan-forbes-cp:
# Overview
Add an initial PR job: mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0-unitcl_vecz
Also fix an upper-case anomaly with the build action inputs strings.

# Reason for change
Allows basic PR testing - a single job with others to follow.

# Description of change
Add pull-request workflow containing single PR job.

Notable changes from Gitlab:
- build.py is retired. PR job calls an updated Github build action instead.
- setup and build actions are subject to change as other PR jobs are added.

Test results are the same as for the latest corresponding Gitlab MR job.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,0 +1,73 @@
+name: ComputeAorta/ci-github-mrexport
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+    - source/**
+    - clik/**
+    - modules/**
+    - examples/**
+    - cmake/**
+    - hal/**
+    - .github/actions/do_build_ock/**
+    - .github/actions/setup_ubuntu_build/**
+    - .github/workflows/run_pr_tests.yml
+    - CMakeLists.txt
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+jobs:
+
+############### JOB   mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0-unitcl_vecz:
+
+  mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0-unitcl_vecz:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 60
+    env:
+      EXTERNAL_MUX_COMPILER_DIRS: ${{ github.workspace }}/code/examples/refsi/refsi_m1/compiler/refsi_m1
+      MUX_COMPILERS_TO_ENABLE: riscv
+      HAL_DESCRIPTION: RV64GCV_Zfh
+      HAL_REFSI_SOC: G1
+      HAL_REFSI_THREAD_MODE: WG
+      TARGET: check-ock-UnitCL-group-vecz
+      DISABLE_VECZ_CHECKS: OFF
+      Compiler: gcc
+      CXXCompiler: g++
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v4.1.0
+      with:
+        path: code
+    - run: cp -R code/.github .github
+    - name: setup-ubuntu
+      uses: ./.github/actions/setup_ubuntu_build
+      with:
+        llvm_version: '18'
+        llvm_build_type: RelAssert
+    - run: echo WORKSPACE is $GITHUB_WORKSPACE && echo PWD is `pwd` && ls -al
+    - name: build ock
+      uses: ./.github/actions/do_build_ock
+      with:
+        path: $GITHUB_WORKSPACE/code
+        install_dir: $GITHUB_WORKSPACE/install
+        llvm_install_dir: $GITHUB_WORKSPACE/llvm_install
+        build_type: ReleaseAssert
+        build_targets: $TARGET
+        host_image: OFF
+        enable_api: cl
+        mux_targets_enable: riscv
+        external_compiler_dirs: $EXTERNAL_MUX_COMPILER_DIRS
+        mux_compilers_enable: $MUX_COMPILERS_TO_ENABLE
+        riscv_enabled: ON
+        disable_unitcl_vecz_checks: $DISABLE_VECZ_CHECKS
+        enable_rvv_scalable_vecz_check: ON
+        enable_rvv_scalable_vp_vecz_check: ON
+        command_buffer: ON
+        use_linker: gold
+        hal_description: $HAL_DESCRIPTION
+        hal_refsi_soc: $HAL_REFSI_SOC
+        hal_refsi_thread_mode: $HAL_REFSI_THREAD_MODE
+        build_dir: $GITHUB_WORKSPACE/build
+        debug_support: ON
+        gtest_launcher: "/usr/bin/python;-u;${{ github.workspace }}/code/scripts/gtest-terse-runner.py"
+        extra_flags: -DCMAKE_C_COMPILER=$Compiler -DCMAKE_CXX_COMPILER=$CXXCompiler
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -11,7 +11,6 @@ on:
     - hal/**
     - .github/actions/do_build_ock/**
     - .github/actions/setup_ubuntu_build/**
-    - .github/workflows/run_pr_tests.yml
     - .github/workflows/pull-request.yml
     - CMakeLists.txt
 concurrency:
@@ -43,7 +42,6 @@ jobs:
         install_dir: $GITHUB_WORKSPACE/install
         build_targets: check-ock-UnitCL-group-vecz
         mux_targets_enable: riscv
-        external_compiler_dirs: ${{ github.workspace }}/code/examples/refsi/refsi_m1/compiler/refsi_m1
         mux_compilers_enable: riscv
         riscv_enabled: ON
         enable_rvv_scalable_vecz_check: ON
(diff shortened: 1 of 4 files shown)


.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,58 @@
...
+    paths:
+    - source/**
+    - clik/**
+    - modules/**
+    - examples/**
+    - cmake/**
+    - hal/**
+    - .github/actions/do_build_ock/**
+    - .github/actions/setup_ubuntu_build/**
+    - .github/workflows/run_pr_tests.yml
```
coldav, line 14: This is not part of this worflow so should not be triggered.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,58 @@
...
+############### JOB   mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0-unitcl_vecz:
+
+  mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0-unitcl_vecz:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 60
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v4.1.0
+      with:
+        path: code
```
coldav, line 31: still not really sure why we need this, since we don't have it in our other jobs.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,58 @@
...
+        path: $GITHUB_WORKSPACE/code
+        install_dir: $GITHUB_WORKSPACE/install
+        build_targets: check-ock-UnitCL-group-vecz
+        mux_targets_enable: riscv
+        external_compiler_dirs: ${{ github.workspace }}/code/examples/refsi/refsi_m1/compiler/refsi_m1
+        mux_compilers_enable: riscv
+        riscv_enabled: ON
+        enable_rvv_scalable_vecz_check: ON
+        enable_rvv_scalable_vp_vecz_check: ON
+        use_linker: gold
```
coldav, line 51: if this is better, lets just make it the default.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,58 @@
...
+        llvm_build_type: RelAssert
+    - run: echo WORKSPACE is $GITHUB_WORKSPACE && echo PWD is `pwd` && ls -al
+    - name: build ock
+      uses: ./.github/actions/do_build_ock
+      with:
+        path: $GITHUB_WORKSPACE/code
+        install_dir: $GITHUB_WORKSPACE/install
+        build_targets: check-ock-UnitCL-group-vecz
+        mux_targets_enable: riscv
+        external_compiler_dirs: ${{ github.workspace }}/code/examples/refsi/refsi_m1/compiler/refsi_m1
```
coldav, line 46: This maybe allows us to remove the riscv job we already have since I think that is also m1.

alan-forbes-cp: fyi ... the single difference between the 2 jobs is
new PR: `-DCA_MUX_COMPILERS_TO_ENABLE=riscv`
existing: `-DCA_MUX_COMPILERS_TO_ENABLE=refsi_m1`

coldav: hmm, that is odd. The old one seems more correct to me.

.github/workflows/pull-request.yml
alan-forbes-cp: All of the jobs I've looked at so far do use it.
What does "better" mean? How is it measured?

coldav: I believe the gold linker is faster. Perhaps we can review the defaults separately.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,58 @@
...
+        enable_rvv_scalable_vecz_check: ON
+        enable_rvv_scalable_vp_vecz_check: ON
+        use_linker: gold
+        hal_description: RV64GCV_Zfh
+        hal_refsi_soc: G1
+        hal_refsi_thread_mode: WG
+        build_dir: $GITHUB_WORKSPACE/build
+        debug_support: ON
+        gtest_launcher: "/usr/bin/python;-u;${{ github.workspace }}/code/scripts/gtest-terse-runner.py"
+        extra_flags: -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++
```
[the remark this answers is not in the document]
alan-forbes-cp: Should we add new inputs for the 2 compiler settings then? I'm not sure if `extra_flags` is the best place for options with overridable defaults.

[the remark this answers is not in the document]
coldav: perhaps? It defaults to this anyway, so I think it's really only clang that requires the flags.

.github/workflows/pull-request.yml
alan-forbes-cp: This becomes a little knotted in that case. We've actually got 2 related Gitlab jobs
```
mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0-unitcl_vecz
mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0
```
... both with provisional migrations to Github. (first one under this PR). And we've got the existing Github riscv job mentioned above.
I don't have a clear view on which, if any, are correct or how to resolve them otherwise.
Any suggestions  - with a view to unblocking this PR specifically?

coldav: I think if it says g1 it shouldn't point to m1 compiler, so just drop that.
We can look at the others as they come in.

.github/workflows/pull-request.yml
alan-forbes-cp: I'll add that to the "To Do"list in the ticket and resolve this conversation.

.github/workflows/pull-request.yml
alan-forbes-cp: As Gitlab uses `build.py` the model used there doesn't migrate unchanged. Its not shown in this job, but some Gitlab jobs also create build dirs outwith the workspace with `build.py` jumping in and out of the workspace and using command line settings to point it back to "source" to perform parts of the runs. 
I tried to simplify that so Github jobs do all their work in the workspace with "top-level" dirs in there for various purposes to stop things becoming cluttered: i.e. code, build, install, llvm, any "extra dirs" the build needs, and probably other things I haven't met yet. It also helps to work out how all the pieces fit together during job development.
However ... much of it comes down to personal preference of course so I have no objections if you want to suggest an alternative scheme.

coldav: I don't have strong opinions, it just feels like a departure from what we have and calling it code - just means oneapi-construction-kit will be checked out as code - is that really clearer?

alan-forbes-cp: There's no build.py now so it will be different. I used "code" as that's the name Github uses to refer to repo content but we can rename that if required. I think overall its clearer as the structure is more consistent between jobs and everything is under the workspace (although admittedly I've not looked at all jobs as yet).

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,55 @@
...
+############### JOB   mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0-unitcl_vecz:
+
+  mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0-unitcl_vecz:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 60
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v4.1.0
+      with:
+        path: code
```
coldav, line 30: I still think I'd need convincing on this one.

coldav (approved):

.github/workflows/pull-request.yml
alan-forbes-cp: Removed

.github/workflows/pull-request.yml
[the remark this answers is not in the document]
alan-forbes-cp: Inputs added for clang

.github/workflows/pull-request.yml
alan-forbes-cp: m1 compiler dropped

.github/workflows/pull_request.yml
alan-forbes-cp: Sure. This PR's job doesn't show all the issues that well. [Here's another example in Gitlab](https://git.office.codeplay.com/ComputeAorta/ci-github/-/blob/main/.gitlab/ci/merge-request.yml?ref_type=heads#L496)
... note the `--source_dir` option to `build.py`.
I guess we're looking for a consistent structure that works across all jobs when `build.py` is removed from each. And possibly something that tidies up from the Gitlab offering too as it can be tricky to follow. Any improvement on my offering is fine by me.

.github/workflows/pull-request.yml
alan-forbes-cp: Closing in favour of a later conversation.

.github/workflows/pull_request.yml
coldav: that still doesn't really convince me, the checkout to code and copy of .github feels a bit clumsy to me. I don't see this approach on other github workflows elsewhere. I'd welcome @hvdijk's opinion though.

alan-forbes-cp: As discussed off-line - I will refactor to restore the Gitlab model.

alan-forbes-cp: Done

PR #559, opened 2024-10-11
Demonstrate callable workflow methodology

coldav:
# Overview

Show how we can call the same workflow in different ways

# Reason for change

Allow flexibility in how we run tests

# Description of change

Added a calling workflow and made the main pr tests required to be called.


.github/actions/setup_build/action.yml
```diff
@@ -32,6 +32,12 @@ runs:
       if: ${{ inputs.os == 'ubuntu' }}
       shell: bash
       run: |
+        echo llvm_build_type "${{ inputs.llvm_build_type }}"
+        echo llvm_version "${{ inputs.llvm_version }}"
+        echo ubuntu_version "${{ inputs.ubuntu_version }}"
+        echo llvm_build_type "${{ inputs.llvm_build_type }}"
```
alan-forbes-cp, line 38: Duplicate of line 35: **llvm_build_type**

.github/workflows/run_pr_tests.yml
```diff
@@ -163,3 +164,33 @@ jobs:
           build_dir: ${{ github.workspace }}/build_offline
           c_compiler: 'cl.exe'
           cxx_compiler: 'cl.exe'
+
+############### JOB   mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0-unitcl_vecz:
```
alan-forbes-cp, line 168: Delete "JOB" comment maybe?

.github/workflows/run_pr_tests.yml
```diff
@@ -163,3 +164,33 @@ jobs:
...
+
+  mr-ubuntu-gcc-x86_64-riscv-fp16-cl3-0-unitcl_vecz:
```
alan-forbes-cp, line 170: Do we want to drop the "mr" as the job name prefix and replace with "run"?
Also other jobs use underscores as separators in the name - not fussy about that but I guess we should be consistent.

.github/workflows/run_pr_tests_caller.yml
```diff
@@ -0,0 +1,35 @@
...
+      - 'modules/**'
+      - 'examples/**'
+      - 'cmake/**'
+      - 'hal/**'
+      - '.github/actions/do_build_ock/**'
+      - '.github/actions/setup_build/**'
+      - '.github/workflows/run_pr_tests.yml'
+      - 'CMakeLists.txt'
+  schedule:
+    - cron: '00 19 * * 1-5'
```
alan-forbes-cp, line 17: Maybe a comment to explain the cron settings?

.github/workflows/run_pr_tests.yml
coldav: This is just moving your job across, i think it helps to cross-correlate for now.

alan-forbes-cp (approved):

PR #560, opened 2024-10-11
Add first x86 PR job

alan-forbes-cp:
# Overview

and consequences.*

# Reason for change

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -47,3 +47,45 @@ jobs:
         hal_refsi_thread_mode: WG
         debug_support: ON
         gtest_launcher: "/usr/bin/python;-u;${{ github.workspace }}/scripts/gtest-terse-runner.py"
+
+############### JOB   mr-ubuntu-clang-x86-llvm-latest-cl3-0-offline:
+
+  mr-ubuntu-clang-x86-llvm-latest-cl3-0-offline:
+    runs-on: ubuntu-22.04
+    timeout-minutes: 90
+    steps:
+    - name: Checkout repo
+      uses: actions/checkout@v4.1.0
+    - name: setup-ubuntu
+      uses: ./.github/actions/setup_ubuntu_build
+      with:
+        llvm_version: '18'
+        llvm_build_type: RelAssert
+        arch: x86
+    - run: echo WORKSPACE is $GITHUB_WORKSPACE && echo PWD is `pwd` && ls -al
+    - name: build ock one
+      uses: ./.github/actions/do_build_ock
+      with:
+        arch: x86
+        enable_api: ""
+        c_compiler: $GITHUB_WORKSPACE/llvm_install/bin/clang
+        cxx_compiler: $GITHUB_WORKSPACE/llvm_install/bin/clang++
+        build_targets: check-ock
+        builtin_kernel: ON
+        use_linker: gold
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -47,3 +47,45 @@ jobs:
...
+    - name: Checkout repo
+      uses: actions/checkout@v4.1.0
+    - name: setup-ubuntu
+      uses: ./.github/actions/setup_ubuntu_build
+      with:
+        llvm_version: '18'
+        llvm_build_type: RelAssert
+        arch: x86
+    - run: echo WORKSPACE is $GITHUB_WORKSPACE && echo PWD is `pwd` && ls -al
+    - name: build ock one
```
coldav, line 66: fix the names please, they are clearly doing different things. (Note I have an online/offline example already).

.github/workflows/pull_request.yml
```diff
@@ -47,3 +47,45 @@ jobs:
...
+      uses: ./.github/actions/do_build_ock
+      with:
+        arch: x86
+        enable_api: ""
```
coldav, line 70: I'd definitely like to review these that are likely to be defaults, it feels like this is one of them (linker etc is another).

alan-forbes-cp: I think you're right on this one. I was assuming changes needed to legacy defaults and any others would become clearer as more jobs were added and we reach "critical mass" in terms of job numbers. We could review in advance if that makes more sense. For now I've updated the `enable_api` default.

.github/workflows/pull_request.yml
```diff
@@ -47,3 +47,45 @@ jobs:
...
+        arch: x86
+    - run: echo WORKSPACE is $GITHUB_WORKSPACE && echo PWD is `pwd` && ls -al
+    - name: build ock x86 relassert
+      uses: ./.github/actions/do_build_ock
+      with:
+        arch: x86
+        c_compiler: $GITHUB_WORKSPACE/llvm_install/bin/clang
+        cxx_compiler: $GITHUB_WORKSPACE/llvm_install/bin/clang++
+        build_targets: check-ock
+        enable_api: ""
```
coldav, line 73: again would like to see some of these "defaults" disappear.

.github/workflows/run_pr_tests.yml
```diff
@@ -194,3 +194,43 @@ jobs:
...
+        arch: x86
+    - run: echo WORKSPACE is $GITHUB_WORKSPACE && echo PWD is `pwd` && ls -al
+    - name: build ock x86 relassert
+      uses: ./.github/actions/do_build_ock
+      with:
+        build_32_bit: ON
+        c_compiler: $GITHUB_WORKSPACE/llvm_install/bin/clang
+        cxx_compiler: $GITHUB_WORKSPACE/llvm_install/bin/clang++
+        build_targets: check-ock
+        enable_api: ""
```
alan-forbes-cp, line 219: @coldav Do you want this to be the default across the board? 
Currently about 67% of the do_build_ock calls use "cl".

coldav: keep the old behaviour for the old tests, I think they will get dropped in time as you move things across anyway.

coldav (approved):

PR #567, opened 2024-10-16
Add support for ConstantAray in replace-module-scope pass.

coldav:
# Overview

Add support for ConstantAray in replace-module-scope pass.

# Reason for change

Native CPU for aarch64 and risc-v was outputting constant arrays which pointed to local variables.
This was not handled by the replace-module-scope pass.

# Description of change

Fixed by iterating through the ConstantArray and turning each element into an instruction in replaceConstantExpressionWithInstruction().

Added lit test


modules/compiler/compiler_pipeline/source/pass_functions.cpp
```diff
@@ -152,10 +152,10 @@ void replaceConstantExpressionWithInstruction(llvm::Constant *const constant) {
   // remove all dead constant users (sometimes these are left over by previous
   // passes)
   constant->removeDeadConstantUsers();
-
   // Only handle constants which are ConstantExpr or ConstantVector
```
PietroGhg, line 155: We should also update this comment

PietroGhg (approved):

PR #579, opened 2024-11-01
Enable vfp4 for ARM.

hvdijk:
# Overview

Enable vfp4 for ARM.

# Reason for change

We made the decision at the start of the year to only support targets that have hardware FMA support. This is available on 32-bit ARM, but only as of VFPv4.

# Description of change

Make sure to enable this.

# Checklist

### DIFF
--- a/modules/compiler/targets/host/source/target.cpp
+++ b/modules/compiler/targets/host/source/target.cpp
@@ -256,14 +256,13 @@ compiler::Result HostTarget::initWithBuiltins(
       // for single precision floating points.
       Features.AddFeature("strict-align", true);
       Features.AddFeature("neonfp", true);
-      Features.AddFeature("neon", true);
+      // We need hardware FMA support which is only available as of VFP4.
+      // VFP4 also includes FP16.
+      Features.AddFeature("vfp4", true);
       // Hardware division instructions might not exist on all ARMv7 CPUs, but
       // they probably exist on all the ones we might care about.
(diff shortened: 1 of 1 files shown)


modules/compiler/targets/host/source/target.cpp
```diff
@@ -256,14 +256,13 @@ compiler::Result HostTarget::initWithBuiltins(
       // for single precision floating points.
       Features.AddFeature("strict-align", true);
       Features.AddFeature("neonfp", true);
-      Features.AddFeature("neon", true);
+      // We need hardware FMA support which is only available as of VFP4.
+      // VFP4 also includes FP16.
+      Features.AddFeature("vfp4", true);
```
coldav, line 261: Does this force us to only run on aarch64 which supports vfp4? Or is that the norm?

hvdijk: This is not only AArch64, no, VFP4 is also available on several 32-bit ARMv7 CPUs.

coldav: what I mean - is this something guaranteed on aarch64? If so, I'm more than happy to approve.

hvdijk: Ah, sorry, misunderstood. Yes, my understanding is that this is always available on ARMv8+ which is the minimum that supports 64-bit.

coldav (approved):

PR #582, opened 2024-11-07
[NFC] Avoid #if in intrinsic.h

hvdijk:
# Overview

[NFC] Avoid #if in intrinsic.h

# Reason for change

Current multi_llvm::GetOrInsertIntrinsicDeclaration works on LLVM 18, 19, and main, but not on DPC++ since it is based on an older version of LLVM main. For NativeCPU, it would help us to not need a separate branch.

# Description of change

Re-work multi_llvm::GetOrInsertIntrinsicDeclaration to work without a hardcoded check for LLVM version, change it so that it detects the available functions within the llvm::Intrinsic namespace instead.

# Checklist

### DIFF
--- a/modules/compiler/multi_llvm/include/multi_llvm/intrinsic.h
+++ b/modules/compiler/multi_llvm/include/multi_llvm/intrinsic.h
@@ -19,15 +19,46 @@
 
 #include <multi_llvm/llvm_version.h>
 
+namespace llvm {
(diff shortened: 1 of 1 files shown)


hvdijk:
This will not be needed as DPC++ is working on another LLVM pulldown, which is recent enough to no longer need a different version of OCK.

PR #588, opened 2024-11-14
Link against static versions of lld.

hvdijk:
# Overview

Link against static versions of lld.

# Reason for change

When LLVM is built with LLVM_LINK_LLVM_DYLIB=ON, in modules/compiler/source/base/CMakeLists.txt, we make sure to override that and forcibly link against the static versions of LLVM and Clang. In modules/compiler/utils/CMakeLists.txt, we need to do the same for LLD.

# Description of change

points, etc.*

# Checklist


modules/compiler/utils/CMakeLists.txt
```diff
@@ -53,7 +53,30 @@ if(EXISTS "${CA_LLVM_INSTALL_DIR}/lib/cmake/lld/LLDConfig.cmake")
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
   )
 
+  # Explicitly link against the individual static versions of the lld
```
coldav, line 56: Before we checked if LLD was enabled - should we do that here?

coldav (approved):

modules/compiler/utils/CMakeLists.txt
hvdijk: We checked by `if(EXISTS "${CA_LLVM_INSTALL_DIR}/lib/cmake/lld/LLDConfig.cmake")`. This new logic I added is inside that `if`, so that should be handled already, this should do nothing when LLD is disabled.

coldav (approved):

PR #592, opened 2024-11-19
Install clang-tidy-19 rather than use pip install

coldav:
This resolves this by using the apt-get install method.

# Overview

Switch to using apt-get install rather than pip for clang-tidy-19

# Reason for change

pip installing clang-tidy-19 on aarch64 was causing a full rebuild rather than just fetching.


.github/actions/setup_build/action.yml
```diff
@@ -38,8 +38,15 @@ runs:
         if [ "${{ inputs.ubuntu_version }}" = "20.04" ]; then sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.3.243-focal.list https://packages.lunarg.com/vulkan/1.3.243/lunarg-vulkan-1.3.243-focal.list; fi
         if [ "${{ inputs.ubuntu_version }}" = "22.04" ]; then sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.3.243-jammy.list https://packages.lunarg.com/vulkan/1.3.243/lunarg-vulkan-1.3.243-jammy.list; fi
         if [ "${{ inputs.ubuntu_version }}" = "24.04" ]; then sudo wget -qO /etc/apt/sources.list.d/lunarg-vulkan-noble.list https://packages.lunarg.com/vulkan/lunarg-vulkan-noble.list; fi
+
+        sudo wget -qO - https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor - | tee /usr/share/keyrings/llvm-archive-keyring.gpg >/dev/null
```
hvdijk, line 42: This doesn't look like it can work, it's `tee` that needs to be run as root, not `wget`.

.github/actions/setup_build/action.yml
```diff
@@ -38,8 +38,15 @@ runs:
...
+        if [ "${{ inputs.ubuntu_version }}" = "20.04" ]; then (sudo echo 'deb [signed-by=/usr/share/keyrings/llvm-archive-keyring.gpg] http://apt.llvm.org/focal/ llvm-toolchain-focal-19 main' | sudo tee /etc/apt/sources.list.d/llvm.list >/dev/null); fi
```
hvdijk, line 43: The parentheses here are confusing, they are not just for grouping like in other languages, they are specifically for creating subshells (shell environments that operate in an isolated fashion, so that e.g. variable assignments in the subshell do not last beyond it), but you don't do anything that makes use of the subshell. Is there a reason that I am missing?

coldav: it was to keep the `|` inside the if

hvdijk: You don't need parentheses for that, `if a; then b | c; fi` is valid.

.github/actions/setup_build/action.yml
hvdijk: If it works, then it means the user already has write permission to `/usr/share/keyrings/llvm-archive-keyring.gpg` even without `sudo` for some reason. In that case you would be able to take out the `sudo` entirely and you would see the same result.

hvdijk (approved):

PR #611, opened 2024-12-13
Add opencl cts aarch64 linux to planned testing

alan-forbes-cp:
# Overview

Add opencl cts aarch64 linux to planned testing

# Reason for change

Requirement to add OpenCl CTS testing for aarch64 (and, by separate ticket, riscv64) 

# Description of change

aarch64 added here - riscv64 to follow.

* `fail-fast: false` has been added to certain matrices to allow all jobs at the same level to run to completion should any individual job fail. 
* New true/false flag added for vulkan SDK install (will also be used by forthcoming riscv64 testing)
* Corrected riscv64 toolchain filename.
* Added a few `set -x` bash commands to improve logging.
* Pull requests are visible in `planned_testing_caller` for dev test purposes only and will be removed.
* Current test run results for aarch64 match those of x86_64 - i.e. one known fail - and similar jobs on Gitlab.


.github/actions/do_build_opencl_cts/action.yml
```diff
@@ -35,13 +35,26 @@ runs:
         repository: KhronosGroup/OpenCL-CTS
         path: OpenCL-CTS
 
-    - name: build opencl cts
+    - name: aarch64 set-up
```
coldav, line 38: technically this is only for when we use qemu rather than the machine, so we'll have to review this at that point.

.github/actions/do_build_opencl_cts/action.yml
```diff
@@ -35,13 +35,26 @@ runs:
...
       shell: bash    
+      run: |
+        set -x
+        sudo sed -i -e '/^deb /{h;s|deb |&[arch=amd64,i386] |p;g;s|deb http://[^ ]*|deb [arch=arm64,riscv64] http://ports.ubuntu.com/ubuntu-ports|p;d}' /etc/apt/sources.list
+        sudo dpkg --add-architecture arm64
+        sudo apt-get install --yes gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
+        sudo apt-get update
+        if [[ "${{inputs.target}}" =~ .*aarch64.* ]] ; then
+          echo QEMU INSTALL MATCH FOR ${{inputs.target}}
+          sudo apt-get install --yes qemu-user
```
coldav, line 49: This is only needed for running it, so should ideally be in that section (we've talked about it being in a different workflow in the future).
I can live with a TODO: for now to avoid a complete rerun though

.github/actions/run_opencl_cts/action.yml
```diff
@@ -23,19 +23,26 @@ runs:
...
+          --color=always --timeout $CTS_TIMEOUT \
+          $QEMU_SETTING \
+          -b $GITHUB_WORKSPACE/test_conformance \
+          -L $GITHUB_WORKSPACE/install_icd/lib \
+          -e CLC_EXECUTABLE=$GITHUB_WORKSPACE/install_ock/bin/clc \
+          -e OCL_ICD_FILENAMES=$GITHUB_WORKSPACE/install_ock/lib/libCL.so \
+          -e CL_PLATFORM_INDEX=0 \
+          -s $GITHUB_WORKSPACE/test_conformance/$CTS_CSV_FILE \
+          -i $GITHUB_WORKSPACE/source/cl/scripts/$CTS_FILTER"
+        eval $RUN_CITIES
```
coldav, line 48: Maybe a comment to explain this workaround?

alan-forbes-cp: Done

.github/actions/do_build_opencl_cts/action.yml
alan-forbes-cp: TODO added - I will also review as part of the riscv64 work.

.github/actions/do_build_opencl_cts/action.yml
alan-forbes-cp: Understood - comment added.

PR #615, opened 2024-12-19
Use GitHub Discussions instead of Slack.

hvdijk:
# Overview

Use GitHub Discussions instead of Slack.

# Reason for change

#612 updated the documentation to refer to a new Slack channel, but after some more discussion, that turns out to not be what we want to be using since we have GitHub Discussions already.

Additionally, the CONTRIBUTING page is not where I would expect to find the instructions on how to ask questions.

# Description of change

- Change references to Slack to reference GitHub Discussions instead
- Move support section of CONTRIBUTING to README

# Checklist


uwedolinsky (approved):
👍

PR #621, opened 2025-01-07
Use cp-ubuntu-24.04 with ubuntu 22.04 image on slow PR job

coldav:
# Overview

Use cp-ubuntu-24.04 with ubuntu 22.04 image on slow PR job

# Reason for change

Desire to speed up jobs and provide a pathway to use these runners.

# Description of change

Uses cp-ubuntu-24.04, but with ubuntu 22.04 image from intel/llvm. Required a few small additional installs/env vars etc to make it work.

### DIFF
--- a/.github/actions/setup_build/action.yml
+++ b/.github/actions/setup_build/action.yml
@@ -36,6 +36,11 @@ runs:
       if: ${{ inputs.os == 'ubuntu' }}
       shell: bash
       run: |
+        # required for virtualenv running in docker
+        echo "PATH=$PATH:$HOME/.local/bin" >> $GITHUB_ENV
+        # required due to using non docker to build llvm
+        # If we switch to always using docker we can drop.
+        sudo apt-get install --yes lib32ncurses-dev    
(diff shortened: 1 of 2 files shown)


.github/actions/setup_build/action.yml
```diff
@@ -52,6 +57,8 @@ runs:
         sudo apt-get update
         if [ "${{ inputs.clang_tidy }}" = "true" ]; then sudo apt-get install --yes clang-tidy-19; fi
         pip install lit clang-format==19.1.0 virtualenv
+        # required for gtest-terse-runner running in docker
+        pip install colorama
```
hvdijk, line 61: We shouldn't be using `pip install` for anything, it will interfere when we move to Ubuntu 24. It's not necessary to change what we have already, but for new things, can we do `apt-get install --yes python3-colorama` here?

coldav: sure

hvdijk (approved):

PR #623, opened 2025-01-08
Switch to ccache

coldav:
sccache requires root access which fails when we use a docker, so move to the simpler ccache which uses apt install.

# Overview

move to the simpler ccache which uses apt install.

# Reason for change

sccache requires root access which fails when we use a docker

# Description of change

Replace everywhere that mentions sccache with cccahe.


alan-forbes-cp (approved):

.github/actions/setup_build/action.yml
```diff
@@ -121,10 +121,11 @@ runs:
...
+    - name: Setup ccache
       uses: hendrikmuhs/ccache-action@ed74d11c0b343532753ecead8a951bb09bb34bc9 # v1.2.14
       with:
         max-size: 200M
-        key: sccache-build
-        variant: sccache
+        key: ccache-build
+        variant: ccache
         save: ${{ inputs.save }}
+
```
hvdijk, line 131: I assume the blank line here is accidental?

hvdijk:
The change looks fine to me (aside from the one nit pick), but I don't understand the description, "sccache requires root access which fails when we use a docker, so move to the simpler ccache which uses apt install." apt install also requires root access, and root access inside the Docker should be available. So what is it about sccache that doesn't work?

.github/actions/setup_build/action.yml
coldav: yes, I'll remove

coldav:
> The change looks fine to me (aside from the one nit pick), but I don't understand the description, "sccache requires root access which fails when we use a docker, so move to the simpler ccache which uses apt install." apt install also requires root access, and root access inside the Docker should be available. So what is it about sccache that doesn't work?

ccache uses sudo, which does work, sccache seems to assume permissions.

PR #624, opened 2025-01-08
Add riscv64 support to create_llvm

alan-forbes-cp:
# Overview

Add riscv64 support to create_llvm

# Reason for change

Part of OpenCL CTS work for aarch64 and riscv64

# Description of change

Added riscv64 support on the same basis as aarch64

Version numbers for build tool install (aarch64 & riscv64) have been removed.

### DIFF
--- a/.github/workflows/create_llvm.yml
+++ b/.github/workflows/create_llvm.yml
@@ -23,13 +23,17 @@ jobs:
         version: [18, 19]
         os: [ubuntu-22.04, windows-2019]
         build_type: [Release, RelAssert]
-        arch : [x86, x86_64, aarch64]
+        arch : [x86, x86_64, aarch64, riscv64]
         exclude:
-          # Reduce the versions of llvm for aarch64 & x86 architectures and windows, as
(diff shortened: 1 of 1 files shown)


.github/workflows/create_llvm.yml
```diff
@@ -106,6 +115,11 @@ jobs:
         run:
           sudo apt-get install --yes g++-11-aarch64-linux-gnu
 
+      - name: install riscv64 build tools
+        if: ${{ steps.cache.outputs.cache-hit != 'true' && matrix.arch == 'riscv64' }}
+        run:
+          sudo apt-get install --yes g++-11-riscv64-linux-gnu
```
hvdijk, line 121: This should be `g++-riscv64-linux-gnu` and likewise for AArch64. In !619 I changed that for `.github/actions/setup_build/action.yml` but I see I didn't update it everywhere. Can you update that here? Alternatively, a separate PR to update that for AArch64 first would also be fine if you prefer.

alan-forbes-cp: Good spot - thanks. I have updated both and checked there are no other similar usages.

hvdijk (approved):

PR #627, opened 2025-01-09
Switch over planned testing to use cp-ubuntu-24.04 for ock/OpenCL

coldav:
# Overview

Switch over planned testing to use cp-ubuntu-24.04 for ock/OpenCL

# Reason for change

Speed up overnight builds

# Description of change

Switched ock artefact creation and OpenCL build/run to use cp-ubuntu-24.04 (with docker).

### DIFF
--- a/.github/workflows/planned_testing.yml
+++ b/.github/workflows/planned_testing.yml
@@ -51,7 +51,27 @@ jobs:
         cat $GITHUB_OUTPUT    
 
 
-  create_ock_artefacts:
+  create_ock_artefacts_windows:
+    needs: [workflow_vars]
+    # Currently there is only one windows target so we don't bother with the
+    # matrix, just check if it's enabled
+    runs-on: windows-2019
+    if : inputs.ock && contains(inputs.target_list, 'host_x86_64_windows')
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/planned_testing.yml
```diff
@@ -51,18 +51,41 @@ jobs:
...
+  create_ock_artefacts_windows:
+    needs: [workflow_vars]
+    # Currently there is only one windows target so we don't bother with the
+    # matrix, just check if it's enabled
+    runs-on: windows-2019
+    if : inputs.ock && contains(inputs.target_list, 'host_x86_64_windows')
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+      # Todo: move to a common place e.g. setup_build
```
alan-forbes-cp, line 63: Why defer the move?

alan-forbes-cp (approved):

.github/workflows/planned_testing.yml
coldav: I didn't want to through the pain for rerunning the tests at this point and delaying things (also  may be a better way). I will do it in the future

PR #629, opened 2025-01-09
Reuse previous workflow run's artefacts

coldav:
# Overview

Support a method of using a previous workflow runs's artefact

# Reason for change

Sometimes we want to fix something far down the chain running sycl cts and don't want to run all the way through.

# Description of change

Add inputs to planned_testing:
   download_ock_artefact, download download_dpccp_artefact and download_sycl_cts_artefact which can allow downloading an artefact from a previous workflow run for a particular target or set of targets.

### DIFF
--- a/.github/actions/setup_gh/action.yml
+++ b/.github/actions/setup_gh/action.yml
@@ -1,0 +1,17 @@
+name: setup gh
+description: Set up github's cli gh
+
+inputs:
+  os:
+    description: 'os to use (contains windows or ubuntu)'
+    default: 'ubuntu'
+  token:
+    description: 'token for gh'
+
+runs:
+  using: "composite"
+  steps:
+      - name: set up secret ubuntu
+        shell: bash
+        run:
+          echo "GH_TOKEN=${{ inputs.token }}" >> $GITHUB_ENV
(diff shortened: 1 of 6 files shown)


.github/workflows/planned_testing_caller.yml
```diff
@@ -30,4 +30,9 @@ jobs:
       test_opencl_cts: true
       # Have a pull request setting which can be used to test the flow as best as possible
       # in a reasonable time
+
```
alan-forbes-cp, line 33: `download_opencl_artefact:` too? - if the intention was completeness rather than examples

.github/actions/setup_gh/action.yml
```diff
@@ -0,0 +1,22 @@
...
+  steps:
+      - name: set up secret ubuntu
+        shell: bash
+        run:
+          echo "GH_TOKEN=${{ inputs.token }}" >> $GITHUB_ENV
+      - name: install gh
+        if: contains(inputs.os, 'ubuntu')
+        shell: bash
+        run:
+         sudo apt-get install --yes gh
```
alan-forbes-cp, line 22: gh now in the dockers as discussed (I'll merge first thing tomorrow)

.github/workflows/planned_testing_caller.yml
coldav: actually that's something that doesn't really exist, it didn't seem worth it time wise, but I can see it still exists in some form but not used.

PR #635, opened 2025-01-10
[compiler] Do not mix vector and scalar kernels.

hvdijk:
# Overview

[compiler] Do not mix vector and scalar kernels.

# Reason for change

Per OpenCL 3.0 API 3.2.1 Mapping Work-items Onto an Nd-range, all sub-groups within a work-group will be the same size, apart from the sub-group with the maximum index which may be smaller if the size of the work-group is not evenly divisible by the size of the sub-groups. We were not meeting this requirement: in cases where we would not or could not generate a predicated vectorized kernel, we would execute the scalar kernel in a loop for any remaining work items, possibly resulting in multiple sub-groups that are smaller than the maximum sub-group size.

# Description of change

To avoid this situation, we need to avoid mixing vector and scalar kernels. If we can handle all work items with vector kernels, possibly with predication, continue to do so, but if we cannot, we need to switch to the scalar kernel for all work items.

This includes one bugfix to ControlFlowConversionPass to fix a crash seen now, where we use the result of createMasked{Load,Store} before checking whether it succeeded.

There is a risk of performance degradations, but only in cases where we are able to generate a vectorized kernel, but unable to generate a predicated vectorized kernel. If this turns out to be a problem, we can work on improving the predication.

# Checklist


coldav (approved):
Approve in principal if we can identify any performance regressions.

PR #645, opened 2025-01-16
Fix clang-tidy job

coldav:
# Overview

Fix clang-tidy job

# Reason for change

This was both incorrect and did not work with the new dockers.

# Description of change

This replicates our original code in our previous internal CI.  We cannot reliably get dependencies of clang-tidy targets by grepping for add_dependencies. Luckily, we do not need to: we can simply look for all targets ending in '.cpp' or '.h' instead.


alan-forbes-cp (approved):
I will remove `parallel` install when new dockers are ready

PR #646, opened 2025-01-16
[mux] Remove degenerate subgroup support.

hvdijk:
# Overview

[mux] Remove degenerate subgroup support.

# Reason for change

Older versions of oneAPI Construction Kit used an implementation of subgroups where the subgroup size was always equal to the workgroup size. This implementation is no longer used by any targets, but the code to support it was still in place. This commit removes it.

# Description of change

points, etc.*

# Checklist


modules/mux/targets/host/source/kernel.cpp
```diff
@@ -64,7 +64,7 @@ kernel_s::kernel_s(mux_device_t device, mux::allocator allocator,
   this->device = device;
   this->local_memory_size = 0;
   auto err = variant_data.push_back(kernel_variant_s{
-      std::string(kern_name, name_length), hook, 0u, 1u, 1u, 0u});
+      std::string(kern_name, name_length), hook, 0u, 1u, 1u, 1u});
```
coldav, line 67: I understand the last parameter is the subgroup size, but can you explain this change please?

modules/mux/targets/riscv/source/kernel.cpp
```diff
@@ -105,16 +105,11 @@ mux_result_t kernel_s::getSubGroupSizeForLocalSize(size_t local_size_x,
...
-  // work-group size.
-  if (variant.sub_group_size == 0) {
-    *out_sub_group_size = local_size_x * local_size_y * local_size_z;
-  } else {
-    // Otherwise, on risc-v we always use vectorize in the x-dimension, so
-    // sub-groups "go" in the x-dimension.
-    *out_sub_group_size =
-        std::min(local_size_x, static_cast<size_t>(variant.sub_group_size));
-  }
+  // On risc-v we always use vectorize in the x-dimension, so sub-groups "go" in
```
coldav, line 108:
```suggestion
  // On risc-v we always vectorize in the x-dimension, so sub-groups "go" in
```

modules/mux/targets/host/source/kernel.cpp
hvdijk: `0u` was a special subgroup size that marked the kernel as using degenerate subgroups. This is no longer valid and now results in division by zero. None of the affected kernels (the built-in kernels) use subgroups so it is okay to mark them as having a subgroup size of 1.

modules/mux/targets/riscv/source/kernel.cpp
hvdijk: Done (along with a rewrap according to `clang-format` rules)

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # ComputeAorta Changes
```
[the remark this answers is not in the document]
coldav: I think whilst you are changing this file, it would be good to fix it up (since it's more public facing). All the cmake stuff is more fiddly.

coldav (approved):

CHANGELOG.md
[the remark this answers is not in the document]
hvdijk: Alright, done.

PR #648, opened 2025-01-17
DRAFT: Reference removal of Vulkan SDK dependency

coldav:
# Overview

and consequences.*

# Reason for change

Demonstrate what would need to happen for RFC for Vulkan SDK removal

# Description of change

points, etc.*

# Checklist


hvdijk:
> This also increases the version to 5.0.0.

Why is this part of this PR?

coldav:
> > This also increases the version to 5.0.0.
> 
> Why is this part of this PR?

I thought as it was removing an API, this counted as a breaking change and we had to up the version. If you disagree we can remove it and do that later.

doc/cmake.rst
```diff
@@ -86,20 +85,19 @@ hardware is unavailable.
 Shared Library Naming and Versioning
 ####################################
 
-Both OpenCL and Vulkan APIs are designed to work through an ICD loader, a
-shared library named ``OpenCL`` and ``Vulkan`` respectively in the form
+Both OpenCL APIs are designed to work through an ICD loader, a
```
hvdijk, line 88: This sentence is no longer right, "Both" referred to "OpenCL and Vulkan APIs", there is no "Both OpenCL APIs".

doc/cmake.rst
```diff
@@ -86,20 +85,19 @@ hardware is unavailable.
...
 the same system.
 
-To avoid naming collisions with the ICD our OpenCL and Vulkan builds by default
-are named ``CL`` and ``VK``, however these can be overridden via the
-``CA_CL_LIBRARY_NAME`` and ``CA_VK_LIBRARY_NAME`` options. Providing customer
-teams with the flexibility to deliver on embedded systems without an ICD loader,
-where only our platform will exist and can be linked to directly by the
-application.
+To avoid naming collisions with the ICD our OpenCL builds by default are named
+``CL``, however these can be overridden via the ``CA_CL_LIBRARY_NAME`` option.
```
hvdijk, line 97: these -> this

doc/cmake.rst
```diff
@@ -86,20 +85,19 @@ hardware is unavailable.
...
+Providing customer teams with the flexibility to deliver on embedded systems
```
hvdijk, line 98: Pre-existing problem but as long as we're making changes here: this is not a complete sentence. This works if `. Providing` is changed to `, providing`.

doc/developer-guide.md
```diff
@@ -253,10 +250,10 @@ The builtin CMake options used when invoking CMake on the command line.
   relevant llvm headers and support tools, and their version must match
   a supported LLVM version.
 
-* `CA_ENABLE_API`: Semi-colon separated list of APIs to enable. Valid values
-  are `cl` for OpenCL, and `vk` for Vulkan. Enabling an API when an optional
-  component is not present dependent on license agreement will result in a CMake
-  error. The default is `cl;vk`.
+* `CA_ENABLE_API`: Semi-colon separated list of APIs to enable. Valid values are
+  `cl` for OpenCL. Enabling an API when an optional component is not present
+  dependent on license agreement will result in a CMake error. The default is
```
hvdijk, line 255: Another pre-existing problem, there is no license agreement that applies to this option.

Given that we're changing this to make `CA_ENABLE_API` to have two valid values, `""` and `"cl"`, both of which do exactly the same thing, do we want to just get rid of the `CA_ENABLE_API` option entirely?

doc/getting-started.rst
```diff
@@ -168,53 +168,20 @@ Windows 11
...
 Package Manager Installation
 ............................
 
-Users of Ubuntu 20.04 and later should configure their apt package manager to
-install the Vulkan SDK packages directly as described by the `LunarG Vulkan SDK
-packages`_ page. This installation method does neither set nor require the
-:envvar:`VULKAN_SDK` environment variable, avoiding the configuration
-complexities experienced when installing manually.
-
+SPIRV-Tools is needed to build the OneAPI Construction Kit. This can either come from Vulkan
```
hvdijk, line 174: OneAPI -> oneAPI. This doesn't look like it should be under a "Package Manager Installation" heading?

doc/overview/introduction/architecture.rst
```diff
@@ -176,7 +176,7 @@ Clang
   from the OpenCL API.
 SPIR-V Parser
   An in house SPIR-V to LLVM IR translator for consuming SPIR-V from the OpenCL
-  and Vulkan APIs.
+   API.
```
hvdijk, line 179: Likewise, unintended space?

doc/overview/preface.rst
```diff
@@ -4,9 +4,9 @@ Preface
 This document provides an introduction to Codeplay's oneAPI Construction Kit.
 It should help the reader understand how to use the oneAPI Construction Kit to
 create a high-performance implementation of open-standards for a processor or
-system, i.e. SYCL, OpenCL, or Vulkan.  This implementation will enable
-programmers to use a well-understood programming model, and allow all parties
-to benefit from the open-standards software ecosystem.
+system, i.e. SYCL, OpenCL.  This implementation will enable programmers to use a
```
hvdijk, line 7: "SYCL, OpenCL" -> "SYCL or OpenCL"

modules/compiler/compiler_pipeline/source/pass_functions.cpp
```diff
@@ -60,8 +60,7 @@ uint64_t computeApproximatePrivateMemoryUsage(const llvm::Function &fn) {
     if (alloca_inst.isArrayAllocation()) {
       auto *arr_size_val = alloca_inst.getArraySize();
       auto *const_int = llvm::dyn_cast<llvm::ConstantInt>(arr_size_val);
-      assert(const_int != nullptr &&
-             "OpenCL or Vulkan Array Allocation of dynamic size");
+      assert(const_int != nullptr && "OpenCL Array Allocation of dynamic size");
```
hvdijk, line 63: It was always unnecessary to include "OpenCL or Vulkan" here, could just be "Array allocation of dynamic size"

modules/compiler/include/compiler/module.h
```diff
@@ -476,7 +476,7 @@ struct DeviceInfo {
   spv::AddressingModel addressing_model;
   /// @brief Supported memory model.
   spv::MemoryModel memory_model;
-  /// @brief Size of a device memory address in bits (Vulkan only).
+  /// @brief Size of a device memory address in bits.
```
hvdijk, line 479: If this variable is only for Vulkan we should just get rid of it.

coldav: agreed, but I looked at it and it had lots of tendrils out through the code, so i think we need to see if it can truly be removed, which I'll do in a follow up PR

doc/cmake.rst
coldav: I've changed to This provides to avoid an super long sentence

doc/developer-guide.md
coldav: I'm not sure - I think there was talk of a UR API or L0 API, but we've not gone that route so far, so I'd rather not completely drop the idea at the moment.

coldav: Regardless, perhaps do it in another pull request if we agree that's the way forward.

doc/getting-started.rst
coldav: I've rejigged this a bit now.

modules/compiler/include/compiler/module.h
hvdijk: That's fine, but in that case, can you either leave the comment here as is, or update it to say that this is not meaningful? As it stands the comment update suggests that this will become meaningful for OpenCL.

coldav: I'll add a comment.

doc/cmake.rst
```diff
@@ -86,20 +85,18 @@ hardware is unavailable.
...
-the same system.
-
-To avoid naming collisions with the ICD our OpenCL and Vulkan builds by default
-are named ``CL`` and ``VK``, however these can be overridden via the
-``CA_CL_LIBRARY_NAME`` and ``CA_VK_LIBRARY_NAME`` options. Providing customer
-teams with the flexibility to deliver on embedded systems without an ICD loader,
-where only our platform will exist and can be linked to directly by the
-application.
+The OpenCL API is designed to work through an ICD loader, a shared library named
+``OpenCL`` respectively in the form ``lib<name>.so`` or ``<name>.dll``,
```
hvdijk, line 89: "respectively" no longer makes sense here.

doc/developer-guide.md
```diff
@@ -253,10 +250,9 @@ The builtin CMake options used when invoking CMake on the command line.
   relevant llvm headers and support tools, and their version must match
   a supported LLVM version.
 
-* `CA_ENABLE_API`: Semi-colon separated list of APIs to enable. Valid values
-  are `cl` for OpenCL, and `vk` for Vulkan. Enabling an API when an optional
-  component is not present dependent on license agreement will result in a CMake
-  error. The default is `cl;vk`.
+* `CA_ENABLE_API`: Semi-colon separated list of APIs to enable. Valid values are
+  `cl` or '' for OpenCL. This currently only supports OpenCL, but that may change in
+  the future.
```
hvdijk, line 255: If we don't have any plans to add an API other than `cl`, I'm not sure we should suggest that we're considering that.

doc/modules/mux.rst
```diff
@@ -185,11 +185,10 @@ would increase build time and bloat the final binary. The way
 capabilities are handled in the build system may not be immediately
 obvious because of the number of CMake variables and the number of
 components involved. Capability-related variables are present in Abacus
-builtins, Mux, and all Mux target devices. The OpenCL and Vulkan
-implementations (``source/cl`` and ``source/vk``) do not rely on
-capability variables, as they need to generically support more than one
-Mux device, each with potentially different capabilities, at the same
-time.
+builtins, Mux, and all Mux target devices. The OpenCL implementation
+(``source/cl``) does not rely on capability variables, as they need to
```
hvdijk, line 189:
```suggestion
(``source/cl``) does not rely on capability variables, as it needs to
```

doc/getting-started.rst
```diff
@@ -168,53 +168,20 @@ Windows 11
...
 -----------
 
-When OpenCL support is required but Vulkan support is not, `SPIRV-Tools`_
-**may** optionally be installed on the system :envvar:`PATH`. This is
+SPIRV-Tools is needed to build the oneAPI Construction Kit with the SPIR-V
+testing, otherwise those tests will be disabled. This can either come from
+Vulkan SDK OR installing SPIRV-Tools directly. Windows users should install
+directly from `LunarG <https://www.lunarg.com/vulkan-sdk/>`_, as at the time of
+writing the Vulkan SDK cannot be installed through Chocolatey. On Ubuntu,
+installing spirv-tools directly is recommended.
```
hvdijk, line 182: I'm not sure this paragraph is needed, I think just the next paragraph is enough since users get a fully featured OCK build regardless of whether SPIRV-Tools are available. If we do want to keep this, it doesn't read right, it recommends too strongly to install this, it still makes it seem as if it's required.

PR #651, opened 2025-01-21
DRAFT: Fix for spir-v implicit address space cast

coldav:
# Overview

Fix for spir-v implicit address space cast

# Reason for change

It is legal for spirv to implicitly cast the address space within the bitcast instruction, but we don't handle this case. 

# Description of change

This adds a check for this and creates an AddrSpaceCast instruction.

### DIFF
--- a/modules/compiler/spirv-ll/source/builder_core.cpp
+++ b/modules/compiler/spirv-ll/source/builder_core.cpp
@@ -4018,12 +4018,15 @@ llvm::Error Builder::create<OpBitcast>(const OpBitcast *op) {
 
   llvm::Type *type = module.getLLVMType(op->IdResultType());
   llvm::Value *result = nullptr;
-  if (value->getType()->getPointerAddressSpace() != type->getPointerAddressSpace()) {
-    // We need to do an addrspacecast first as the address spaces are different.
-    // Create actual instructions as spirv_ll::Builder::replaceBuiltinGlobals() make
-    // assumptions about the user being an actual instruction, wherwas the irbuilder tries
-    // to fold it. Although that could be less brittle the solution would be to convert back
-    // to an instruction so just create as an instruction here.
+  if (llvm::isa<llvm::PointerType>(value->getType()) &&
+      llvm::isa<llvm::PointerType>(type) &&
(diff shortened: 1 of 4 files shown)


modules/compiler/spirv-ll/source/builder_core.cpp
```diff
@@ -4017,9 +4017,21 @@ llvm::Error Builder::create<OpBitcast>(const OpBitcast *op) {
...
+    // We need to do an addrspacecast first as the address spaces are
+    // different. Create actual instructions as
+    // spirv_ll::Builder::replaceBuiltinGlobals() make assumptions about the
+    // user being an actual instruction, wherwas the irbuilder tries to fold
+    // it. Although that could be less brittle the solution would be to
+    // convert back to an instruction so just create as an instruction here.
+    //value = new llvm::AddrSpaceCastInst(value, type);
+    //IRBuilder.Insert(value);
+    value = IRBuilder.CreateAddrSpaceCast(value, type);
+  }
```
hvdijk, line 4033: The effect of a bitcast is supposed to be that of a store to memory as one type, and a load from the same memory as another type. I could be wrong, but I believe in general an `addrspacecast` is not guaranteed to be bit-pattern-preserving and this could be a wrong translation, unless I am missing a guarantee that we make somewhere that it doesn't matter?

coldav: https://llvm.org/docs/LangRef.html#addrspacecast-to-instruction "(i.e. casting the result back to the original address space should yield the original bit pattern)." - are you sure it's not reasonable to do this? How else would we get the same effect?

hvdijk: Casting the result back may yield the original bit pattern, but in the different address space it may have a different bit pattern. Also from that link: "It can be a no-op cast or a complex value modification, depending on the target and the address space pair." I'm not sure what the best way to translate this is, one possibility is actually doing an `alloca` to have somewhere to store and load from, another possibility is it may be okay to do `ptrtoint` and then `inttoptr` to the other address space, there may yet be more options that I'm not seeing right now.

PR #654, opened 2025-01-23
Production docker updates.

alan-forbes-cp:
# Overview

Docker updates for the first production versions.

# Reason for change

Workflows are being "containerised". A further PR for workflow/action updates to follow when the new dockers are built.

# Description of change

New containers will be built from these dockerfiles and the updated workflows will then be pointed at them in a second related PR.

Plan is to merge the second PR immediately the new containers are available.

### DIFF
--- a/.github/dockerfiles/Dockerfile_24.04
+++ b/.github/dockerfiles/Dockerfile_24.04
@@ -58,25 +58,9 @@ RUN apt-get install --yes python3-colorama python3-distro python3-flake8 python3
 RUN PIPX_HOME=/usr/local/share/pipx PIPX_BIN_DIR=/usr/local/bin PIPX_MAN_DIR=/usr/local/share/man pipx install cmakelint lit pylint clang-format==19.1.0
 
 RUN apt-get -y install sudo
+RUN apt-get -y install gh
 RUN apt install -y zstd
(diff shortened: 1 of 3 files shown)


.github/dockerfiles/Dockerfile_24.04
```diff
@@ -78,3 +78,5 @@ RUN git clone https://github.com/ccache/ccache.git && \
     -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DENABLE_TESTING=OFF && \
   cmake --build build --target install && \
   cd - && rm -rf ccache
+
```
coldav, line 81: weird empty lines

coldav (approved):

.github/dockerfiles/Dockerfile_24.04
coldav: is there no way to install this on 24.04? My local ccache says 4.5.1 - so I don't think this is required.

alan-forbes-cp: That was taken from the Infra baseline docker - so common to 20/22/24 on Gitlab.
But you may be right ... I'll check ...

PR #655, opened 2025-01-23
Added rfc for removing vulkan SDK.

coldav:
# Overview

Added rfc for removing vulkan SDK.

# Reason for change

See rfc

# Description of change

See rfc

### DIFF
--- a/rfc-0003.md
+++ b/rfc-0003.md
@@ -1,0 +1,85 @@
+---
+rfc: 3
+title: Removing support for Vulkan
+author: Colin Davidson <colin.davidson@codeplay.com>
+discussions-to: https://github.com/uxlfoundation/oneapi-construction-kit/discussions/670
+status: draft
+---
+
+<details markdown="1">
+<summary>Table of Contents</summary>
+* Table of Contents
+{:toc}
+</details>
+
+# Abstract
+
+Vulkan API is incomplete and requires installing Vulkan SDK, as well as being
+enabled by default. It has a maintenance and design overhead and is not being
+used, as well as leading to customer confusion. It is proposed that it should be
+removed in its entirety.
+
+# Motivation
+
+Support for Vulkan is incomplete and cannot be used for full Vulkan support. It
+is not being used anywhere at present and leads to requirement on installing the
+Vulkan SDK in order to build OCK. This is also enabled by default and has led to
+confusion among users of OCK, particularly in the requirements for building. It also
+leads to strange dependency such as the mutex which doesn't really belong in the
+`compiler_pipeline` passes.
+
+# Rationale
+
+By removing it altogether this avoids any confusion on the part of the user and
+reduces the maintenance overhead.
+
+# Specification
+
+Remove everything under source/vk and from all documentation that refers to it,
+as well as any github workflow or docker requirements such as installing the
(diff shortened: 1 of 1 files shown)


rfc-0004.md
```diff
@@ -0,0 +1,94 @@
+---
+rfc: 4
+title: Removing support for Vulkan
+author: Colin Davidson <colin.davidson@codeplay.com>
+status: draft
+---
+
+``` text
```
hvdijk, line 8:
```suggestion
```
The idea was that https://uxlfoundation.github.io/oneapi-construction-kit/rfc-0002.html showed what should be used as a template for new files, but you have taken the whole `rfc-0002.md` as the template for new files. This may have been unclear in the process, thoughts welcome on how to make this clearer.

coldav: I'm a little unclear on why we should remove these lines? They seemed to give useful information.

hvdijk: For the first six lines, this is useful but it is duplicated, this is already present in the immediately following bit. For the eighth line, it means the whole RFC is treated as plain text rather than Markdown.

coldav: Yes, of course you are quite right, I should have noticed the intention. Personally I prefer a template I can just copy as is.

rfc-0004.md
```diff
@@ -0,0 +1,85 @@
+---
+rfc: 0004
```
hvdijk, line 2:
```suggestion
rfc: 3
```
No need for the leading `0`s outside of the file name, including them here would cause them to show up on the index page, and the first available RFC number is 3.

coldav: sure, renamed as well.

rfc-0003.md
```diff
@@ -0,0 +1,85 @@
+---
+rfc: 3
+title: Removing support for Vulkan
+author: Colin Davidson <colin.davidson@codeplay.com>
+sponsor: Uwe Dolinski
```
hvdijk, line 5: I know you said you just put Uwe as a sponsor to go through the same process as everyone else, but this is one case where the process is intentionally slightly different for OCK developers vs. third parties. Quoting from [RFC 1](https://uxlfoundation.github.io/oneapi-construction-kit/rfc-0001.html): "If one or more of the RFC’s co-authors are core developers, they are responsible for following the process outlined below. Otherwise (i.e. none of the co-authors are core developers), then the RFC author(s) will need to find a sponsor for the RFC." As you are a core developer, it should normally be your responsibility for following the process. I'm not necessarily opposed to having Uwe take on that responsibility just to see how it would go, but to the best of my knowledge he has not really been involved in this. "The sponsor’s job is to provide guidance to the RFC author to help them through the logistics of the RFC process (somewhat acting like a mentor)." If Uwe has been doing this and intends to continue doing this, then I'm fine with it, otherwise I think this line should just be dropped.

Other than that, I think this is ready to merge. I see some issues with the wording, but taking guidance from "Editors should generally be quite lenient about this initial review, expecting that problems will be corrected by the reviewing process", they are things we can address later, and addressing it later avoids blurring the line between editorial comments and comments on the substance.

coldav: I understand your point, but I thought it useful to keep in to show others what is expected. Although, he didn't get involved, he could of. However, I'll remove it if you wish for correctness sake.

hvdijk (approved):

PR #660, opened 2025-01-24
Opencl cts riscv64 linux

alan-forbes-cp:
# Overview

Add OpenCL CTS riscv64 linux to testing

# Reason for change

'Planned testing' requirement.  x86_64 & aarch64 (qemu) already included.

# Description of change

Fully support expanded target list including riscv64 (qemu). 
Also requires ubuntu 24.04 (docker) support.
Also adds ubuntu 24.04 llvm cache support for x86_64 & riscv64

- `vulkan_sdk_install` references removed (no longer used - included in docker)
- Looks (on the basis of 2 completed runs) like we're seeing a test fail that's not seen on Gitlab or with Github x86_64/aarch64: [cl12/math_brute_force/test_bruteforce/](https://github.com/uxlfoundation/oneapi-construction-kit/actions/runs/12990991107/job/36228345460#step:6:40633)
- Changes made in the caller workflows to limit the number of jobs run during PR development will be reversed for final merge.

### DIFF
--- a/.github/workflows/create_llvm.yml
+++ b/.github/workflows/create_llvm.yml
@@ -19,6 +19,7 @@ permissions: {}
 jobs:
   create_llvm_cache:
     strategy:
+      fail-fast: false  # at least for now
       matrix:
         version: [18, 19]
         os: [ubuntu-22.04, ubuntu-24.04, windows-2019]
@@ -72,7 +73,7 @@ jobs:
           - arch_flags: -DCMAKE_TOOLCHAIN_FILE="$GITHUB_WORKSPACE/ock/platform/arm-linux/aarch64-toolchain.cmake"
                         -DLLVM_HOST_TRIPLE=aarch64-unknown-linux-gnu
             arch: aarch64
-          - arch_flags: -DCMAKE_TOOLCHAIN_FILE="$GITHUB_WORKSPACE/ock/platform/arm-linux/riscv64-gcc-toolchain.cmake"
+          - arch_flags: -DCMAKE_TOOLCHAIN_FILE="$GITHUB_WORKSPACE/ock/platform/risc64-linux/riscv64-gcc-toolchain.cmake"
                         -DLLVM_HOST_TRIPLE=riscv64-unknown-linux-gnu
             arch: riscv64
           - arch_flags: -DLLVM_BUILD_32_BITS=ON -DLIBXML2_LIBRARIES=IGNORE -DLLVM_ENABLE_TERMINFO=OFF
--- a/.github/workflows/create_llvm.yml
+++ b/.github/workflows/create_llvm.yml
@@ -19,7 +19,6 @@ permissions: {}
 jobs:
   create_llvm_cache:
     strategy:
-      fail-fast: false  # ALAN TODO: at least for now
       matrix:
         version: [18, 19]
         os: [ubuntu-22.04, ubuntu-24.04, windows-2019]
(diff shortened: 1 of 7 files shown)


.github/workflows/create_llvm.yml
```diff
@@ -65,6 +84,8 @@ jobs:
           # This will probably require some exclude: on the different os or building all on each os.
           - os: ubuntu-22.04
             os_flags: -DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_LINK_LLVM_DYLIB=ON
+          - os: ubuntu-24.04
```
alan-forbes-cp, line 87: Review here please - these flags were a cut'n'paste from ubuntu 22

.github/workflows/create_llvm.yml
```diff
@@ -41,8 +45,20 @@ jobs:
...
             arch: aarch64
+          - os: windows-2019
+            arch: riscv64
           - os: windows-2019
             arch: x86
+          - os: ubuntu-24.04
+            version: 18
+          - os: ubuntu-24.04
+            build_type: Release
+          - os: ubuntu-22.04
```
coldav, line 56: I don't think there is much point in building it for 22.04

.github/workflows/create_llvm.yml
```diff
@@ -41,8 +45,20 @@ jobs:
...
+            arch: riscv64
+          - os: ubuntu-24.04
+            arch: aarch64
+          - os: ubuntu-24.04
+            arch: x86
```
coldav, line 61: I'm not sure right now we need to build for anything but riscv64, given how we cache it etc, I think it would lead to complications.

.github/workflows/create_llvm.yml
coldav: we might be able to set some env variables or default inputs to avoid duplications here, maybe worth doing in some future refactor - I think we'll need to rejig how we do the llvms anyway,

.github/workflows/create_llvm.yml
alan-forbes-cp: Yes - riscv64 is 24.04/RelAssert only as things stand.

.github/workflows/create_llvm.yml
alan-forbes-cp: Only riscv64 and x86_64 are 24.04 (RelAssert) as things stand.

.github/workflows/create_llvm.yml
alan-forbes-cp: I think our options for a minor refactor here are limited. The value (e.g. `ubuntu-24.04`) cannot be an array. It may be possible to use "if/then/else" logic (there are some `exclude` examples showing that) but as it must resolve to a string the complexity of the condition would likely be no more appealing than the current duplication. My vote would be to retain current form pending a future llvm refactor.

.github/workflows/create_llvm.yml
```diff
@@ -41,8 +45,20 @@ jobs:
             build_type: Release
           - os: windows-2019
             arch: aarch64
+          - os: windows-2019
+            arch: riscv64
           - os: windows-2019
             arch: x86
+          - os: ubuntu-24.04
+            version: 18
```
coldav, line 53: I'd be tempted to just do 24.04 19, riscv RelAssert (and possible x86_64 if needed).

alan-forbes-cp: Yes - these are the two new caches with this PR. (The highlighted text being from the `exclude` list.)

.github/workflows/create_llvm.yml
coldav: I think it's fine for now, I plan to change how it's done.

.github/workflows/create_llvm.yml
coldav: oh yes, sorry, it's confusing

PR #664, opened 2025-01-29
DRAFT: add cp-graviton

alan-forbes-cp:
# Overview

and consequences.*

# Reason for change

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/.github/workflows/planned_testing_caller.yml
+++ b/.github/workflows/planned_testing_caller.yml
@@ -23,10 +23,11 @@ jobs:
     if: github.repository == 'uxlfoundation/oneapi-construction-kit' || github.event_name != 'schedule'
     uses: ./.github/workflows/planned_testing.yml
     with:
-      target_list: '[ "host_x86_64_linux", "host_aarch64_linux", "host_riscv64_linux", "host_i686_linux", "host_refsi_linux", "host_x86_64_windows" ]'
+      #target_list: '[ "host_x86_64_linux", "host_aarch64_linux", "host_riscv64_linux", "host_i686_linux", "host_refsi_linux", "host_x86_64_windows" ]'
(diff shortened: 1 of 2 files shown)


.github/dockerfiles/Dockerfile_22.04-graviton
```diff
@@ -0,0 +1,9 @@
+FROM ubuntu:22.04
```
coldav, line 1: nothing specific to graviton here - shall we name it -aarch64? Also should we consider renaming the others with suffixes?

.github/dockerfiles/Dockerfile_22.04-graviton
```diff
@@ -0,0 +1,9 @@
...
+
+RUN apt-get update
+RUN apt-get -y install sudo
+RUN apt-get install -y git
+RUN apt-get install --yes cmake
+RUN apt-get install --yes gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
+RUN apt-get install --yes python3 python3-pip
+RUN apt-get install --yes spirv-tools
```
coldav, line 9: Seems like this is enough for opencl-cts, but a much cut down version than what we have elsewhere - it's possibly not enough to build dpc++ (not sure).

alan-forbes-cp: Its specifically for OpenCL-CTS as things stand. My thinking was that anything further would follow as required.

.github/dockerfiles/Dockerfile_22.04-graviton
coldav: I can accept that for now, but it does feel a bit lopsided if for example we wanted to push another job to aarch64 or suggest it as a base to customers.

coldav (approved):

PR #665, opened 2025-01-29
Ensure that expected fails on CTS always show as a pass.

coldav:
# Overview

Ensure that expected fails on CTS always show as a pass.
    
# Reason for change

When we have regular fails which we do not expect an immediate fix, we
want to show as pass so we can catch be failures more easily.

# Description of change

When we know there is an expected fail, we create files to cover fails
on all targets or fails on a particular target for OpenCL CTS and
SYCL CTS. These will be under .github/opencl_cts and .github/sycl_cts.
Fails files will be of the form expect_fail_all.csv or expect_fail_<target>.csv.
These should match lines in the original csv file being used.

### DIFF
--- a/.github/actions/run_opencl_cts/action.yml
+++ b/.github/actions/run_opencl_cts/action.yml
@@ -27,7 +27,6 @@ runs:
         CTS_FILTER: cts-3.0-online-ignore-linux-host.csv
         CTS_TIMEOUT: 18:00:00 # OK for github?
       run: |
-        cd "$GITHUB_WORKSPACE/OpenCL-CTS"
         echo "Running OpenCL CTS tests with CTS file $CTS_CSV_FILE with filter $CTS_FILTER"
         export QEMU_SETTING=""
         if [[ "${{inputs.target}}" =~ .*aarch64.* ]] ; then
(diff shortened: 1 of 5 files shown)


.github/actions/run_opencl_cts/action.yml
```diff
@@ -37,6 +36,11 @@ runs:
         fi
         echo QEMU SETTING: $QEMU SETTING
         set -x
+        echo > expect_fail.csv
+        # $CTS_FILTER ignores certain test, so is treated differently to temporary fails.
```
hvdijk, line 40: This suggests to me that `CTS_FILTER` and `expect_fail_*.csv` are treated differently, but they are concatenated into a single `disable.csv` file and then treated the same? I think the result is probably fine, I just don't understand the comment.

We may be able to simplify this somewhat: something along the lines of `for file in ".github/opencl_cts/expect_fail_all.csv" ".github/opencl_cts/expect_fail_${{ inputs.target }}.csv" "$GITHUB_WORKSPACE/source/cl/scripts/$CTS_FILTER"; do [ -f "$file" ] && cat "$file"; done > disable.csv`?

alan-forbes-cp: `expect_fail.csv` is also being re-used in the next step so should be retained.

hvdijk (approved):

PR #669, opened 2025-02-05
RFC 2: Include more detail on how to write RFCs.

hvdijk:
# Overview

RFC 2: Include more detail on how to write RFCs.

# Reason for change

# Description of change

points, etc.*

Also fix the headers of RFC 1.

# Checklist

### DIFF
--- a/rfc-0002.md
+++ b/rfc-0002.md
@@ -2,16 +2,7 @@
 rfc: 2
 title: RFC template
 author: Harald van Dijk <harald.vandijk@codeplay.com>
-status: draft
----
-
-``` text
----
-rfc: <REQUIRED: RFC number>
-title: <REQUIRED: RFC title>
-author: <REQUIRED: list of authors' names and optionally, email addresses>
-sponsor: <name of sponsor>
-status: <REQUIRED: draft | active | accepted | rejected | withdrawn | final>
+status: active
 ---
 
 <details markdown="1">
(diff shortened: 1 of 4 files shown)


rfc-0002.md
```diff
@@ -22,50 +13,67 @@ status: <REQUIRED: draft | active | accepted | rejected | withdrawn | final>
...
+  address. Be sure to follow the format carefully: your name must appear first,
+  and it must not be contained in parentheses. Your email address may appear
+  second (or it can be omitted) and if it appears, it must appear in angle
+  brackets. It is okay to obfuscate your email address.
+- If none of the authors are core developers, include a Sponsor header with the
+  name of the core developer sponsoring your RFC.
+- Add the direct URL of the RFC's canonical discussion thread (on e.g. GitHub
+  Discussions) under the discussions-to header. If the thread will be created
+  after the RFC is submitted as an official draft, it is okay to just list the
+  venue name initially, but remember to update the RFC with the URL as soon as
```
coldav, line 58: what does venue name mean?

rfc-0002.md
```diff
@@ -22,50 +13,67 @@ status: <REQUIRED: draft | active | accepted | rejected | withdrawn | final>
...
+  should not (`rfc: 2`).
+- Change the title header to the title of your RFC.
+- Change the author header to include your name, and optionally your email
+  address. Be sure to follow the format carefully: your name must appear first,
+  and it must not be contained in parentheses. Your email address may appear
+  second (or it can be omitted) and if it appears, it must appear in angle
+  brackets. It is okay to obfuscate your email address.
+- If none of the authors are core developers, include a Sponsor header with the
+  name of the core developer sponsoring your RFC.
+- Add the direct URL of the RFC's canonical discussion thread (on e.g. GitHub
```
coldav, line 55: suggest we prefer on github discussions under category rfcs (for now).

rfc-0002.md
hvdijk: RFC 1 allows different venues for discussion, while preferring GitHub Discussions:

> The RFC authors (or sponsor, if applicable) may select any reasonable venue for the discussion, so long as the the following criteria are met:
> 
> - The forum is appropriate to the RFC’s topic.
> - The thread is publicly available on the web so that all interested parties can participate.
> - The discussion is subject to the [oneAPI Construction Kit Code of Conduct](https://github.com/uxlfoundation/oneapi-construction-kit/blob/main/CODE_OF_CONDUCT.md).
> - A direct link to the current discussion thread is provided in the RFC under the `discussions-to` header.
> 
> The [oneAPI Construction Kit Discussions](https://github.com/uxlfoundation/oneapi-construction-kit/discussions) is the preferred choice for most RFCs.

So the venue name in most cases would be "oneAPI Construction Kit Discussions" or something that expresses the same thing, to be replaced by an URL when the discussion thread gets created.

rfc-0002.md
hvdijk: We do, but that's in RFC 1 rather than in RFC 2. Should I add that here too?

rfc-0002.md
coldav: right I see, it wasn't obvious to me what that meant reading it.

rfc-0002.md
```diff
@@ -22,50 +13,67 @@ status: <REQUIRED: draft | active | accepted | rejected | withdrawn | final>
...
+ automatically rejected because of form.
+
+Markdown provides PEP authors with useful functionality and expressivity, while
+maintaining easy readability in the source text. The processed HTML form makes
+the functionality accessible to readers: live hyperlinks, styled text, tables,
+images, and automatic tables of contents, among other advantages.
+
+# How to Use This Template
+
+- Make a copy of [`rfc-NNNN.md`][rfc-nnnn] (the `.md` file, **not** the HTML!)
```
coldav, line 41: should this be rfc-0002/rfc.NNNN.md?

rfc-0002.md
coldav: it's probably okay - although we should consider shepherding them towards a single category (e.g. rfcs) under discussions.

coldav (approved):

rfc-0002.md
hvdijk: The link will go to the right file, but no harm in including the path name in the link text as well, will update.

rfc-0002.md
hvdijk: That's probably a good idea. We don't have that category at the moment, so I don't think I can include it in this PR just yet? It seems like a good improvement to make afterwards though.

coldav: I can create it now if we agree on it?

hvdijk: Created now, and PR updated to refer to it (in RFC 1 rather than RFC 2, based on the PEPs that were the basis for it)

PR #675, opened 2025-02-13
[CI] Change to use --prepend-path for qemu using city runner

coldav:
# Overview

 The qemu path for city runner was having slightly different behaviour  than the default profile. Rather than fixing it there,
 move to using  --prepend-path which means the same profile is used. 

# Reason for change

aarch64/riscv64 qemu failures on spirv_new in opencl-cts


hvdijk (approved):
As usual, approval based on the `.github/workflows/planned_testing_caller.yml` changes being reverted before merging.

PR #678, opened 2025-02-14
Switch to calling an action for PR tests and improve caching

coldav:
# Overview

Switch to calling an action for PR tests and populate the cache sequentially

# Reason for change

There was a race condition in updating the cache as it just ran the tests overnight in parallel. 

# Description of change

Move all the build parts of PR jobs to an action. Add a workflow pr_tests_cache.yml to seed the caches by calling the actions in sequence. The scheduled run_pr_tests overnight will no longer have the ability to seed the cache and this will only be done via pr_tests_cache.yml.

This does not yet delete the caches. We can't be sure this works so this is initially commented out.

# Checklist

### DIFF
--- a/.github/actions/do_build_pr/run_ubuntu_clang_x86_llvm_latest_cl3_0_offline/action.yml
+++ b/.github/actions/do_build_pr/run_ubuntu_clang_x86_llvm_latest_cl3_0_offline/action.yml
@@ -1,0 +1,42 @@
+name: build_pr_ubuntu_clang_x86_llvm_latest_cl3_0_offline
+description: Build pr ubuntu_clang_x86_llvm_latest_cl3_0_offline
+
+inputs:
+  cache_seed:
+    type: boolean
+    default: false
+
+runs:
+  using: "composite"
+  steps:
+      - name: remove any old dirs
+        shell: bash  
+        run:
+          rm -rf build build_offline install_offline
+
+      - name: build ock x86 relassert
+        uses: ./.github/actions/do_build_ock
+        with:
+          build_32_bit: ON
+          extra_flags: -DCMAKE_C_COMPILER=$GITHUB_WORKSPACE/llvm_install/bin/clang -DCMAKE_CXX_COMPILER=$GITHUB_WORKSPACE/llvm_install/bin/clang++
+          build_targets: ${{ inputs.cache_seed == 'true' && 'UnitCL clc' || 'check-ock' }}
+          enable_api: ""
+          builtin_kernel: ON
+          use_linker: gold
+          debug_support: ON
+          offline_kernel_tests: ${{ inputs.cache_seed == 'true' && 'OFF' || 'ON' }}
+          
+      - name: build ock x86 offline
+        uses: ./.github/actions/do_build_ock
+        with:
+          build_32_bit: ON
+          extra_flags: -DCMAKE_C_COMPILER=$GITHUB_WORKSPACE/llvm_install/bin/clang -DCMAKE_CXX_COMPILER=$GITHUB_WORKSPACE/llvm_install/bin/clang++
+          build_targets:  ${{ inputs.cache_seed == 'true' && 'UnitCL' || 'check-ock' }}
+          runtime_compiler_enabled: OFF
+          assemble_spirv_ll_lit_test_offline: ${{ inputs.cache_seed == 'true' && 'OFF' || 'ON' }}
+          external_clc: ${GITHUB_WORKSPACE}/build/bin/clc
+          use_linker: gold
+          debug_support: ON
+          install_dir: $GITHUB_WORKSPACE/install_offline
+          build_dir: $GITHUB_WORKSPACE/build_offline
+          offline_kernel_tests: ${{ inputs.cache_seed == 'true' && 'OFF' || 'ON' }}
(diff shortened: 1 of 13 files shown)


.github/workflows/pr_tests_cache.yml
```diff
@@ -0,0 +1,222 @@
...
+        - ${{github.workspace}}:${{github.workspace}}    
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+        # installs tools, ninja, installs llvm and sets up sccache
+      - name: setup-ubuntu
+        uses: ./.github/actions/setup_build
+        with:
+          llvm_version: '19'
```
alan-forbes-cp, line 141: Should all `llvm_version` settings use an env var?

.github/workflows/pr_tests_cache.yml
```diff
@@ -0,0 +1,222 @@
...
+        uses: ./.github/actions/setup_build
+        with:
+          llvm_version: ${{ env.llvm_previous }}
+          llvm_build_type: RelAssert
+          save: true
+
+      - name: build host_x86_64
+        uses:  ./.github/actions/do_build_pr/run_host_x86_64
+        with:
+          cache_seed: true
```
alan-forbes-cp, line 54: Is there a case for flipping the default setting of `cache_seed` rather than always using `true` with each action call?

.github/workflows/run_pr_tests.yml
```diff
@@ -3,10 +3,6 @@ name: Run ock tests for PR testing
 on:
   workflow_call:
     inputs:
-      update_cache:
-        required: false
-        type: boolean
-        default: false
```
hvdijk, line 9: If we remove this input, is the effect that we always update the cache, or that we never update the cache? Is there a description somewhere, whether in this PR or elsewhere, of what the new behaviour is meant to be?

.github/workflows/pr_tests_cache.yml
coldav: no, I prefer to keep the non-cache seed calls simple

.github/workflows/run_pr_tests.yml
coldav: never update the cache. I can improve the description.

.github/workflows/pr_tests_cache.yml
coldav: fixed

.github/workflows/run_pr_tests.yml
coldav: I've updated the description on the PR and commit.

.github/actions/do_build_pr/run_ubuntu_gcc_x86_64_riscv_fp16_cl3_0_unitcl_vecz/action.yml
```diff
@@ -0,0 +1,31 @@
...
+  steps:
+      - name: remove any old dirs
+        shell: bash  
+        run:
+          rm -rf build
+
+      - name: build ock
+        uses: ./.github/actions/do_build_ock
+        with:
+          build_targets: ${{ inputs.cache_seed == 'true' && 'clc UnitCL' || 'check-ock-UnitCL-group-vecz' }}
```
hvdijk, line 20: As `inputs.cache_seed` is a boolean rather than a string,
```suggestion
          build_targets: ${{ inputs.cache_seed && 'clc UnitCL' || 'check-ock-UnitCL-group-vecz' }}
```
should be sufficient? Likewise in many other places. But I'm not entirely sure of what data types exist in GitHub actions, and if this suggestion does not work, then what you have already is fine.

.github/actions/do_build_pr/run_windows_msvc_x86_64_llvm_latest_cl3_0_offline/action.yml
```diff
@@ -0,0 +1,43 @@
...
+    type: boolean
+    default: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: remove any old dirs
+      shell: bash  
+      run:
+        rm -rf build build_offline
```
hvdijk, line 18: Is it intended that previous `install`/`install_offline` directories are not cleaned up?

.github/actions/do_build_pr/run_ubuntu_gcc_x86_64_riscv_fp16_cl3_0_unitcl_vecz/action.yml
coldav: I thought that too, but it doesn't work - it even goes wrong without the quotes.

.github/actions/do_build_pr/run_windows_msvc_x86_64_llvm_latest_cl3_0_offline/action.yml
coldav: good spot, I think it's the only time we do the install. will fix.

.github/actions/do_build_pr/run_ubuntu_clang_x86_llvm_latest_cl3_0_offline/action.yml
```diff
@@ -0,0 +1,42 @@
...
+    type: boolean
+    default: false
+
+runs:
+  using: "composite"
+  steps:
+      - name: remove any old dirs
+        shell: bash  
+        run:
+          rm -rf build build_offline
```
hvdijk, line 15: This one also looks like it should clean `install_offline`

hvdijk (approved):

PR #684, opened 2025-02-18
Drop CA_RISCV_ENABLED CMake option.

hvdijk:
# Overview

Drop CA_RISCV_ENABLED CMake option.

# Reason for change

The CA_RISCV_ENABLED option was provided because we simultaneously want to enable all available Mux targets by default, and disable the RISC-V Mux target by default. This is confusing when CA_RISCV_ENABLED and CA_MUX_TARGETS_TO_ENABLE are mismatched: it causes the RISC-V Mux target to be disabled when either variable disables it, even when the other variable suggests that it should be enabled.

# Description of change

Drop CA_RISCV_ENABLED and rely entirely on CA_MUX_TARGETS_TO_ENABLE instead.

For consistency, make the same change to CA_MUX_COMPILERS_TO_ENABLE.

For both CA_MUX_TARGETS_TO_ENABLE and CA_MUX_COMPILERS_TO_ENABLE, change the default to "host": we expect builds that rely on other targets or other compilers to already explicitly specify the desired target and compiler.

# Checklist


doc/modules/riscv.rst
```diff
@@ -138,7 +137,6 @@ target has additional features. This is done by adding to the build options:
 .. code-block:: console
 
  $ cmake -GNinja \
-   -DCA_RISCV_ENABLED=ON \
```
coldav, line 141: Please add to CHANGELOG.md so we catch it for the release notes.

hvdijk: Thanks, done.

doc/modules/riscv.rst
```diff
@@ -125,7 +125,6 @@ Currently recommended build options include:
...
    -DCA_MUX_TARGETS_TO_ENABLE="riscv" \
```
coldav, line 128: surely this is wrong now if we have to enable the compiler one?
I'm not totally convinced that we shouldn't pick up a default one if we have the mux one enabled, it feels quite a lot and I think it might confuse people.

CMakeLists.txt
```diff
@@ -112,11 +112,11 @@ endif()
 ca_option(CA_ENABLE_DOCUMENTATION BOOL
   "Enable documentation generation" ON)
 ca_option(CA_MUX_TARGETS_TO_ENABLE STRING
-  "Semi-colon separated list of ComputeMux targets to enable, empty enables all" "")
+  "Semi-colon separated list of ComputeMux targets to enable, empty enables all" "host")
 ca_option(CA_EXTERNAL_MUX_TARGET_DIRS STRING
   "Semi-colon separated list of external ComputeMux target source directories" "")
 ca_option(CA_MUX_COMPILERS_TO_ENABLE STRING
-  "Semi-colon separated list of ComputeMux compilers to enable, empty enables all" "")
+  "Semi-colon separated list of ComputeMux compilers to enable, empty enables all" "host")
```
coldav, line 119: could we perhaps make it so it stays "" by default, but this means match the mux targets enabled if they exist? That would be closer to what we had before.

coldav: Also the description doesn't match the default here.

hvdijk: The description says empty enables all, it doesn't say empty is the default, I think the description is right. But I like your idea of matching the Mux targets.

coldav: yes you are right about the default, sorry about that.

PR #686, opened 2025-02-19
Added Xfail and Mayfail attributes to city runner

coldav:
# Overview

Support attribute Xfail and Mayfail for city runner and an add a --override option to override entries in the main csv file. 

# Reason for change

We want to be able to run tests that we know fail, but we don't want an overall fail.

# Description of change

Added xfail and Mayfail as attributes to city runner.

Xfail means we expect it to fail and will give a failing return code if it unexpectedly pass, otherwise it does not contribute to an overall fail.

Mayfail means it sometimes fails and will not contribute to a failing return code.

Unlike some of the other attributes, these cannot be set with additional csv files, as it was felt this leads to overcomplexity. Instead it is expected that some form of csv combination script could be used alongside this.

Added --override option. This provides an additional csv file which can override entries in the main csv file if it matches the first two values in an entry. This is useful for overriding attributes such as Xfail.

The following is example output

[dump.txt](https://github.com/user-attachments/files/18872722/dump.txt)

# Checklist

### DIFF
--- a/.github/actions/run_opencl_cts/action.yml
+++ b/.github/actions/run_opencl_cts/action.yml
@@ -34,11 +34,16 @@ runs:
         fi
         echo QEMU SETTING: $QEMU_SETTING
         set -x
-        echo > expect_fail.csv
+        # Build override file, all is done first, then the target specific.
+        # The last file can overwrite previous overrides.        
+        for csv in .github/opencl_cts/override_all.csv .github/opencl_cts/override_${{ inputs.target }}.csv
+        do
+           [ -f $csv ] && cat $csv >> override.csv
+        done > override.csv
+        echo override file:
+        cat override.csv
+
         # $CTS_FILTER ignores certain test, so is treated differently to temporary fails.
-        [ -f .github/opencl_cts/expect_fail_all.csv ] && cat .github/opencl_cts/expect_fail_all.csv >> expect_fail.csv
-        [ -f .github/opencl_cts/expect_fail_${{ inputs.target }}.csv ] && cat .github/opencl_cts/expect_fail_${{ inputs.target }}.csv >> expect_fail.csv
-        cat expect_fail.csv "$GITHUB_WORKSPACE/source/cl/scripts/$CTS_FILTER" > disable.csv 
         # Note: use 'eval' built-in to handle quoting/escaping/splitting reqs
         RUN_CITIES="python3 -u $GITHUB_WORKSPACE/scripts/testing/run_cities.py -v \
           --color=always --timeout $CTS_TIMEOUT \
@@ -49,23 +54,6 @@ runs:
           -e OCL_ICD_FILENAMES=$GITHUB_WORKSPACE/install_ock/lib/libCL.so \
           -e CL_PLATFORM_INDEX=0 \
           -s $GITHUB_WORKSPACE/test_conformance/$CTS_CSV_FILE \
-          -i disable.csv"
+          -i $GITHUB_WORKSPACE/source/cl/scripts/$CTS_FILTER \
+          -o override.csv"
         eval $RUN_CITIES
-
-    - name: Run expected failed opencl cts
-      shell: bash
-      env:
-        CTS_TIMEOUT: 18:00:00
-      run: |
-        echo "Running OpenCL CTS tests with CTS file $CTS_CSV_FILE with filter $CTS_FILTER"
-        set -x
-        RUN_CITIES="python3 -u $GITHUB_WORKSPACE/scripts/testing/run_cities.py -v \
-          --color=always --timeout $CTS_TIMEOUT \
-          $QEMU_SETTING \
-          -b $GITHUB_WORKSPACE/test_conformance \
-          -L $GITHUB_WORKSPACE/install_icd/lib \
-          -e CLC_EXECUTABLE=$GITHUB_WORKSPACE/install_ock/bin/clc \
-          -e OCL_ICD_FILENAMES=$GITHUB_WORKSPACE/install_ock/lib/libCL.so \
-          -e CL_PLATFORM_INDEX=0 \
-          -s expect_fail.csv"
-        eval $RUN_CITIES || echo failed as expected
--- a/.github/actions/run_opencl_cts/action.yml
+++ b/.github/actions/run_opencl_cts/action.yml
@@ -35,7 +35,7 @@ runs:
         echo QEMU SETTING: $QEMU_SETTING
         set -x
         # Build override file, all is done first, then the target specific.
-        # The last file can overwrite previous overrides.        
+        # The last file can overwrite previous overrides.
         for csv in .github/opencl_cts/override_all.csv .github/opencl_cts/override_${{ inputs.target }}.csv
         do
            [ -f $csv ] && cat $csv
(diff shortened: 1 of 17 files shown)


scripts/testing/city_runner/test_info.py
```diff
@@ -156,12 +174,17 @@ def from_file(cls, list_file_paths, disabled_file_path, ignored_file_path, prefi
 
                     if len(chunks) >= 3:
                         attribute = chunks[2]
+                        print("___CSD__ attribute=", attribute)
```
hvdijk, line 177: This was probably left in by mistake?

scripts/testing/city_runner/test_info.py
```diff
@@ -156,12 +174,17 @@ def from_file(cls, list_file_paths, disabled_file_path, ignored_file_path, prefi
...
                         if attribute == 'Ignore':
                             ignored = True
                         elif attribute == 'Disabled':
                             disabled = True
                         elif attribute == 'Unimplemented':
                             unimplemented = True
+                        elif attribute == 'Xfail':
+                            xfail = True
+                        elif attribute == 'Mayfail':
+                            mayfail = True
```
hvdijk, line 187: While we're making changes here, do you think we could make these attributes case-insensitive? For consistency with the existing attributes, absolutely we should accept `Xfail` but for ease of typing I'd like it if we could accept `xfail` too. But I'm okay leaving it like this too if there's any reason to not do that.

scripts/testing/city_runner/test_info.py
```diff
@@ -265,6 +290,9 @@ def __init__(self, tests):
         self.num_tests = len(tests)
         self.num_passes = 0
         self.num_fails = 0
+        self.num_xfail_unexpectedly_passed = 0
+        self.num_xfail_expected_fails = 0 
+        self.num_mayfail_fails = 0;
```
hvdijk, line 295:
```suggestion
        self.num_mayfail_fails = 0
```

scripts/testing/city_runner/test_info.py
coldav: yes, still tidying up, just firing up a run.

scripts/testing/city_runner/test_info.py
```diff
@@ -317,6 +358,9 @@ def finish(self, profile):
         not_runs = []
         self.fail_list = []
         self.timeout_list = []
+        self.xfail_unexpectedly_passed_list = []
```
hvdijk, line 361: There's a common enough term for xfailed tests that end up passing that we could use:
```suggestion
        self.xpass_list = []
```

scripts/testing/city_runner/runner.py
```diff
@@ -260,8 +262,11 @@ def process_results(self):
         # Return the city runner exit code.
         if self.aborted:
             return 130
-        if self.results.num_fails and not self.args.relaxed:
-            return 1
+        if not self.args.relaxed:
+            if self.results.num_fails > self.results.num_xfail_expected_fails + self.results.num_mayfail_fails:
```
hvdijk, line 266: This logic looks suspicious. Suppose we have two tests, `a` and `b`, where `a` has no attributes and `b` is `mayfail`. Suppose `a` fails and `b` passes. I think this logic would report that as a success but it should be a failure.

coldav: after discussion, I think this goes away if num_fails is actually fails that don't from an xfail fail or a mayfail fail, so we can simplify this logic.

scripts/testing/city_runner/test_info.py
coldav: good idea

scripts/testing/city_runner/test_info.py
```diff
@@ -138,11 +146,20 @@ def from_file(cls, list_file_paths, disabled_file_path, ignored_file_path, prefi
                         device_filter = chunks.pop(0)
                     if len(chunks) < 2:
                         raise Exception("Not enough columns in the CSV file")
+                    
+                    # match on the first 2
+                    for o in override_tests:
+                        if chunks[:2] == o[:2]:
+                            chunks = o
```
hvdijk, line 153: Ideally we would use a dictionary for `override_tests` so that we would need nested loops, but ideally we would also use a set for `ignored_tests` and `disabled_tests` and avoid the nested loop there. If you like, we could change that now, but this is also okay to leave as you have it now.

.github/actions/run_opencl_cts/action.yml
```diff
@@ -34,11 +34,16 @@ runs:
         fi
         echo QEMU SETTING: $QEMU_SETTING
         set -x
-        echo > expect_fail.csv
+        # Build override file, all is done first, then the target specific.
+        # The last file can overwrite previous overrides.        
+        for csv in .github/opencl_cts/override_all.csv .github/opencl_cts/override_${{ inputs.target }}.csv
+        do
+           [ -f $csv ] && cat $csv >> override.csv
```
hvdijk, line 41:
```suggestion
           [ -f $csv ] && cat $csv
```

.github/actions/run_sycl_cts/action.yml
```diff
@@ -67,12 +67,18 @@ runs:
         export LD_LIBRARY_PATH=$GITHUB_WORKSPACE/install_dpcpp/lib:$GITHUB_WORKSPACE/install_ock/lib
         export ONEAPI_DEVICE_SELECTOR=opencl:0
         export CTS_CSV_FILE=$GITHUB_WORKSPACE/.github/scripts/sycl-cts.csv
-        echo > expect_fail.csv
+        echo > override.csv
```
hvdijk, line 70:
```suggestion
```

.github/actions/run_sycl_cts/action.yml
```diff
@@ -85,30 +91,11 @@ runs:
           -l SYCL-CTS/cts.log -f SYCL-CTS/cts.fail \
           -r SYCL-CTS/cts.xml \
           -v \
-          -i disable.csv || exitcode=$?
+          -o override.csv \
+          $SYCL_CTS_FILTER
```
hvdijk, line 95: Is removing the setting of `exitcode` intentional?

.github/actions/run_sycl_cts/action.yml
coldav: I don't think the line is needed at all given the for loop >

hvdijk: Yes, my suggested change is to remove that line, that's why it shows in red :)

scripts/testing/city_runner/test_info.py
coldav: I thought about it but this file should be quite small and we are allowing successive overrides on the same key. I don't think this will be at all noticeable in the start up time, so I'd rather leave it as is for now anyway.

hvdijk: Makes sense

hvdijk (commented):
Looks pretty good! Some odd whitespace and perhaps slightly more verbose naming and complicated logic than needed, left some suggestions for that, but the logic for the results looks right as far as I can see

hvdijk (approved):

PR #691, opened 2025-02-26
Add a delete cache action

coldav:
# Overview

Add a generic delete cache action. use this in the workflow. Long term plan to use this action in codeql, pr cache workflow etc.

# Reason for change

Use in different workflows.


alan-forbes-cp (approved):
Like the dryrun option - nice.

PR #697, opened 2025-02-27
Add a workflow to create llvm artefacts for a branch

coldav:
# Overview

Change planned testing to build llvm and run all tests.

# Reason for change

The current system was very limited to being able to cope with updates of llvm. It also did not allow us to run all testing on
one llvm or the other. Additionally, it lacked flexibility.

# Description of change

Build llvm as an artefact before testing. Option to support using caching on a per llvm basis - but unlike before this will build it if it does not find it.

planned_testing_caller.yml has been changed to call both internal tests (run_ock_internal_tests.yml renamed from run_pr_tests.yml) and external tests (run_ock_external_tests.yml, renamed from
planned_testing.yml). It has also been changed so it can only be run via being called. It allows a wide variety of inputs so different variants of the pipeline can be run.

At the top of the planned testing change are per llvm callers, planned_testing_caller_19.yml and planned_testing_caller_20.yml. These do little more than call planned_testing.yml with different llvms on a cron job. They are not expected to change much.
llvm_source is passed down to the run_* workflows, and has a run id option which means it downloads it from that run id. If it is the same workflow id as being run on, then it will use download-artifact, otherwise it will use gh.

This additionally switches the overnight testing to llvm 19 and llvm 20 from a previous 18/19 mix, run on different nights.

Some extra flexibility has also been added for which runners are used as it was difficult to test with the conflict on available runners.


.github/actions/download_artifact/action.yml
```diff
@@ -0,0 +1,50 @@
...
+      run: |
+        git config --global --add safe.directory $PWD
+        GH_TOKEN=${{ inputs.token }} gh run download ${{ inputs.run_id }} -n ${{ inputs.name }} -D ${{ inputs.needs_tar == 'true' && github.workspace || inputs.path }}
+
+    - name: untar artefact
+      if: inputs.needs_tar == 'true'
+      shell: bash
+      run: |
+        # The strip-components means everything goes under the new path
+        pwd
```
coldav, line 46: remove debug

.github/actions/download_artifact/action.yml
```diff
@@ -0,0 +1,50 @@
+name: download artifact
+description: Download an artifact, either through the action of using gh, depending which workflow it came from.
```
coldav, line 2: of -> or

.github/workflows/planned_testing_caller_20.yml
```diff
@@ -0,0 +1,33 @@
...
+  L20_:
+    if: github.repository == 'uxlfoundation/oneapi-construction-kit' || github.event_name != 'schedule'
+    uses: ./.github/workflows/planned_testing_caller.yml
+    with:
+      llvm_version: '20'
+      llvm_branch: 'release/20.x'
+      # We can set ock, test_sycl_cts etc here optionally if this is used as a
+      # pull request. Any parameters below this is intended for local testing
+      # and should not be merged nor reviewed (other than checking it should not be merged).
+      ock: true
```
coldav, line 25: remove all this debug

.github/sycl_cts/override_opencl_llvm_19.csv
```diff
@@ -0,0 +1,4 @@
+SYCL_CTS,test_math_builtin_api "math_builtin_float_base_*",Xfail
+SYCL_CTS,test_math_builtin_api "math_builtin_float_double_*",Xfail
+SYCL_CTS,test_event "event::wait does not report asynchronous errors",Mayfail
+SYCL_CTS,test_language,Xfail
```
hvdijk, line 4:
```suggestion
```
The former is still in `override_all.csv` so shouldn't be needed in this file, the latter was probably left over from back when it failed to build?

.github/workflows/planned_testing_caller_19.yml
```diff
@@ -0,0 +1,24 @@
+name: run planned tests for llvm 19
```
alan-forbes-cp, line 1: Maybe a comment as to how the llvm callers should be updated when a new llvm version is added? 
I'm assuming we want to change the schedule and job params in several places.

.github/workflows/planned_testing_caller.yml
```diff
@@ -1,44 +1,115 @@
...
       # The following can be used to download from a previous workflow run (change id)
-      #download_ock_artefact: host_x86_64_linux=14110849023;host_aarch64_linux=14110849023;host_riscv64_linux=14110849023
-      #download_dpcpp_artefact: host_x86_64_linux=14110849023;host_aarch64_linux=14110849023;host_riscv64_linux=14110849023
-      #download_sycl_cts_artefact: host_x86_64_linux=14110849023;host_aarch64_linux=14110849023;host_riscv64_linux=14110849023
-      pull_request: ${{ github.event_name == 'pull_request' }}
+      download_ock_artefact: ${{ inputs.download_ock_artefact }}
+      download_dpcpp_artefact: ${{ inputs.download_dpcpp_artefact }}
+      download_sycl_cts_artefact: ${{ inputs.download_sycl_cts_artefact }}
+
+      # Override this value if we wish to use a previous workflow's llvm binary
```
alan-forbes-cp, line 99: Could these changes be triggered via a new input param? e.g. `llvm_workflow_id` ??

coldav: I believe this will be infrequent, so at the moment I don't want to overcomplicate it (there are issues with having to not run the other job etc) for the PR at least.

.github/workflows/planned_testing_caller_19.yml
coldav: I can add some comments, but the process will involve changes in multiple places, so it's tricky to do that now.

.github/sycl_cts/override_opencl_llvm_19.csv
coldav: I've removed.

.github/actions/setup_build/action.yml
```diff
@@ -24,8 +24,12 @@ inputs:
     description: 'Cross compilation architecture from: x86, arm, aarch64, riscv64. Default: "none" (no cross compile), will auto fetch native arch'
     default: "none"
   llvm_source:
-    description: 'source to get llvm from - one of install or cache'
+    description: 'source to get llvm from - one of install, cache or workflow id'
     default: "cache"
+  github_token:
+    description: "gh token for downloading artifacts"
+    default: 'TOKEN_NOT_SET'
```
hvdijk, line 31: Do we not have `${{ env.GH_TOKEN }}` available here?

.github/workflows/run_ock_internal_tests.yml
```diff
@@ -186,9 +201,9 @@ jobs:
 
   # Based on: mr-windows-msvc-x86_64-llvm-previous-cl3.0-offline
   run_windows_msvc_x86_64_llvm_latest_cl3_0_offline:
-
     runs-on: windows-2019
-
+    # do not run as PR job due to windows install being more complex
```
hvdijk, line 205: Does this mean it doesn't currently work, or it would cause PRs to take too long to get tested?

.github/workflows/planned_testing_caller_19.yml
```diff
@@ -0,0 +1,24 @@
...
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/planned_testing_caller_19.yml'
+  schedule:
+    # Run Tues, Thurs Sat at 7pm
+    - cron: '0 19 * * 0,2,4,6'
```
hvdijk, line 8: The comment doesn't match the `cron` line.

.github/workflows/run_ock_internal_tests.yml
coldav: I'll improve the wording, it is purely because we hadn't got it working right for install.

.github/actions/setup_build/action.yml
coldav: we don't have access to the secrets at action level, I'm moving towards passing it down, I've not yet updated everywhere to do it that way.

.github/workflows/create_llvm_artefacts.yml
```diff
@@ -0,0 +1,113 @@
+# Workflow for creating llvm artefacts
+name: create llvm artefacts
+on:
+  workflow_call:
+    inputs:
+      llvm_branch:
+        required: false
+        type: string
+        default: 'release/19.x'
```
hvdijk, line 9: I think this is a risky thing to have as a default, it makes it so that if we call this with `llvm_version: 20` and no `llvm_branch`, it will look valid but it will silently do the wrong thing. Could we just drop the default or are there currently calls that rely on it?

coldav: I can drop the defaults.

alan-forbes-cp (approved):

PR #701, opened 2025-03-03
SYCL CTS CI for riscv64 qemu

alan-forbes-cp:
# Overview

Add SYCL CTS tests for riscv64 qemu

# Reason for change

Add remaining architecture, plus qemu support.

# Description of change

* Update `planned_testing` SYCL CTS build/run with support for riscv64 plus correct container image.
* Add SYCL CTS override .csv file for riscv64.
* Update DPCPP workflow for riscv64, cross-build options and configuration (.cfg) files.
* Expand SYCL CTS build to include support for riscv64 qemu
* Expand SYCL CTS run to include qemu support (`$PREPEND_PATH` var)

* riscv64 run is green
* This PR is fully tested for, and merge-able for, all 3 architectures
* A number of temporary updates were added to facilitate development testing. These will be removed prior to final merge.


.github/workflows/planned_testing.yml
```diff
@@ -305,18 +306,18 @@ jobs:
           download_dpcpp_artefact: ${{ inputs.download_dpcpp_artefact }}
 
   build_sycl_cts:
-    needs: [workflow_vars, build_icd, build_dpcpp_native]
+    needs: [workflow_vars, build_icd, build_dpcpp_native, build_dpcpp_cross]
```
alan-forbes-cp, line 309: There's a slight gotcha here in that `build_dpcpp_cross` is now only needed for riscv64 (true also for SYCL CTS run job below).
If riscv64 is excluded from the `target_list` and this setting is retained then the SYCL CTS build job will not fire.
However, this is not currently an issue for production. Comments welcome.

coldav: yes it's a bit icky the dependencies, we could either fire build_dpcc_cross regardless (but do nothing if not needed) or we use the json method we've used elsewhere.

alan-forbes-cp: Closing this one on the basis that production is not impacted and future dev reqs are a separate issue (and unknown).

.github/workflows/planned_testing.yml
```diff
@@ -305,18 +306,18 @@ jobs:
           download_dpcpp_artefact: ${{ inputs.download_dpcpp_artefact }}
 
   build_sycl_cts:
```
coldav, line 308: You'll need to split this along with the others once the sycl-cts split has gone in.

.github/actions/do_build_dpcpp/action.yml
```diff
@@ -45,27 +61,29 @@ runs:
...
-        --host-target="X86;AArch64;RISCV"
-        --native_cpu
-        --llvm-external-projects=lld
-        --cmake-opt=-DNATIVECPU_USE_OCK=ON
-        --cmake-opt=-DLLVM_ENABLE_ZLIB=OFF
-        --cmake-opt=-DLLVM_ENABLE_ZSTD=OFF
+      run: |
+        CROSS_OPTS=""
+        if [[ "${{inputs.target}}" =~ .*riscv64.* ]] ; then
+          CROSS_OPTS="--cmake-opt=-DCMAKE_TOOLCHAIN_FILE=${{steps.calc_vars.outputs.toolchain}} --cmake-opt=-DLLVM_HOST_TRIPLE=${{steps.calc_vars.outputs.arch}}-unknown-linux-gnu --cmake-opt=-DLLVM_NATIVE_TOOL_DIR=$GITHUB_WORKSPACE/llvm/build/x86_64-linux/install/bin"
```
coldav, line 67: This giant line is a bit ugly.

.github/actions/do_build_dpcpp/action.yml
```diff
@@ -45,27 +61,29 @@ runs:
...
+        fi
+        cd llvm
+        set -x
+        python3 buildbot/configure.py -o build/${{steps.calc_vars.outputs.arch}}-linux --host-target="X86;AArch64;RISCV" --native_cpu \
```
coldav, line 71: if we move towards conditional steps then we could avoid all the \ and this would be just one step.

.github/actions/do_build_sycl_cts/action.yml
```diff
@@ -35,18 +35,50 @@ runs:
...
       uses: actions/download-artifact@v4
       with:
-        name: dpcpp_${{inputs.target}}
-        path: install_dpcpp
+        name: dpcpp_host_x86_64_linux
+        path: dpcpp/x86_64-linux/install
 
-    - name: unpackage dpc++ artifacts # package/unpackage avoids known 'permissions loss' issue
-      if: contains(inputs.download_sycl_cts_artefact, inputs.target) != true
+    - name: unpackage native dpc++ artifact (for x86_64 & riscv64) # package/unpackage avoids known 'permissions loss' issue
```
coldav, line 46: since you are rejigging this perhaps you could use my download/upload artifact action which does this for you?

PR #709, opened 2025-03-11
Use find_package for locating LLVM/Clang/LLD.

hvdijk:
# Overview

Use find_package for locating LLVM/Clang/LLD.

# Reason for change

This means that when LLVM is installed in a location where it can be picked up automatically, users do not need to set `CA_LLVM_INSTALL_DIR`.

# Description of change

points, etc.*

This does not work for host builtins, but we do not need host builtins, we just have it as an example. We can potentially add that back in in a future version if we have a need for it.

# Checklist


modules/compiler/riscv/CMakeLists.txt
```diff
@@ -48,7 +48,8 @@ if (LLVMRISCVCODEGEN)
 
     if(NOT TARGET compiler-linker-utils)
       set(error "compiler-riscv-utils requires compiler-linker-utils")
-      if(NOT EXISTS "${CA_LLVM_INSTALL_DIR}/lib${CA_LLVM_LIBDIR_SUFFIX}/cmake/lld/LLDConfig.cmake")
+      find_package(LLD REQUIRED CONFIG HINTS ${CA_LLVM_INSTALL_DIR})
```
hvdijk, line 51: This should not have `REQUIRED`.

coldav (approved):

PR #710, opened 2025-03-11
DRAFT: overload creation helper scripts

coldav:
# Overview

and consequences.*

# Reason for change

# Description of change

points, etc.*

# Checklist


.github/scripts/create_overload_csv.py
```diff
@@ -0,0 +1,92 @@
...
+# under the License.
+#
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+"""Create combined overload file based on a list of keys and a directory, using the keys
+to combine together to create possible filename combinations, all starting with overload_"""
+
+import argparse
+import os
+
+def get_args():
```
alan-forbes-cp, line 22: Is it worth adding an option to print the final output list to the log prior to run?
Or add it as part of verbose?

coldav: maybe - but it can get quite big. I suppose adding to verbose would be okay.

coldav: The final output is the combined csv file. The file_list from create_combos wll have all the filenames it will consider for each directory (but not checked for existance).

.github/workflows/planned_testing_caller.yml
```diff
@@ -30,14 +30,14 @@ jobs:
...
       ock: true
-      test_tornado: true
+      test_tornado: false
       test_sycl_cts: true
       test_opencl_cts: true
       # Have a pull request setting which can be used to test the flow as best as possible
       # in a reasonable time
 
       # The following can be used to download from a previous workflow run (change id)
-      # download_ock_artefact: host_x86_64_linux=12915462445;host_aarch64_linux=12915462445
```
coldav, line 40: will remove changes to this file before merging.

scripts/testing/create_override_csv.py
```diff
@@ -0,0 +1,132 @@
...
+    Search through the file list and directories to find files in the file_list.
+    Concat them if found onto the output_file.
+    """
+    for f in file_list:
+        for d in dirs:
+            filename = os.path.join(d, f)
+            if os.path.exists(filename):
+                if verbose > 0:
+                    print("Found: ", filename)
+                with open(filename, "r") as fp:
```
hvdijk, line 107: Apologies if the syntax isn't quite right yet but...
```suggestion
            try:
                fp = open(filename, "r")
            except FileNotFoundError:
                continue
            with fp:
                if verbose > 0:
                    print("Found: ", filename)
```

scripts/testing/create_override_csv.py
```diff
@@ -0,0 +1,132 @@
...
+
+# Create the list of combinations based on the keys passed in
+make_combos(args.key, "override", file_list, 1)
+
+# flatten list of lists
+new_list = [f for fs in file_list for f in fs]
+
+# Open the output file and concat in order.
+try:
+    f = open(args.output, "w+")
```
hvdijk, line 128:
```suggestion
    with open(args.output, "w+") as f:
```

scripts/testing/create_override_csv.py
```diff
@@ -0,0 +1,132 @@
...
+        keys      -- List of keys to combine, suitable for the base
+        base      -- base filename to extend
+        file_list -- List of lists of filenames, one list for each level
+        level     -- Current level
+
+    We want to create combinations of keys based on maintaining the order passed
+    in and combining with any keys which are beyond them in the list. Also we
+    want to make combinations of different numbers of keys each at a distinct
+    level. e.g. if we have keys one,two,three, we'd have keys L 1: one, two,
+    three, L 2: one_two, one_three, two_three L 3: one_two_three. make_combos
```
hvdijk, line 76: I think this means we can, depending on how we call this, have either e.g. `override_opencl_x86_64.csv` or `override_x86_64_opencl.csv`, but not both, right? That's probably fine but given that it silently ignores files that have the wrong name, I think we should be very clear about what we expect.

coldav: yes the order is important - what are you suggesting? I probably could make it so it considers all orders or perhaps you just mean make it clearer in the documentation?

hvdijk (approved):

PR #711, opened 2025-03-12
Update sycl-cts.csv, move to a more general location.

hvdijk:
# Overview

Update sycl-cts.csv, move to a more general location.

# Reason for change

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/scripts/testing/sycl-cts.csv
+++ b/scripts/testing/sycl-cts.csv
@@ -164,8 +164,7 @@ SYCL_CTS,test_has_kernel_bundle_zero_kernels_ctx_only_and_ctx_and_dev
 SYCL_CTS,test_header
 SYCL_CTS,test_hierarchical hierarchical_functor
 SYCL_CTS,test_hierarchical hierarchical_id
-# https://github.com/intel/llvm/issues/13574
-#SYCL_CTS,test_hierarchical hierarchical_implicit_barriers
+SYCL_CTS,test_hierarchical hierarchical_implicit_barriers
 SYCL_CTS,test_hierarchical hierarchical_implicit_conditional
 SYCL_CTS,test_hierarchical hierarchical_implicit_reduce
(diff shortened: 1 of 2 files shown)


scripts/testing/sycl-cts.csv
```diff
@@ -164,8 +164,7 @@ SYCL_CTS,test_has_kernel_bundle_zero_kernels_ctx_only_and_ctx_and_dev
 SYCL_CTS,test_header
 SYCL_CTS,test_hierarchical hierarchical_functor
 SYCL_CTS,test_hierarchical hierarchical_id
-# https://github.com/intel/llvm/issues/13574
-#SYCL_CTS,test_hierarchical hierarchical_implicit_barriers
+SYCL_CTS,test_hierarchical hierarchical_implicit_barriers
```
coldav, line 167: We have a https://github.com/uxlfoundation/oneapi-construction-kit/tree/main/.github/sycl_cts directory - maybe move it there?

.github/workflows/planned_testing_caller.yml
```diff
@@ -30,9 +30,9 @@ jobs:
                       "host_refsi_linux",
                       "host_x86_64_windows" ]'
       ock: true
-      test_tornado: true
+      test_tornado: false
```
coldav, line 33: Adding a comment just as a placeholder to remove.

scripts/testing/sycl-cts.csv
hvdijk: I wanted to not have it in `.github` since we want to use this file not just for GitHub CI. I used `scripts/testing` since that is the location we were using internally, but even if we're okay with it being outside of `.github`, there may still be a better location for it than this one, I'm not sure.

coldav: okay, i don't feel too strongly about it, so lets try it here.

PR #714, opened 2025-03-13
Update patches.

hvdijk:
# Overview

Update patches.

# Reason for change

# Description of change

* Move patches to scripts/patches/ as they are not limited to GitHub CI.
* Update DPCPP-0002-SYCL-NativeCPU-Build-libclc-target-independently in response to upstream feedback.
* Drop DPCPP-0003-add-cl_khr_int64_extended_atomics as it is no longer needed.

# Checklist


.github/actions/do_build_dpcpp/action.yml
```diff
@@ -32,11 +32,15 @@ runs:
...
       run: |
-        cd llvm
-        git apply $GITHUB_WORKSPACE/.github/patches/DPCPP-0001-Fix-static_assert-with-funsigned-char.patch
-        git apply $GITHUB_WORKSPACE/.github/patches/DPCPP-0002-SYCL-NativeCPU-Build-libclc-target-independently.patch
-        git apply $GITHUB_WORKSPACE/.github/patches/DPCPP-0003-add-cl_khr_int64_extended_atomics.patch
-        git apply $GITHUB_WORKSPACE/.github/patches/DPCPP-0004-SYCL-NativeCPU-Fix-__clc_exp10.patch
+        git -C llvm log -1
+        for patch in $GITHUB_WORKSPACE/scripts/testing/patches/DPCPP-*.patch
+        do
+          if test -f "$patch"
```
coldav, line 38: why do you need to do the -f - doesn't the * list effectively mean it exists?

coldav (approved):

PR #716, opened 2025-03-14
Drop one DPCPP patch.

hvdijk:
# Overview

Drop one DPCPP patch.

# Reason for change

DPCPP-0001-Fix-static_assert-with-funsigned-char.patch has been merged (with one slight change) upstream so we no longer need to apply it.

# Description of change

points, etc.*

# Checklist


.github/workflows/planned_testing_caller.yml
```diff
@@ -41,3 +41,4 @@ jobs:
       # download_dpcpp_artefact: host_x86_64_linux=12915462445;host_aarch64_linux=12915462445
       # download_sycl_cts_artefact: host_x86_64_linux=12915462445;host_aarch64_linux=12915462445      
       pull_request: ${{ github.event_name == 'pull_request' }}
+
```
hvdijk, line 44: Reminder to revert this.

PR #719, opened 2025-03-17
DRAFT: runs-on: ubuntu-22.04-arm

alan-forbes-cp:
# Overview

and consequences.*

# Reason for change

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/.github/workflows/planned_testing.yml
+++ b/.github/workflows/planned_testing.yml
@@ -128,7 +128,7 @@ jobs:
 
     # risc-v needs ubuntu 24.04 so we get the latest qemu as well as how we
     # build llvm. Otherwise we choose ubuntu-22.04 (use a container for both for consistency).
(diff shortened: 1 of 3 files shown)


.github/workflows/planned_testing.yml
```diff
@@ -128,7 +128,7 @@ jobs:
 
     # risc-v needs ubuntu 24.04 so we get the latest qemu as well as how we
     # build llvm. Otherwise we choose ubuntu-22.04 (use a container for both for consistency).
-    runs-on: ${{ contains(matrix.target, 'host_aarch64') && 'cp-graviton' || 'cp-ubuntu-24.04' }}
```
coldav, line 131: Please close this PR, I will pass a parameter as part of my llvm change (getting too clogged up on runs)

alan-forbes-cp: Done

PR #728, opened 2025-03-19
Fix UnitCL warning llvm 20 on whitespace in a literal operator declaration

coldav:
# Overview

Fix UnitCL warning llvm 20 on whitespace in a literal operator declaration

# Reason for change

Warning stopped us building with clang-20. Fixed by removing whitespace


hvdijk:
The change itself is fine (though IMHO the function is a poor idea implemented poorly and I'd also be okay with a bigger change to remove it), but the commit is mistakenly attributed to me.

coldav:
I agree, and i've fixed the ownership, sorry about that. I'd rather leave it as it is. I did consider a wider change, but I've plumped for the smallest change that works.

PR #736, opened 2025-03-25
Address more clang & clang-tidy warnings.

hvdijk:
# Overview

Address more clang & clang-tidy warnings.

# Reason for change

# Description of change

* Add more const where we can.
* Use contains() rather than count() to check for existence.
* Stop looking up options through llvm::cl::getRegisteredOptions() when we can use their corresponding variables instead.
* Stop creating unused .blend PHI nodes.
* Enforce W^X in our loader.
* Use remove_if() rather than stable_partition() to remove released semaphores.
* Ensure driver::CompilerInfo is always initialized.

# Checklist

### DIFF
--- a/modules/compiler/vecz/source/transform/simplify_infinite_loop_pass.cpp
+++ b/modules/compiler/vecz/source/transform/simplify_infinite_loop_pass.cpp
@@ -74,39 +74,11 @@ PreservedAnalyses vecz::SimplifyInfiniteLoopPass::run(
     assert(AR.DT.verify() &&
            "SimplifyInfiniteLoopPass: Dominator Tree failed verification");
 
-    std::unordered_set<Instruction *> toBlend;
-    // Find all instructions used in `target` that may be defined after the
-    // infinite loop, for which adding the edge from the infinite loop to the
-    // return block may break the SSA form.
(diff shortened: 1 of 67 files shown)


modules/compiler/vecz/source/transform/simplify_infinite_loop_pass.cpp
```diff
@@ -74,41 +74,13 @@ PreservedAnalyses vecz::SimplifyInfiniteLoopPass::run(
     assert(AR.DT.verify() &&
            "SimplifyInfiniteLoopPass: Dominator Tree failed verification");
```
coldav, line 76: This seems quite a chunky change - can you explain why we no longer need this bit and just below the next loop?

hvdijk: It was doing nothing. This block was the pre-computation for the next deleted block, and the next deleted block was creating PHI nodes but then generating no uses of those PHI nodes, they could immediately be deleted again as dead code. I think it was an attempt to fix a particular problem with infinite loops that was then accidentally left in.

coldav (approved):

PR #738, opened 2025-03-26
Switch most jobs over to internal runners and split sycl-cts jobs

coldav:
# Overview
Switch most jobs over to internal runners and split sycl-cts jobs

# Reason for change

We were seeing some issue with the github hosted runners which meant that we could lose the whole run if anything went wrong.

# Description of change

 Reduce dependency by moving most jobs over to github runners and splitting sycl cts so we don't stall the x86 run.

### DIFF
--- a/.github/workflows/planned_testing.yml
+++ b/.github/workflows/planned_testing.yml
@@ -128,7 +128,7 @@ jobs:
 
     # risc-v needs ubuntu 24.04 so we get the latest qemu as well as how we
     # build llvm. Otherwise we choose ubuntu-22.04 (use a container for both for consistency).
-    runs-on: ${{ contains(matrix.target, 'host_aarch64') && 'cp-graviton' || 'cp-ubuntu-24.04' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/planned_testing_caller.yml
```diff
@@ -41,3 +41,4 @@ jobs:
       # download_dpcpp_artefact: host_x86_64_linux=12915462445;host_aarch64_linux=12915462445
       # download_sycl_cts_artefact: host_x86_64_linux=12915462445;host_aarch64_linux=12915462445      
       pull_request: ${{ github.event_name == 'pull_request' }}
+# Force a pull request run.
```
coldav, line 44: Remove this before merging.

coldav:
This does not change the ock pull tag which we agreed would be mandatory before switching to testing nativecpu, so this needs to be done first (possible reason for splitting dpcpp build workflow job it as it may fail sometimes).
(I could possibly accept this going in then another PR to resolve that).

alan-forbes-cp (approved):

PR #742, opened 2025-03-31
Add SYCL CTS native_cpu aarch64 to CI 

alan-forbes-cp:
# Overview

Add SYCL CTS native_cpu aarch64 to CI

# Reason for change

Similar functionality is already in place for x86_64; riscv64 is next.

# Description of change

- add SYCL_CTS aarch64 build & run jobs for both `opencl` and `native_cpu`.
- add a check that arch-specific jobs only fire when the arch concerned is included in `target_list`.
- fix `-k` command line args to `create_override_csv.py` script.
- update override files as required - including for opencl.

- At time of writing there's a full test running for both x86_64 and aarch64 to verify the new override set.
- Note that there is a change to `override_opencl_host_aarch64_linux.csv` which previously wasn't marked MayFail anywhere. So this change adds a new OpenCL MayFail. The change is included in this PR, but is called out so that reviewers know to review this with extra care.


.github/workflows/planned_testing_caller.yml
```diff
@@ -23,22 +23,24 @@ jobs:
     if: github.repository == 'uxlfoundation/oneapi-construction-kit' || github.event_name != 'schedule'
     uses: ./.github/workflows/planned_testing.yml
     with:
+      #target_list: '[ "host_x86_64_linux",
```
alan-forbes-cp, line 26: Restore original settings in this file following final test and before merge.

.github/sycl_cts/override_native_cpu_host_x86_64_linux.csv
```diff
@@ -0,0 +1 @@
+SYCL_CTS,test_kernel "Behavior of kernel attribute*",XFail
```
hvdijk, line 1: Does this really still fail on x86_64?

.github/sycl_cts/override_opencl_host_aarch64_linux.csv
```diff
@@ -1,2 +1,3 @@
 SYCL_CTS,test_atomic_ref_stress "sycl::atomic_ref atomicity *",Mayfail
 SYCL_CTS,test_atomic_ref "sycl::atomic_ref fetch_add()/fetch_sub() test. *",Mayfail
+SYCL_CTS,test_reduction reduction_without_identity_param_core --allow-running-no-tests,Mayfail
```
coldav, line 3: What's this got to do with native_cpu? Are we sure this is a MayFail?

.github/sycl_cts/override_native_cpu_host_x86_64_linux.csv
alan-forbes-cp: It does.
https://github.com/uxlfoundation/oneapi-construction-kit/actions/runs/14195919883/job/39835776903#step:4:17444

.github/sycl_cts/override_opencl_host_aarch64_linux.csv
alan-forbes-cp: This PR splits the existing aarch64 jobs into opencl & native_cpu.
Example of fail: https://github.com/uxlfoundation/oneapi-construction-kit/actions/runs/14191939348/job/39758673805#step:4:13822

hvdijk: Right, but it previously wasn't marked MayFail anywhere, was it? So this change isn't to preserve the existing OpenCL MayFail, this change is to add a new OpenCL MayFail. I think Colin is right that this isn't related to the NativeCPU changes. If the change is correct, then I guess it makes sense to me to just include it in this PR, but it should be called out so that we know to review this with extra care.

alan-forbes-cp: Comment added to that effect in the PR description.

.github/sycl_cts/override_native_cpu_host_x86_64_linux.csv
alan-forbes-cp: As suggested, this test has been moved back to the override_native_cpu.csv but marked as MayFail instead.

coldav (approved):

PR #745, opened 2025-04-03
Set minimum cmake version via cmake options to 3.5

coldav:
# Overview

Windows build was complaining because the ICD loader minumum version was < 3.5. This fixes this until they have updated that.


alan-forbes-cp (approved):

hvdijk (changes requested):
As mentioned internally, I don't think this is right. It's failing because we're building with CMake 4 an external project which has not been updated to be compatible with CMake 4. I think we should either ensure we build with CMake 3 for now, or do the work of making it compatible with CMake 4.

PR #751, opened 2025-04-04
[LLVM 21] Take address space into account for legality.

hvdijk:
# Overview

[LLVM 21] Take address space into account for legality.

# Reason for change

LLVM 21 allows for the possibility that some operations are legal only in some address spaces.

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/modules/compiler/vecz/source/vector_target_info.cpp
+++ b/modules/compiler/vecz/source/vector_target_info.cpp
@@ -21,6 +21,7 @@
 #include <llvm/Target/TargetMachine.h>
 #include <llvm/TargetParser/Triple.h>
 #include <multi_llvm/intrinsic.h>
+#include <multi_llvm/target_transform_info.h>
 #include <multi_llvm/vector_type_helper.h>
 
 #include "debugging.h"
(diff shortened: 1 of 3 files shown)


modules/compiler/vecz/source/vector_target_info.cpp
```diff
@@ -46,23 +46,56 @@ Value *applyEVLToMask(IRBuilder<> &B, Value *EVL, Value *Mask) {
...
+auto isLegalMaskedStoreImpl(const TargetTransformInfo &TTI, Type *Ty,
+                            unsigned Alignment, unsigned AddrSpace)
+    -> decltype(TTI.isLegalMaskedStore(Ty, Align(Alignment), AddrSpace)) {
+  return TTI.isLegalMaskedStore(Ty, Align(Alignment), AddrSpace);
+}
+#endif
+
+bool isLegalMaskedLoad(const TargetTransformInfo &TTI, Type *Ty,
+                       unsigned Alignment, unsigned AddrSpace) {
+  return isLegalMaskedLoadImpl(TTI, Ty, Alignment, AddrSpace);
```
coldav, line 84: How does this work for < 21 - is it not undefined?
Also should this kind of logic be in multi_llvm?

hvdijk: `isLegalMaskedLoadImpl` is overloaded; the version that works for LLVM 19 and 20, and older snapshots of LLVM 21, is always defined but silently ignored on newer LLVM 21.

This could be in `multi_llvm` but we've generally only done that when we want something re-used in multiple places, have we not?

coldav: ah fair enough I couldn't see in the change that it existed elsewhere.
multi_llvm was just a thought, but it's where we've done this new method of supporting functions in this way, so it might benefit from consistency in one place.
Up to you though. I approve regardless.

hvdijk: Alright, done.

coldav (approved):

PR #757, opened 2025-04-08
Remove all conditional LLVM code for LLVM 18

coldav:
# Overview

Remove all conditional LLVM code for LLVM 18

# Reason for change

LLVM 18 is no longer supported.


hvdijk:
We failed to update the lit tests the last time we removed support for previous versions, should we do that now? I mean the `CHECK-LT19`/`CHECK-GE19` etc. in lit tests

hvdijk (approved):
It's fine to leave the lit test updates for older LLVM versions to a separate PR, we should have cleaned that up earlier. But for the `CHECK-LT19` tests, can you also take out the `-GE19`? If you have this commit checked out, ``sed -i -e s/-GE19// `git diff-tree -r --no-commit-id --name-only @` `` will do it. Approved with that change, or approved as is if you feel that that is also better left to the follow-up PR.

PR #760, opened 2025-04-09
Do not vectorize NoInline kernels.

hvdijk:
# Overview

Do not vectorize NoInline kernels.

# Reason for change

A recent LLVM 21 change prevents us from being able to vectorize NoInline kernels.

# Description of change

It was probably never a good idea to vectorize them anyway: it is not clear whether the vectorization we perform is something that a user who writes `__attribute__((noinline))` would want. Therefore, explicitly reject vectorization of NoInline functions and restore affected tests to their original version that permitted them to not be vectorized.

# Checklist

### DIFF
--- a/modules/compiler/vecz/source/analysis/vectorizable_function_analysis.cpp
+++ b/modules/compiler/vecz/source/analysis/vectorizable_function_analysis.cpp
@@ -47,7 +47,7 @@ namespace {
 /// @param[in] I Instruction to check for vectorizability.
 /// @param[in] Ctx VectorizationContext for BuiltinInfo.
 ///
-/// @return true if I can be vectorized.
+/// @return true if I can be vectorized, false otherwise.
 bool canVectorize(const Instruction &I, const VectorizationContext &Ctx) {
   // Certain instructions just cannot appear.
   switch (I.getOpcode()) {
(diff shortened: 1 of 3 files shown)


modules/compiler/vecz/source/analysis/vectorizable_function_analysis.cpp
```diff
@@ -119,13 +126,6 @@ VectorizableFunctionAnalysis::Result VectorizableFunctionAnalysis::run(
...
 
-  // Do not vectorize functions with the OptNone attribute
-  if (F.hasFnAttribute(Attribute::OptimizeNone)) {
-    res.canVectorize = false;
-    return res;
-  }
-
-  res.failedAt = canVectorize(F, Ctx);
-  res.canVectorize = !res.failedAt;
+  res.canVectorize = canVectorize(F, Ctx);
```
coldav, line 129: I can't see anywhere that .failedAt is used at after you've removed this - could you also remove that too?

coldav (approved):

modules/compiler/vecz/source/analysis/vectorizable_function_analysis.cpp
hvdijk: It was already not used anywhere and so this PR removes the field entirely.

coldav: ah sorry, my bad.

coldav (approved):

PR #762, opened 2025-04-10
[LLVM 19] Work around loop vectorizer issue.

hvdijk:
# Overview

[LLVM 19] Work around loop vectorizer issue.

# Reason for change

The loop vectorizer gets confused by having pointers to distinct address spaces that are derived from the same object. This has been fixed for LLVM 20, but as long as we support LLVM 19, work around it there.

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/modules/compiler/compiler_pipeline/source/remove_address_spaces_pass.cpp
+++ b/modules/compiler/compiler_pipeline/source/remove_address_spaces_pass.cpp
@@ -1,0 +1,146 @@
+// Copyright (C) Codeplay Software Limited
+//
+// Licensed under the Apache License, Version 2.0 (the "License") with LLVM
+// Exceptions; you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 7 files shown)


modules/compiler/compiler_pipeline/source/remove_address_spaces_pass.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+#include <compiler/utils/remove_address_spaces_pass.h>
+#include <llvm/IR/Function.h>
+#include <llvm/IR/IRBuilder.h>
+#include <llvm/IR/Instructions.h>
+
+using namespace llvm;
+
+#if LLVM_VERSION_LESS(20, 0)
+static Type *removeAddressSpaces(Type *T) {
```
hvdijk, line 25: This function can be shorter and simpler.

uwedolinsky (approved):

PR #763, opened 2025-04-10
CI ock_external workflow update for stand-alone riscv64 

alan-forbes-cp:
# Overview

Allow riscv64 variant of the ock_external workflow to be run independently from other architectures.

# Reason for change

Cross-compile job dependencies were wrong for riscv64 when run alone, leaving a loophole which failed the ock_external workflow. This fix restores "run alone" support for riscv, bringing it into line with other architectures.

# Description of change

- Refactor the `build_dpcpp_native_*` ock_external  jobs to allow riscv64 to be run independently of x86_64. 
- Update downstream jobs `needs:` to indicate the new dependencies involved.

- Various combinations of architectures - run-alone and otherwise - have been tested to verify the fix.
- Test results match those currently produced on `main`.
- Some planned_testing_caller development/testing scaffolding remains in place to reduce test time and will be removed prior to final merge.

### DIFF
--- a/.github/workflows/planned_testing_caller_19.yml
+++ b/.github/workflows/planned_testing_caller_19.yml
@@ -23,6 +23,7 @@ jobs:
       # We can set ock, test_sycl_cts etc here optionally if this is used as a pull request.
       # Any with parameters below this is intended for local testing and should not be merged.
       target_list: '[ "host_x86_64_linux",
+                      "host_aarch64_linux",
                       "host_riscv64_linux" ]'
       # ock: true
       # native_cpu: true
--- a/.github/workflows/planned_testing_caller_19.yml
+++ b/.github/workflows/planned_testing_caller_19.yml
@@ -22,13 +22,13 @@ jobs:
       use_llvm_github_cache: true
       # We can set ock, test_sycl_cts etc here optionally if this is used as a pull request.
       # Any with parameters below this is intended for local testing and should not be merged.
-      # target_list: '[ "host_x86_64_linux" ]'
+      target_list: '[ "host_riscv64_linux" ]'
       # ock: true
       # native_cpu: true
-      # test_tornado: true
+      test_tornado: false
       # test_sycl_cts: true
-      # test_opencl_cts: true
-      # run_internal: true
+      test_opencl_cts: false
+      run_internal: false
       # run_external: true
       # build_llvm: true
 
--- a/.github/workflows/planned_testing_caller_19.yml
+++ b/.github/workflows/planned_testing_caller_19.yml
@@ -22,15 +22,13 @@ jobs:
       use_llvm_github_cache: true
       # We can set ock, test_sycl_cts etc here optionally if this is used as a pull request.
       # Any with parameters below this is intended for local testing and should not be merged.
-      target_list: '[ "host_x86_64_linux",
-                      "host_aarch64_linux",
-                      "host_riscv64_linux" ]'
+      # target_list: '[ "host_x86_64_linux" ]'
       # ock: true
       # native_cpu: true
-      test_tornado: false
+      # test_tornado: true
       # test_sycl_cts: true
-      test_opencl_cts: false
-      run_internal: false
+      # test_opencl_cts: true
+      # run_internal: true
       # run_external: true
       # build_llvm: true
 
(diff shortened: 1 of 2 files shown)


.github/workflows/planned_testing_caller_19.yml
```diff
@@ -22,13 +22,15 @@ jobs:
       use_llvm_github_cache: true
       # We can set ock, test_sycl_cts etc here optionally if this is used as a pull request.
       # Any with parameters below this is intended for local testing and should not be merged.
-      # target_list: '[ "host_x86_64_linux" ]'
+      target_list: '[ "host_x86_64_linux",
```
alan-forbes-cp, line 25: Restore file defaults prior to final merge.

.github/workflows/run_ock_external_tests.yml
```diff
@@ -265,38 +265,54 @@ jobs:
...
-        target: ${{ fromJson(inputs.target_list) }}
-        exclude: ${{ fromJson(needs.workflow_vars.outputs.matrix_only_linux_x86_64_aarch64) }}
+    runs-on: 'ubuntu-22.04'
+    container:
+      image: 'ghcr.io/uxlfoundation/ock_ubuntu_22.04-x86-64:latest'
+      volumes:
+        - ${{github.workspace}}:${{github.workspace}}
 
-    runs-on: ${{ contains(matrix.target, 'host_aarch64') && 'ubuntu-22.04-arm' || (contains(matrix.target, 'host_riscv64') && 'ubuntu-24.04' || 'ubuntu-22.04') }}
+    if: inputs.test_sycl_cts && ( contains(inputs.target_list, 'host_x86_64_linux') || contains(inputs.target_list, 'host_riscv64_linux') )
```
hvdijk, line 276: We also want to cross-compile for AArch64 even if we currently don't, maybe we should just change this to `if: inputs.test_sycl_cts` (without a `target_list` check) now?

alan-forbes-cp: We could do - however I'm inclined to leave it as it is for now as that functionality is not yet on the Github CI task list.

.github/workflows/run_ock_external_tests.yml
```diff
@@ -265,38 +265,54 @@ jobs:
...
+          os: 'ubuntu'
           token: ${{ secrets.GITHUB_TOKEN }}
       - name: build dpc++ artefact
         uses: ./.github/actions/do_build_dpcpp
         with:
-          target: ${{ matrix.target }}
+          target: host_aarch64_linux
           download_dpcpp_artefact: ${{ inputs.download_dpcpp_artefact }}
 
   build_dpcpp_cross:
```
coldav, line 314: This is a bad name, if it's just for riscv 64, also there is not matrix, why we do we have all these fail-fasts?

alan-forbes-cp: I can change the name to `build_dpcpp_riscv64` if that sounds better?
Re. matrix/fail-fast ... this came up in a PR review a while back (in fact I may have raised it myself). As I recall, the thinking was that in some contexts it would be useful to temporarily retain the matrix (and by implication the fail-fast when that was added) even when the `exclude:` resulted in only a single element at that time. The specific reasons are a little hazy ... possibly for consistency between jobs at a time when there was a significant amount of refactoring?
Worth noting that the `build_run_tornado` job also uses that model so perhaps another ticket is needed to do a general refactor?
I hope I've understood your comment here.

coldav: I meant it wasn't a matrix job, but i can't see it on the right side any more so maybe I was just confused.

coldav: build_dpcpp_riscv64 sounds fine.

alan-forbes-cp: Updated as described

.github/workflows/run_ock_external_tests.yml
```diff
@@ -265,38 +265,54 @@ jobs:
...
 
-    if: inputs.test_sycl_cts
+    if: inputs.test_sycl_cts && contains(inputs.target_list, 'host_aarch64_linux')
     steps:
       - name: Checkout repo
         uses: actions/checkout@v4
       - name: set up gh
         uses: ./.github/actions/setup_gh
         with:
-          os: ${{ contains( matrix.target, 'windows') && 'windows' || 'ubuntu' }}
```
hvdijk, line 290: What happens now if we try to build for Windows? Or was this something that we never used?

coldav: In theory this was required when it was a matrix (although we never tried it), but these are individual jobs now with specific targets and no matrix.

coldav (approved):

coldav (approved):

PR #766, opened 2025-04-14
CI artifact upload/download consolidation

alan-forbes-cp:
# Overview

Consolidated CI artifact upload/download functionality via amended local actions.

# Reason for change

Where upload/download involves tarring/untarring components (due to the known permissions loss issue) there are currently two models in play - one embedded in the jobs, the other using local actions. This change consolidates these using updated versions of the existing local actions.

# Description of change

- Point all uploads/downloads with tar at our local actions.
- Update actions with parameterised support for tar file naming and deletion. 
- Rationale for local upload/download action updates: 
  - by default the tar name (currently `tmp.tar`) is now the provided artifact name with a `.tar` suffix. This can be overridden by the provision of a full tar name as input param.
  - by default the tar file is now treated as a temporary job file and is deleted after both upload and download/extraction.  This can be overridden by the provision of a 'do not delete' flag as input param.
  - `name` and `path` action inputs are now `required`.

At time of writing, a final test run is on-going.
All development/test scaffolding will be restored to original settings prior to final merge.


.github/actions/do_build_dpcpp/action.yml
```diff
@@ -28,21 +28,31 @@ runs:
...
-
-    - name: unpackage native dpc++ artifacts for cross builds # package/unpackage avoids known 'permissions loss' issue
-      if: contains(inputs.download_dpcpp_artefact, inputs.target) != true && contains(inputs.target, 'host_riscv')
-      shell: bash
-      run: |
-        cd llvm/build/x86_64-linux/install
-        tar xf dpcpp.tar
-        rm dpcpp.tar
+        needs_tar: 'true'
+        run_id: ${{ github.run_id }}
```
coldav, line 55: We talked about naming it rather than tmp.tar - that would also allow it to use artifacts from the old runs.

.github/actions/download_artifact/action.yml
```diff
@@ -44,4 +44,5 @@ runs:
       run: |
         # Everything goes under the new path
         mkdir -p ${{ inputs.path }}
+        # ALAN TODO: use inputs: name rather than tmp here?
```
coldav, line 47: we also suggested deleting the tar file.

alan-forbes-cp: Tar files are now deleted by default with an override param if required.

.github/workflows/planned_testing_caller_19.yml
```diff
@@ -22,20 +22,23 @@ jobs:
       use_llvm_github_cache: true
       # We can set ock, test_sycl_cts etc here optionally if this is used as a pull request.
       # Any with parameters below this is intended for local testing and should not be merged.
-      # target_list: '[ "host_x86_64_linux" ]'
+      #target_list: '[ "host_x86_64_linux" ]'
+      target_list: '[ "host_x86_64_linux",
```
alan-forbes-cp, line 26: Restore all development/test scaffolding to original settings prior to final merge.

.github/actions/do_build_sycl_cts/action.yml
```diff
@@ -149,16 +125,16 @@ runs:
...
-        mkdir -p SYCL-CTS
+        mkdir -p SYCL-CTS/bin
         git config --global --add safe.directory $PWD        
-        gh run download $download_id -n sycl_cts_${{ inputs.sycl_device }}_${{ inputs.target }} -D SYCL-CTS
-        ls SYCL-CTS
-
+        gh run download $download_id -n sycl_cts_${{ inputs.sycl_device }}_${{ inputs.target }}
+        # Downloaded tar file for extraction is: sycl_cts_${{inputs.sycl_device}}_${{inputs.target}}.tar
+        tar xf sycl_cts_${{inputs.sycl_device}}_${{inputs.target}}.tar -C SYCL-CTS/bin
+        ls SYCL-CTS/bin
```
coldav, line 133: we could probably drop this line now.

coldav (approved):

.github/actions/upload_artifact/action.yml
```diff
@@ -3,34 +3,65 @@ description: upload an artifact, tarring first if necessary
...
 runs:
   using: "composite"
   steps:
+    - name: set tar name
+      id: set-tar-name
+      shell: bash
+      if: inputs.needs_tar == 'true'
+      run: |
+        echo Set tar file name ...
+        if [ -z "${{ inputs.tar_name }}" ] ; then
```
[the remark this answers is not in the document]
alan-forbes-cp: I think this is OK as a default. Tar files can be retained rather than deleted so the name used by the action should be unique.

.github/actions/do_build_sycl_cts/action.yml
alan-forbes-cp: There's a similar line (untouched by this PR) in the dpcpp build action.  Given its a minorr issue I'll delete both if it becomes annoying.

.github/actions/download_artifact/action.yml
```diff
@@ -4,26 +4,48 @@ description: Download an artifact, either through the action or using gh, depend
...
     description: 'name of artifact to download'
-    default: "none"
+    type: string
+    required: true
   path:
     description: 'path to download to - currently must be a directory if needs_tar is true'
-    default: "none"
+    type: string
+    required: true
+  needs_tar:
```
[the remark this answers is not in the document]
alan-forbes-cp: I moved both of the `required: true` inputs to the top of the list in both local actions.
I'm in two minds about `needs_tar:` as default. Not unreasonable, but we'd lose the fall-thru of calling the standard Github action (plus updated retention) if no additional params are given - also not unreasonable. I'm tempted to stick for now.

PR #769, opened 2025-04-15
[LLVM 21] Avoid calling spir_kernel functions.

hvdijk:
# Overview

[LLVM 21] Avoid calling spir_kernel functions.

# Reason for change

LLVM 21 no longer permits calls to spir_kernel functions, not even if it would be patched out by later passes.

# Description of change

points, etc.*

# Checklist


coldav:
Is FixUpCallingConvention required after this change? If not we should remove it, along with it's lit test, tuner setting etc.

hvdijk:
> Is FixUpCallingConvention required after this change? If not we should remove it, along with it's lit test, tuner setting etc.

It shouldn't be required and I agree we should remove it. I'll do that in a subsequent PR.

PR #770, opened 2025-04-16
LLVM 21: Update more tests for captures(none).

hvdijk:
# Overview

LLVM 21: Update more tests for captures(none).

# Reason for change

# Description of change

points, etc.*

# Checklist


coldav (approved):

hvdijk:
> It would be nice to note what the llvm 21 change is that requires this.

This is a follow-up to #676; the reason is the same, it's just that we missed some tests initially. I'll update the commit message and PR description to say so.

PR #778, opened 2025-04-21
Add ock git tag for native_cpu

alan-forbes-cp:
# Overview

Set OCK Git tag for `native_cpu` dpc++ building

# Reason for change

We want to test against tip of ock for `native_cpu` builds.

# Description of change

Add new `ock_git_tag` input for dpc++ build job. (default: `github.sha`) and assign to OCK_GIT_TAG define for build configuration.

- Uses a work-around to allow SHAs to be specified (rather then branches or tags)

### DIFF
--- a/.github/actions/do_build_dpcpp/action.yml
+++ b/.github/actions/do_build_dpcpp/action.yml
@@ -9,7 +9,8 @@ inputs:
   ock_git_tag:
     description: 'dpcpp build config uses this ock Git tag.'
     type: string
-    default: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
+    default: ${{ github.event_name == 'pull_request' && github.event.pull_request.merge_commit_sha || github.sha }}
+    #default: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
     #default: ${{ github.sha }}
   ock_git_actor:
     description: 'dpcpp build config uses this ock Git tag.'
--- a/.github/actions/do_build_dpcpp/action.yml
+++ b/.github/actions/do_build_dpcpp/action.yml
@@ -13,7 +13,7 @@ inputs:
     #       with testing as appropriate to verify the update.
     #       If NOT a 'pull_request' event (i.e. a 'schedule' event instead) 'github_sha' is the latest
     #       commit on the default branch.
-    default: ${{ github.event_name == 'pull_request' && 'main' || github.sha }}
+    default: ${{ github.event_name == 'pull_request' && github.sha || github.sha }}
   download_dpcpp_artefact:
     description: 'download ock artefact rather than building, of form <target>=id;<target2=id2>. Special value of download_release applies to all targets.'
     type: string
@@ -76,7 +76,7 @@ runs:
         cd llvm
         set -x
         python3 buildbot/configure.py -o build/${{steps.calc_vars.outputs.arch}}-linux \
-                -DOCK_GIT_TAG=${{ inputs.ock_git_tag }} \
+                -DOCK_GIT_TAG="${{ inputs.ock_git_tag }};GIT_CONFIG;remote.origin.fetch=+refs/pull/*:refs/pull/*" \
                 --host-target="X86;AArch64;RISCV" --native_cpu \
                 --llvm-external-projects=lld --cmake-opt=-DNATIVECPU_USE_OCK=ON \
                 $CROSS_OPTS \
--- a/.github/actions/do_build_dpcpp/action.yml
+++ b/.github/actions/do_build_dpcpp/action.yml
@@ -69,8 +69,8 @@ runs:
         fi
         cd llvm
         set -x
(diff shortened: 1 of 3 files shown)


.github/actions/do_build_dpcpp/action.yml
```diff
@@ -4,6 +4,16 @@ description: build dpc++
 inputs:
   target:
     description: 'target architecture'
+    type: boolean
+    default: false
+  ock_git_tag:
+    description: 'dpcpp build config uses this ock Git tag.'
+    type: string
+    default: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.sha || github.sha }}
```
coldav, line 12: does this really work? I thought you'd need to do it at the calling level.

.github/actions/do_build_dpcpp/action.yml
```diff
@@ -4,6 +4,16 @@ description: build dpc++
...
+  ock_git_actor:
+    description: 'dpcpp build config uses this ock Git tag.'
+    type: string
+    default: ${{ github.event_name == 'pull_request' && github.actor || 'uxlfoundation' }}
```
coldav, line 16: We think once we have a pull request it should just use the sha, which means it is the merge, we don't want it coming from a different repo, so I think it's okay for us not to have OCK_GIT_REPO being set.

.github/actions/do_build_dpcpp/action.yml
```diff
@@ -4,6 +4,12 @@ description: build dpc++
 inputs:
   target:
     description: 'target architecture'
+    type: boolean
+    default: false
+  ock_git_tag:
+    description: 'If non-empty string, dpcpp build config uses this ock Git tag.'
+    type: string
+    default: ""
```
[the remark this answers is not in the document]
alan-forbes-cp: Default is: `github.sha`

.github/actions/do_build_dpcpp/action.yml
alan-forbes-cp: Now outdated - but it does work.

.github/actions/do_build_dpcpp/action.yml
alan-forbes-cp: SHAs can now be used - in conjunction with the OCK_GIT_TAG work-around to allow them. So no additional repo spec needed.

.github/actions/do_build_dpcpp/action.yml
```diff
@@ -4,6 +4,14 @@ description: build dpc++
 inputs:
   target:
     description: 'target architecture'
+  ock_git_tag:
+    description: 'dpcpp build config uses this ock Git tag.'
+    type: string
+    # Note: 'github.sha' is good for both schedule and pull_request events. Although CMake will currently only
```
coldav, line 10: Maybe CMake FetchContent. Also slight preference to putting this info where it is used, since it's not overly relevant to its usage.

.github/workflows/run_ock_external_tests.yml
```diff
@@ -316,7 +316,7 @@ jobs:
           target: host_aarch64_linux
           download_dpcpp_artefact: ${{ inputs.download_dpcpp_artefact }}
 
-  build_dpcpp_riscv64:
+  build_dpcpp_cross_riscv64:
```
coldav, line 319: This seems an irrelevant change here and I think unnecessary

.github/actions/do_build_dpcpp/action.yml
```diff
@@ -152,6 +161,6 @@ runs:
     - name: upload dpcpp artifact
       uses: ./.github/actions/upload_artifact
       with:
-        name: dpcpp_${{inputs.target}}
+        name: dpcpp_${{ inputs.target }}
```
coldav, line 164: unnecessary change and doesn't use the same spacing as the one below, I think if we want to tidy up in this way, we should make a single PR.

.github/actions/do_build_dpcpp/action.yml
alan-forbes-cp: Folded some text into `description:` and moved the rest where it is used.

.github/actions/do_build_dpcpp/action.yml
alan-forbes-cp: removed

.github/workflows/run_ock_external_tests.yml
alan-forbes-cp: removed

.github/actions/do_build_dpcpp/action.yml
```diff
@@ -65,7 +69,11 @@ runs:
         fi
         cd llvm
         set -x
+        # Note: Although CMake will currently only handle branches and tags by default in OCK_GIT_TAG, other
```
coldav, line 72: nitpick: it's fetchcontent in cmake that causes the issue, but you can leave if you want.

coldav (approved):

.github/actions/do_build_dpcpp/action.yml
alan-forbes-cp: Done

PR #784, opened 2025-04-23
[vecz] Track barriers by ID.

hvdijk:
# Overview

[vecz] Track barriers by ID.

# Reason for change

We were tracking barriers by their index, which can (rarely) break when optimizations on vectorized kernels cause barriers to be optimized away before the work item loops pass runs.

# Description of change

All barriers have a unique ID at this point already, which we can use to track them instead.

# Checklist

### DIFF
--- a/modules/compiler/compiler_pipeline/source/work_item_loops_pass.cpp
+++ b/modules/compiler/compiler_pipeline/source/work_item_loops_pass.cpp
@@ -1623,13 +1623,14 @@ Function *compiler::utils::WorkItemLoopsPass::makeWrapperFunction(
   AllocaInst *nextID =
       entryIR.CreateAlloca(index_type, nullptr, "next_barrier_id");
 
-  SmallVector<BasicBlock *, 8> bbs;
-  const unsigned num_blocks = barrierMain.getNumSubkernels();
-  assert(!emitTail || barrierTail->getNumSubkernels() == num_blocks);
+  std::map<unsigned, BasicBlock *> bbs;
+  // The vectorized kernel has been further optimized and may have removed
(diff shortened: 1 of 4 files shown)


modules/compiler/compiler_pipeline/source/work_item_loops_pass.cpp
```diff
@@ -1623,13 +1623,14 @@ Function *compiler::utils::WorkItemLoopsPass::makeWrapperFunction(
...
 
-  SmallVector<BasicBlock *, 8> bbs;
-  const unsigned num_blocks = barrierMain.getNumSubkernels();
-  assert(!emitTail || barrierTail->getNumSubkernels() == num_blocks);
-
-  for (unsigned i = kBarrier_EndID; i <= num_blocks; i++) {
-    BasicBlock *bb = BasicBlock::Create(context, "sw.bb", new_wrapper);
-    bbs.push_back(bb);
+  std::map<unsigned, BasicBlock *> bbs;
+  // The vectorized kernel has been further optimized and may have removed
```
coldav, line 1627: Could you explain how we deal with that case? It's not quite clear.

hvdijk: We know since we specifically go out of our way to only optimize the vectorized kernels, and start the vectorized as a clone of the scalar kernel, any barrier in the vectorized kernel must also be present in the scalar kernel. Barriers that are present in the scalar kernel, but not in the vectorized kernel, we can assume to be unreachable because barriers have to be convergent: if one work item reaches a barrier, all work items must reach that barrier.

coldav (approved):

PR #788, opened 2025-04-24
Fix clc device selection on multi-device platform

dlopr:
If a platform reports multiple devices, they will all have the same name and --device selection will be ambiguous.

This adds a --device-idx cmd-line option to directly select by index.

### DIFF
--- a/modules/cargo/include/cargo/argument_parser.h
+++ b/modules/cargo/include/cargo/argument_parser.h
@@ -219,7 +219,7 @@ class argument {
     cargo::array_view<cargo::string_view> Choices;
   };
   union {
-    bool *Bool;
+    bool *Bool = {};
     cargo::string_view *Value;
     ChoiceT Choice;
     cargo::small_vector<cargo::string_view, 4> *Values;
--- a/source/cl/tools/clc/clc.cpp
+++ b/source/cl/tools/clc/clc.cpp
@@ -25,6 +25,7 @@
 
 #include <algorithm>
 #include <array>
+#include <charconv>
 #include <cstdio>
 #include <cstdlib>
 #include <fstream>
@@ -152,6 +153,7 @@ driver::driver()
       dry_run(false),
       input_file(),
       output_file(""),
+      device_idx(0),
       device_name_substring(""),
       cl_build_args(),
       strip_binary_header(false),
@@ -185,6 +187,9 @@ optional arguments:
                         output file path, defaults to the name of the last
                         input file "<input>.bin" if present, or "-" otherwise
                         to write to standard output
+  --device-idx index
+                        the index (1..N) of the device to select. Takes precedence
+                        over --device
   -d name, --device name
                         a substring of the device name to select, choose from:%s
   --list-devices        print the list of available devices and exit
@@ -274,6 +279,19 @@ result driver::parseArguments(int argc, char **argv) {
 
   CHECK(parser.add_argument({"-o", output_file}));
   CHECK(parser.add_argument({"--output", output_file}));
+
+  CHECK(parser.add_argument({
+      "--device-idx",
+      [](auto &&...) { return parse::INCOMPLETE; },
+      [this](cargo::string_view str_value) {
+        auto [ptr, ec] =
+            std::from_chars(str_value.begin(), str_value.end(), device_idx);
+        if ((ptr != str_value.end()) || (device_idx == 0)) {
+          return parse::INVALID;
+        }
+        return parse::COMPLETE;
+      },
+  }));
 
   CHECK(parser.add_argument({"-d", device_name_substring}));
   CHECK(parser.add_argument({"--device", device_name_substring}));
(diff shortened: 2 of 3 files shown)


coldav:
Hi, it would be good to get some context. Why would the devices have the same names?

dlopr:
> Hi, it would be good to get some context. Why would the devices have the same names?

Maybe they shouldn't. The problem is that `hal::hal_t::device_get_info` returns a pointer for some reason. So an implementator is forced to keep that object alive. I guess an option would be to keep a vector of infos, one per device even if they only differ in the name, but that feels rather silly and I thought this wouldn't hurt.

For more context, see the [hal_refsi_tutorial example](https://github.com/uxlfoundation/oneapi-construction-kit/blob/d976cc141beddcfc91c7288fe3d1376f7dd814c7/examples/hals/hal_refsi_tutorial/source/hal_main.cpp#L42). It just keeps an instance of `hal_device_info_t` and returns it regardless of the index.

One extra weirdness is that the fact that the name is a `const char*` means that if it were something dynamic based on the index, one would need to manage the lifetime of that string separately.

I'd think that method should return by value and the returned object should use `std::string` for string fields, but changing that would be a much bigger change.

hvdijk:
> For more context, see the [hal_refsi_tutorial example](https://github.com/uxlfoundation/oneapi-construction-kit/blob/d976cc141beddcfc91c7288fe3d1376f7dd814c7/examples/hals/hal_refsi_tutorial/source/hal_main.cpp#L42). It just keeps an instance of `hal_device_info_t` and returns it regardless of the index.

It also sets `hal_info.num_devices = 1;` though, so it knows in `device_get_info(uint32_t index)` that `index` must necessarily be zero.

The assumption that the name uniquely identifies the device is kind of baked into our build system. We call `clc` during our build with the appropriate `-d` option that we assume is enough. If we make it so that that assumption is no longer valid, we would need to make bigger changes, I am not sure what that would need to look like.

dlopr:
> > For more context, see the [hal_refsi_tutorial example](https://github.com/uxlfoundation/oneapi-construction-kit/blob/d976cc141beddcfc91c7288fe3d1376f7dd814c7/examples/hals/hal_refsi_tutorial/source/hal_main.cpp#L42). It just keeps an instance of `hal_device_info_t` and returns it regardless of the index.
> 
> It also sets `hal_info.num_devices = 1;` though, so it knows in `device_get_info(uint32_t index)` that `index` must necessarily be zero.
> 
> The assumption that the name uniquely identifies the device is kind of baked into our build system. We call `clc` during our build with the appropriate `-d` option that we assume is enough. If we make it so that that assumption is no longer valid, we would need to make bigger changes, I am not sure what that would need to look like.

This PR doesn't affect your build system. This cannot hurt, just help in case there's a name clash either by coincidence or by design.

dlopr:
Note that `muxc` already has a `--device-idx` option.

dlopr:
And also note that it may be more convenient to select a device by index than by name regardless of potential name clashes.

hvdijk:
> Note that `muxc` already has a `--device-idx` option.

That is a strong argument in favour of your PR and I'm okay in principle with it on that basis, but note that the option there is not to deal with duplicate device names. I think we still have to consider duplicate device names an error.

modules/cargo/include/cargo/argument_parser.h
```diff
@@ -219,7 +219,7 @@ class argument {
     cargo::array_view<cargo::string_view> Choices;
   };
   union {
-    bool *Bool;
+    bool *Bool = {};
```
hvdijk, line 222: What is the reason for this change?

dlopr:
The compiler gives a warning if left uninitilized (as it happens with the
custom parser overload of the constructor).

This could all be much better if implemented with std::variant instead of
this ad-hoc semi tagged-union implementation, in my opinion. But that's a
different discussion.

hvdijk:
> The compiler gives a warning if left uninitilized (as it happens with the custom parser overload of the constructor).

Ah, thanks, that makes sense. As you've seen, the idea is that `Type` tells which union member is active, but then we inexplicably don't put `CustomHandler` in that union despite that being controlled by `Type == CUSTOM`. That isn't consistent with the rest of the code, this seems like a good time to fix that.

dlopr:
That's a tagged-union. But in its basic form, it has problems. A plain union won't have an implicitly defined destructor if one of its members has a user-defined destructor. That's why simply adding CustomHandler to the union doesn't work.

std::variant is in fact an advanced implementation of the tagged-union idea that correctly handles that sort of thing. That's why I suggested to use it instead of the plain union.

That should be in a different PR, though.

hvdijk:
If now is not a good time to do that, then I don't know when :) Does https://github.com/hvdijk/oneapi-construction-kit/commit/cargo-argument-variant avoid that problem? If it does, I can put that up as a pull request first, or you can rebase your pull request on it and we can merge them together.

I notice that `muxc`'s `--device-idx` is zero-based, but your new `clc`'s `--device-idx` is one-based. Can you make sure that they work the same way?

dlopr:
That change is not related to this PR. You can do that after if you want.

The index being 1-based is because the list of available devices that gets
printed in case of error is 1-based.

PR #800, opened 2025-04-30
Fix CMake target clash with Abseil; add “clik_tests”

aignermax:
…heck'

# Overview

Abseil 202401 introduces an interface target named check.
Our test driver used the same name, so CMake fails with CMP0002.
This patch adds a new target clik_tests that runs the existing Python test-runner and leaves Abseil’s check untouched.

# Reason for change

With the current main branch + Abseil 202401 the build aborts during the configure step, so no one can compile the Construction Kit on an up-to-date toolchain.

# Description of change

* Replace add_custom_target(check …) with add_custom_target(clik_tests …).
* clik_tests still depends on ClikExamples and the selected HAL.
* Tests are now launched with ninja clik_tests.
No other targets, flags or APIs are affected.

Verified on Ubuntu 22.04 with GCC 11, CMake 3.26, Ninja 1.11;
ninja clik_tests completes and all tests pass.

# Checklist

*  Project builds successfully with the change
*  Tests run locally (ninja clik_tests)
* clang-format-19 run on modified files
* Code of Conduct respected

### DIFF
--- a/clik/README.md
+++ b/clik/README.md
@@ -45,7 +45,7 @@ ComputeMux examples can optionally be built by passing additional options to CMa
 Examples included with clik can be used as a test suite for basic validation:
 
     $ cd path/to/clik/build/clik-debug
-    $ ninja check
+    $ ninja clik_tests
     [100 %] [0:0:11/11] PASS matrix_multiply_tiled
 
     Passed:           11 (100.0 %)
--- a/clik/test/CMakeLists.txt
+++ b/clik/test/CMakeLists.txt
@@ -15,12 +15,12 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 set(SCRIPT_PATH ${CMAKE_CURRENT_SOURCE_DIR}/run_tests.py)
-add_custom_target(check
+add_custom_target(clik_tests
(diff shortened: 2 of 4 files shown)


Zebsi235:
Thanks for the detailed feedback and suggestions!

To clarify the context a bit further: we’re using Abseil alongside gRPC in the oneAPI Construction Kit as part of our HAL development process for simulated devices. Specifically, we followed the HAL REFSI tutorial and extended it with our own functionality (Memory access to our simulated device is handled over gRPC). The clik examples work fine with our HAL implementation. When progressing to the ComputeMux stage, which builds more broadly, the CMake configuration fails due to both Abseil and the OCK's test driver defining a target called check.

“Are you including clik in an external project that also contains Abseil, or are you including Abseil in clik?”
We're including Abseil into clik, rather than the other way around.

“Would something like that work for clik as well?”
I agree that a conditional approach could work and might offer a better developer experience. We actually attempted something similar—renaming the target to clik_tests and trying to hook it conditionally—but it led to cascading issues, possibly because we used a global search-and-replace rather than validating each reference manually across the repo which would have required a more detailed manual update across the repo than we had capacity for at the time.

Since we are not using gRPC directly in the OCK anymore, and therefore abseil is also not needed, this issue is not relevant for us anymore, yet we are available for questions if you think we can still provide useful information. 

Thanks again for your response!

PR #801, opened 2025-04-30
Split sycl cts builds into subsets and move build jobs to Github runners

alan-forbes-cp:
# Overview

Split sycl cts builds into subsets and move build jobs to Github runners

# Reason for change

Improves build times; expands use of Github runners.

# Description of change

- SYCL-CTS build jobs are now split into 3 subsets (A,B,C) and matrix-driven.
- Jobs now run on Github runners.
- Specification of test categories for each subset are contained in 3 new "subset list" files. These are defined to split the existing build time into 3 * 33.3% chunks (approx.).
- New "combination" jobs assemble the 3 * test build outputs into a single artifact for downstream processing (testing).
- New `combine_sycl_cts_artifacts` action handles the downloading, assembly and uploading of the combined artifact.
- Existing build job is passed a new `subset` input to identify each build subset.
- Build job creates an "exclude" list on the fly for each subset i.e. if building subset A, then the exclude list contains categories B and C.
- A new `set build SYCL CTS env vars and outputs` build step consolidates the env vars and outputs needed for the build.
- Downloads from previous workflows operate as before - only now with 3 * subsets per run.

- `cat_all.lst` file is included as an "at time of writing" test category reference upon which the 3 * subsets were based.
- Additional dev/test config in the `planned_testing` file will be removed and the original restored.
- Testing is on-going. Needs a full test across all targets + opencl/native_cpu plus same again with downloads.
- There is a known issue (I now know) of Github workflow UI failing to display all the artifacts correctly in some situations (shows the correct summary number - but fails to display all). This is under separate investigation.


.github/workflows/run_ock_external_tests.yml
```diff
@@ -366,9 +371,45 @@ jobs:
...
+        with:
+          os: 'ubuntu'
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: combine sycl cts artefacts
+        shell: bash
+        run: |
+          git config --global --add safe.directory $PWD
+          mkdir tests_dir
+          for EXCLUDE_SUBSET in A B C ; do
+            gh run download ${{ github.run_id }} -n sycl_cts_opencl_host-x86_64_linux_${EXCLUDE_SUBSET} -D ${{ github.workspace }}
```
coldav, line 401: action/download should be used ideally, since we know the splits we could call it multiple times. If we want to go down the gh run download method looped you might as well move it the download action and call it directly.

gh run download is slower though.

alan-forbes-cp: fyi  - Its looking like we'll have 3 subsets so that translates to 3 * calls obviously. I'll migrate to that.

.github/workflows/run_ock_external_tests.yml
```diff
@@ -345,9 +345,14 @@ jobs:
...
           download_dpcpp_artefact: ${{ inputs.download_dpcpp_artefact }}
 
-  build_sycl_cts_x86_64_opencl:
+  build_sycl_cts_x86_64_opencl_subsets:
     needs: [workflow_vars, build_icd, build_dpcpp_native_x86_64]
-    runs-on: 'cp-ubuntu-24.04'
+    strategy:
+      fail-fast: false  # let all matrix jobs complete
+      matrix:
+        exclude_subset: [ 'A', 'B', 'C' ]
```
coldav, line 353: naming A,B,C is slightly confusing to me but in principal this methodology and below seems okay,

scripts/testing/sycl_cts/tests_subset_A.csv
```diff
@@ -0,0 +1,68 @@
+accessor_basic
```
coldav, line 1: I think it might make more sense to name it as reflecting what it excludes rather than A,B,C, as find it difficult to see from the list.
e.g. if the first one was reflecting A->F, then I'd call it _A_to_F
Then iterating through is more clear.

At a push you could autogenerate it based on the letter range e.g.
ls -d */ | grep '^[a-f]' | sed 's/\///' would list all the directories starting with a-f.

.github/workflows/run_ock_external_tests.yml
```diff
@@ -462,12 +605,45 @@ jobs:
         uses: ./.github/actions/do_build_sycl_cts
         with:
           target: host_riscv64_linux
+          exclude_subset: ${{ matrix.exclude_subset }}
           download_sycl_cts_artefact: ${{ inputs.download_sycl_cts_artefact }}
           sycl_device: opencl
 
+  build_sycl_cts_riscv64_opencl_combine:
+    needs: [build_sycl_cts_riscv64_opencl]
+    runs-on: 'ubuntu-24.04'
```
coldav, line 614: given it's simplicity it neither needs a container or needs to run on ubuntu-24.04 - The container takes about a minute - which isn't needed.

.github/workflows/run_ock_external_tests.yml
alan-forbes-cp: A,B,C naming has been retained  - but I've no objections of renaming to something more meaningful if a scheme can be found.

scripts/testing/sycl_cts/tests_subset_A.csv
alan-forbes-cp: Final excluded test lists are now generated on-the-fly using (manually created) subset lists. i.e. for subset A we are excluding subsets B and C.
See related comment elsewhere re. naming issues.

.github/workflows/run_ock_external_tests.yml
alan-forbes-cp: Combine jobs form and content has now been finalised (for review purposes) and the container reference removed.

.github/workflows/planned_testing_caller_19.yml
```diff
@@ -22,20 +22,23 @@ jobs:
       use_llvm_github_cache: true
       # We can set ock, test_sycl_cts etc here optionally if this is used as a pull request.
       # Any with parameters below this is intended for local testing and should not be merged.
-      # target_list: '[ "host_x86_64_linux" ]'
+      #target_list: '[ "host_x86_64_linux" ]'
```
coldav, line 25: Remember to restore this bit

coldav (approved):

PR #807, opened 2025-05-02
Added refsi_tutorial as an internal test

coldav:
# Overview

 Added refsi_tutorial as an internal test
    
This tests the "end" part of refsi_tutorial as a PR job. "start" is to follow in a separate PR.

# Reason for change

Test not yet ported over. We were getting breakages from changes.

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/.github/actions/do_build_ock/action.yml
+++ b/.github/actions/do_build_ock/action.yml
@@ -110,6 +110,9 @@ inputs:
   toolchain_file:
     description: "Path to toolchain file"
     default: ""
+  source_dir:
+    description: 'cmake source directory'
+    default: '.'
 
 runs:
   # We don't want a new docker just a list of steps, so mark as composite
@@ -168,7 +171,7 @@ runs:
               -DCA_BUILD_32_BITS=${{ inputs.build_32_bit }}
               -DCMAKE_TOOLCHAIN_FILE=${{ inputs.toolchain_file }}
                ${{ inputs.extra_flags }}
-               .
+               ${{ inputs.source_dir }}
     - name: build_ock
       shell: ${{ inputs.shell_to_use }}
       run:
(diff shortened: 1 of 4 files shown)


.github/actions/do_build_ock/action.yml
```diff
@@ -110,6 +110,9 @@ inputs:
   toolchain_file:
     description: "Path to toolchain file"
     default: ""
+  cmake_path:
```
hvdijk, line 113: This is a confusing variable name, it doesn't specify the path to CMake, it specifies the path to pass to CMake as the source directory. Maybe `source_dir` would work?

.github/actions/do_build_pr/run_ubuntu_gcc_x86_64_refsi_tutorial/action.yml
```diff
@@ -0,0 +1,39 @@
...
+runs:
+  using: "composite"
+  steps:
+      # TODO: refsi_tutorial start
+      - name: remove any old dirs and set up new target
+        shell: bash  
+        run: |
+          rm -rf build
+          git config --global --add safe.directory $GITHUB_WORKSPACE         
+          pip install cookiecutter
```
hvdijk, line 18: Is this getting installed globally or in a venv? If globally, can you change this to use a venv so that it does not result in errors when we move to Ubuntu 24?

.github/actions/do_build_ock/action.yml
coldav: that makes sense yes.

.github/actions/do_build_pr/run_ubuntu_gcc_x86_64_refsi_tutorial/action.yml
coldav: agreed

hvdijk (approved):

PR #816, opened 2025-05-09
Migrate riscv MR jobs to Github

alan-forbes-cp:
# Overview

Migrate remaining Gitlab MR riscv jobs to Github.

# Reason for change

Incremental migration of MR jobs - 2 riscv in this case.

# Description of change

- 2 new riscv jobs added to `run_ock_internal_tests`.
- new action added for the 2 jobs with OCK build settings and inputs as required.
- add action call to the to the cache seeding.
- remove `run_riscv_m1` job, action and cache seeding - no longer required.
- `run_cities` testing is not included (will be handled by separate ticket, with placeholders included here as comments for reference.)

- PR testing shows jobs completing with identical results to those seen on Gitlab.


coldav:
As before you need to add to the cache seeding when you make this change.

coldav:
Also now this is added you can remove the risc-v one with no MR mapping `run_riscv_m1`

.github/actions/do_build_pr/run_ubuntu_gcc_x86_64_riscv_cl3_0/action.yml
```diff
@@ -0,0 +1,35 @@
+name: build_pr_ubuntu_gcc_x86_64_riscv_cl3-0
```
coldav, line 1: I think since these are part1/2 and they are so close to being identical, it would make sense to coalesce them them here with a flag.
We also might want to consider whether they really need to be two jobs, but that can wait until later, but a single action will make that step easier.
From what I can tell the only differences are:
build_target
disable_unitcl_vecz_checks

I think from a seeding perspective we'd only need to call a single version to build the ccache.

.github/workflows/run_ock_internal_tests.yml
```diff
@@ -62,7 +62,8 @@ jobs:
 
   # build and run host x86_64, execute UnitCL and lit tests and build and run offline
   run_host_x86_64:
-    if: contains(inputs.target_list, 'host_x86_64_linux')
+    #if: contains(inputs.target_list, 'host_x86_64_linux')
+    if: false
```
alan-forbes-cp, line 66: Note: Remove all `if:` clauses added to limit dev testing

.github/actions/do_build_pr/run_ubuntu_gcc_x86_64_riscv_cl3_0/action.yml
alan-forbes-cp: Merged the actions and added seeding.

.github/workflows/run_ock_internal_tests.yml
```diff
@@ -285,9 +260,65 @@ jobs:
...
+      uses: actions/checkout@v4
+    - name: setup-ubuntu
+      uses: ./.github/actions/setup_build
+      with:
+        llvm_version: ${{ inputs.llvm_current }}
+        llvm_source: ${{ inputs.llvm_source}}
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+    - name: build and run ock
+      uses: ./.github/actions/do_build_pr/run_ubuntu_gcc_x86_64_riscv_cl3_0
+    # For now DO NOT include run_cities.py testing. Commands/sub-commands used in Gitlab are:
```
coldav, line 283: ideally we wouldn't even mention gitlab and just say todo: add limited run-cities or even just put it in the jira. This and elsewhere just make it look untidy.

coldav (approved):

.github/workflows/run_ock_internal_tests.yml
alan-forbes-cp: Done

.github/workflows/run_ock_internal_tests.yml
alan-forbes-cp: Comments (new and existing) updated as suggested

PR #821, opened 2025-05-15
Fix test_api OpenCL CTS test for zero_sized_enqueue

coldav:
# Overview

This fixes up the error checking to allow for global size being a nullptr and adds some UnitCL testing for both the normal clEnqueueNDRangeKernel and clCommandNDRangeKernelKHR.

# Reason for change

opencl cts test_api/zero_sized_enqueue was failing due to passing a NULL pointer for the global size. For this being an error, this
has been deprecated since 2.1.


source/cl/test/UnitCL/source/cl_khr_command_buffer/clCommandNDRangeKernelKHR.cpp
```diff
@@ -72,6 +72,55 @@ TEST_F(CommandNDRangeKernelTest, EmptyKernel) {
   ASSERT_SUCCESS(clReleaseProgram(program));
 }
 
+// Tests the whether we can enqueue an execute a command buffer containing a
+// kernel enqueue created via clCommandNDRangeKernelKHR with global size of 0
+// or a global size of nullptr
+TEST_F(CommandNDRangeKernelTest, GlobalZero) {
```
hvdijk, line 78: The corresponding `clEnqueueNDRangeKernel` test is called `ZeroNDRange`, can we use the same name here? And the difference here where we also check that the kernel is not executed, can we use that in the `clEnqueueNDRangeKernel` test as well?

source/cl/test/UnitCL/source/cl_khr_command_buffer/clCommandNDRangeKernelKHR.cpp
```diff
@@ -72,6 +72,55 @@ TEST_F(CommandNDRangeKernelTest, EmptyKernel) {
...
+  // Requires a compiler to compile the kernel.
+  if (!getDeviceCompilerAvailable()) {
+    GTEST_SKIP();
+  }
+  // Set up the kernel.
+  const char *code = R"OpenCLC(
+  __kernel void empty_kernel() { __builtin_trap(); }
```
hvdijk, line 85: I know I was the one who suggested this, but I forgot that UnitCL is meant to be portable across OpenCL implementations and `__builtin_trap` is not required to be supported. We could put this in an `#ifdef __clang__` or such?

source/cl/test/UnitCL/source/cl_khr_command_buffer/clCommandNDRangeKernelKHR.cpp
```diff
@@ -72,6 +72,55 @@ TEST_F(CommandNDRangeKernelTest, EmptyKernel) {
...
+  ASSERT_SUCCESS(error);
+
+  constexpr size_t global_size = 0;
+  ASSERT_SUCCESS(clCommandNDRangeKernelKHR(
+      command_buffer, nullptr, nullptr, kernel, 1, nullptr, &global_size,
+      nullptr, 0, nullptr, nullptr, nullptr));
+
+  ASSERT_SUCCESS(clCommandNDRangeKernelKHR(
+      command_buffer, nullptr, nullptr, kernel, 1, nullptr,
+      /* global size */ nullptr, nullptr, 0, nullptr, nullptr, nullptr));
```
hvdijk, line 111: Let's use the LLVM convention here that I personally would prefer us to use more consistently:
```suggestion
      /*global_work_size=*/nullptr, nullptr, 0, nullptr, nullptr, nullptr));
```

hvdijk (approved):

PR #822, opened 2025-05-16
Remove CA_ENABLE_API as a CMake option

coldav:
# Overview

Remove CA_ENABLE_API as a CMake option

# Reason for change

We only support OpenCL as a an API now, so it makes sense to remove this option.

### DIFF
--- a/cmake/AddCA.cmake
+++ b/cmake/AddCA.cmake
@@ -1053,10 +1053,7 @@ function(add_ca_copy_file)
     COMMENT "Copying file ${relOut}")
 endfunction()
 
-if(EXISTS ${PROJECT_SOURCE_DIR}/source/cl AND
-    (CA_ENABLE_API STREQUAL "" OR CA_ENABLE_API MATCHES cl))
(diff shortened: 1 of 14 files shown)


cmake/AddCA.cmake
```diff
@@ -1053,10 +1053,7 @@ function(add_ca_copy_file)
     COMMENT "Copying file ${relOut}")
 endfunction()
 
-if(EXISTS ${PROJECT_SOURCE_DIR}/source/cl AND
-    (CA_ENABLE_API STREQUAL "" OR CA_ENABLE_API MATCHES cl))
-  # Cleared to ensure reconfigures behave correctly.
-  set(CA_CL_RUNTIME_EXTENSION_TAGS ""
+set(CA_CL_RUNTIME_EXTENSION_TAGS ""
```
hvdijk, line 1056: There is a larger block that is now no longer inside `if()`, can you unindent the whole block?

coldav: done

hvdijk (approved):

PR #829, opened 2025-05-23
Added e2e native cpu xfail overrides for new FreeFunctionKernels fails

coldav:
# Overview

and consequences.*

# Reason for change

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/scripts/testing/sycl_e2e/override_native_cpu.csv
+++ b/scripts/testing/sycl_e2e/override_native_cpu.csv
@@ -83,3 +83,8 @@ SYCL,KernelAndProgram/spec_constants_after_link.cpp,XFail
 SYCL,KernelAndProgram/fp32-precise-fdiv.cpp,XFail
 SYCL,KernelAndProgram/multiple-kernel-linking.cpp,XFail
 SYCL,KernelAndProgram/kernel-bundle-get-kernel.cpp,XFail
+SYCL,FreeFunctionKernels/range_as_kernel_parameter.cpp,XFail
+SYCL,FreeFunctionKernels/id_as_kernel_parameter.cpp,XFail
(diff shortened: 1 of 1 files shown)


scripts/testing/sycl_e2e/override_native_cpu.csv
```diff
@@ -83,3 +83,8 @@ SYCL,KernelAndProgram/spec_constants_after_link.cpp,XFail
 SYCL,KernelAndProgram/fp32-precise-fdiv.cpp,XFail
 SYCL,KernelAndProgram/multiple-kernel-linking.cpp,XFail
 SYCL,KernelAndProgram/kernel-bundle-get-kernel.cpp,XFail
+SYCL,FreeFunctionKernels/range_as_kernel_parameter.cpp,XFail
+SYCL,FreeFunctionKernels/id_as_kernel_parameter.cpp,XFail
+SYCL,FreeFunctionKernels/vec_as_kernel_parameter.cpp,XFail
+SYCL,FreeFunctionKernels/marray_as_kernel_parameter.cpp,XFail
+SYCL,FreeFunctionKernels/usm_compatibility.cpp,XFail
```
hvdijk, line 90: As this file is getting bigger, it probably makes sense to use some sensible order rather than adding new failures to the end. (But also happy with that as a separate NFC PR after this has gone in, if you prefer.)

hvdijk (approved):

scripts/testing/sycl_e2e/override_native_cpu.csv
coldav: I'll do that separately, since it makes it harder to see this.

coldav: true, lets do that later.

PR #848, opened 2025-06-03
Drop implicit permissions.

hvdijk:
# Overview

Drop implicit permissions.

# Reason for change

The planned testing workflows were not setting permissions, causing them to run with more permissions than intended.

# Description of change

points, etc.*

# Checklist

### DIFF
--- a/.github/workflows/planned_testing_caller.yml
+++ b/.github/workflows/planned_testing_caller.yml
@@ -66,6 +66,9 @@ on:
         type: boolean
         default: true
 
+permissions:
+  packages: read
+
 jobs:
   build_llvm_artefacts:
     name: Call PR testing on schedule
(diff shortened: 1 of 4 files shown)


.github/workflows/planned_testing_caller.yml
```diff
@@ -66,6 +66,9 @@ on:
         type: boolean
         default: true
 
+permissions:
```
coldav, line 69: This workflow is never called except by the 19/20 ones - do we need both?

hvdijk: Really this shouldn't be in this directory for precisely the reason you say. Since it's in this directory though, it gets checked for excessive permissions just in case we somehow do manage to call it directly.

PR #852, opened 2025-06-04
Simplify dockers, including removing vulkan sdk and clang-tidy

coldav:
These can be simplified to be useful for the vast majority to jobs, but still have most of the useful installations present, so installing things will still be mostly not required.

# Overview

Simplify dockers, including removing vulkan sdk and clang-tidy

# Reason for change

The current dockers take about 1m20s to load.  Vulkan SDK is also no longer needed, along with llvm 18.

# Description of change

Removed a lot of installs, particularly llvm ones that only apply to single jobs like clang-tidy. Where it doesn't add too much to the docker and is used in multiple jobs, I've tried to keep it with the aim of having the load under 25 seconds. https://github.com/uxlfoundation/oneapi-construction-kit/pull/819/ shows approximately what would be required beyond now to work with these docker changes.

### DIFF
--- a/.github/dockerfiles/Dockerfile_24.04-x86-64
+++ b/.github/dockerfiles/Dockerfile_24.04-x86-64
@@ -1,12 +1,8 @@
 FROM ubuntu:24.04
 
 RUN apt-get update 
-RUN apt-get install -y git git-crypt wget curl jq xz-utils build-essential zlib1g-dev lsb-release libssl-dev
-# Set the timezone, required to avoid hanging on input
-ENV TZ=Europe/London
-RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+RUN apt-get install -y wget git build-essential
 
-# Enable ability to install foreign packages for cross compilation
 RUN dpkg --add-architecture i386
 RUN dpkg --add-architecture arm64
 RUN dpkg --add-architecture riscv64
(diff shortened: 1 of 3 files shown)


.github/dockerfiles/Dockerfile_24.04-x86-64
```diff
@@ -1,52 +1,33 @@
...
-# Add VulkanSDK latest package repository
-RUN wget -qO - https://packages.lunarg.com/lunarg-signing-key-pub.asc | apt-key add -
-RUN wget -qO /etc/apt/sources.list.d/lunarg-vulkan-noble.list https://packages.lunarg.com/vulkan/lunarg-vulkan-noble.list
-# Sync the two above package repositories
 RUN apt-get update
 
-# Add clang-tidy
 RUN wget -qO - https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor - | tee /usr/share/keyrings/llvm-archive-keyring.gpg >/dev/null
-RUN echo 'deb [signed-by=/usr/share/keyrings/llvm-archive-keyring.gpg] http://apt.llvm.org/noble/ llvm-toolchain-noble-18 main' | tee /etc/apt/sources.list.d/llvm.list >/dev/null
 RUN echo 'deb [signed-by=/usr/share/keyrings/llvm-archive-keyring.gpg] http://apt.llvm.org/noble/ llvm-toolchain-noble-19 main' | tee -a /etc/apt/sources.list.d/llvm.list >/dev/null
```
hvdijk, line 16: Likewise:
```suggestion
RUN echo 'deb [signed-by=/usr/share/keyrings/llvm-archive-keyring.gpg] http://apt.llvm.org/noble/ llvm-toolchain-noble-19 main' | tee /etc/apt/sources.list.d/llvm.list >/dev/null
```

PR #880, opened 2025-07-01
 Add "known" tests for e2e so we don't regress when new tests are added

coldav:
# Overview

 Add "known" tests for e2e so we don't regress when new tests are added

# Reason for change

Avoid keep getting hit by new tests breaking our runs.

# Description of change

Combined known.csv which is a list of tests currently in lit with the overrides currently existing for e2e


.github/workflows/planned_testing_caller_19.yml
```diff
@@ -24,23 +24,23 @@ jobs:
       use_llvm_github_cache: true
       # We can set ock, test_sycl_cts etc here optionally if this is used as a pull request.
       # Any with parameters below this is intended for local testing and should not be merged.
-      # target_list: '[ "host_x86_64_linux" ]'
```
coldav, line 27: To be removed before merging.

hvdijk (approved):

PR #890, opened 2025-07-08
Split some risc-v jobs into two and run on standard github runner

coldav:
# Overview

Split dome risc-v jobs into two and run on github runners

# Reason for change

Switch to using github runners to reduce self-hosted dependence.

### DIFF
--- a/.github/actions/do_build_opencl_cts/action.yml
+++ b/.github/actions/do_build_opencl_cts/action.yml
@@ -53,26 +53,22 @@ runs:
         done
         # do build
         set -x
+        mkdir $GITHUB_WORKSPACE/build
+        cd $GITHUB_WORKSPACE/build
         cmake -G Ninja $CMAKE_TOOLCHAIN $GITHUB_WORKSPACE/OpenCL-CTS \
           -DCMAKE_BUILD_TYPE=Release \
           -DOPENCL_LIBRARIES=OpenCL \
           -DCL_INCLUDE_DIR=$GITHUB_WORKSPACE/install_headers/include \
           -DCL_LIB_DIR=$GITHUB_WORKSPACE/install_icd/lib
         ninja -v
+        # remove some files to reduce the size and complexity of the upload
+        # (internal upload does not support filtering out)
+        find $GITHUB_WORKSPACE/build/test_conformance \( -name "*.o" -o -name "*.cmake" \) -delete
 
     - name: upload opencl cts artifacts
-      uses: actions/upload-artifact@v4
+      uses: ./.github/actions/upload_artifact
       with:
         name: opencl_cts_${{inputs.target}}
-        path: |
-          test_conformance
-          !test_conformance/**/.*
-          !test_conformance/**/CMakeCache.txt
-          !test_conformance/**/CMakeFiles
-          !test_conformance/**/CMakeFiles/**
-          !test_conformance/**/*.cmake
-          !test_conformance/**/*.ninja
-          !test_conformance/test_common
-          !test_conformance/test_common/**
+        path: build/test_conformance
         retention-days: 7
-
+        needs_tar: 'true'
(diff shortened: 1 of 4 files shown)


hvdijk:
The description doesn't look right, "Switch to using internal runners.", isn't this moving away from internal runners?

.github/workflows/run_ock_external_tests.yml
```diff
@@ -293,6 +293,70 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          # scripts: for run_cities.py
+          # source: for CTS filter.csv files
+          # platform: for toolchain files
+          sparse-checkout: |
+            scripts
+            source
+            platform
+            .github
```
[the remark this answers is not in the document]
coldav: sorry I meant to put this in draft:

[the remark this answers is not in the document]
coldav: but you need it to call actions which are under .github

.github/workflows/run_ock_external_tests.yml
```diff
@@ -298,6 +298,87 @@ jobs:
...
+      - name : build and upload opencl_cts
+        uses: ./.github/actions/do_build_opencl_cts
+        with:
+          target: host_riscv64_linux
+
+  run_opencl_cts_riscv64_low:
```
alan-forbes-cp, line 326: As the `needs:` are the same for high/low job variants could we not just have a single job with matrix to avoid the duplication? Apologies if I'm missing something here.

.github/actions/do_build_opencl_cts/action.yml
```diff
@@ -53,26 +53,22 @@ runs:
...
+        cd $GITHUB_WORKSPACE/build
         cmake -G Ninja $CMAKE_TOOLCHAIN $GITHUB_WORKSPACE/OpenCL-CTS \
           -DCMAKE_BUILD_TYPE=Release \
           -DOPENCL_LIBRARIES=OpenCL \
           -DCL_INCLUDE_DIR=$GITHUB_WORKSPACE/install_headers/include \
           -DCL_LIB_DIR=$GITHUB_WORKSPACE/install_icd/lib
         ninja -v
+        # remove some files to reduce the size and complexity of the upload
+        # (internal upload does not support filtering out)
+        find $GITHUB_WORKSPACE/build/test_conformance \( -name "*.o" -o -name "*.cmake" \) -delete
```
hvdijk, line 66: If we need this then fine, but why do we? I get that you switched to `.github/actions/upload_artifact` in order to get that `needs_tar` option but why do we need tar?

coldav: it needs it to keep the executable status. We've had to do that elsewhere for the same reason - see https://github.com/actions/upload-artifact?tab=readme-ov-file#permission-loss

hvdijk: Ah, thanks

.github/actions/run_opencl_cts/action.yml
```diff
@@ -41,6 +58,12 @@ runs:
         python3 scripts/testing/create_override_csv.py -d scripts/testing/opencl_cts \
              -k ${{ inputs.target }} llvm_${{ inputs.llvm_version }} \
              -o override_combined.csv -vv
+        CTS_CSV_FILE_PATH=$GITHUB_WORKSPACE/test_conformance/$CTS_CSV_FILE
+        if [ "${{ inputs.split_index }}" != "" ]; then
+          split -n l/2 $GITHUB_WORKSPACE/test_conformance/$CTS_CSV_FILE --additional-suffix=.csv -d 0
+          CTS_CSV_FILE_PATH=00${{ inputs.split_index }}.csv
+          echo Using $CTS_CSV_FILE
```
hvdijk, line 65: I assume you meant to print `$CTS_CSV_FILE_PATH` here?

.github/workflows/run_ock_external_tests.yml
coldav: good idea and done

alan-forbes-cp (approved):
Looks good - I'll rebase to pick up your artifact handling changes.

PR #894, opened 2025-07-10
Remove any unnecessary ninja installs

coldav:
# Overview

Ninja is in the docker, so no need for extra installs.

### DIFF
--- a/.github/actions/do_build_dpcpp/action.yml
+++ b/.github/actions/do_build_dpcpp/action.yml
@@ -26,9 +26,6 @@ runs:
       with:
         target: ${{ inputs.target }}  
 
-    - name: Install Ninja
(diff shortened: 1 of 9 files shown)


.github/workflows/planned_testing_caller_20.yml
```diff
@@ -59,3 +59,4 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           cache_prefixes: "ccache-ccache-dpcpp-build-host_x86_64_linux ccache-ccache-dpcpp-build-host_aarch64_linux ccache-ccache-dpcpp-build-host_riscv64_linux"
+# DELETE THIS LINE
```
coldav, line 62: To be removed

PR #911, opened 2025-07-16
Add testing for llvm 21 and full main

coldav:
# Overview

Add testing for llvm 21

# Reason for change

llvm 21 is branched


hvdijk:
This changes things so that we test LLVM 19 two days, LLVM 20 two days, LLVM 21 three days of the week. But LLVM 21 is not a version we officially support yet, so it seems strange to make that the version we test most. Could we do LLVM 19 three days of the week, LLVM 20 and LLVM 21 two days each?

PR #917, opened 2025-07-17
Remove ock demo

coldav:
# Overview

This demo is out of date and has not been run for a long time so is being removed.


coldav:
> The `.github` part of this looks fine. As for the `examples/technical_blogs/ock_demo_blog` removal, I don't have strong feelings, but are we sure we want to remove it and not just add a disclaimer that it is outdated but kept just in case?

Could do, I guess there is some useful things in there.

coldav:
> We should also remove `.github/workflows/run_ock_demo.yml`

Done

hvdijk (approved):

PR #946, opened 2025-08-15
[LLVM 22] A few more lifetime fixes

coldav:
# Overview

[LLVM 22] A few more lifetime fixes.

# Description of change

* Obtain lifetime_(start|end) pointer argument by getting last argument.
* Do not update size argument when there is no size argument.
* Update tests.

### DIFF
--- a/modules/compiler/vecz/source/transform/packetizer.cpp
+++ b/modules/compiler/vecz/source/transform/packetizer.cpp
@@ -2050,22 +2050,34 @@ ValuePacket Packetizer::Impl::packetizeCall(CallInst *CI) {
     auto IntrID = Intrinsic::ID(Callee->getIntrinsicID());
     if (IntrID == llvm::Intrinsic::lifetime_end ||
         IntrID == llvm::Intrinsic::lifetime_start) {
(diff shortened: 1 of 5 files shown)


modules/compiler/vecz/source/transform/packetizer.cpp
```diff
@@ -2050,25 +2051,33 @@ ValuePacket Packetizer::Impl::packetizeCall(CallInst *CI) {
...
-                  : allocSize.getKnownMinValue() * SimdWidth.getKnownMinValue();
-          CI->setOperand(
-              0, ConstantInt::get(CI->getOperand(0)->getType(), lifeSize));
-          results.push_back(CI);
+#if LLVM_VERSION_LESS(22, 0) || 1
+      // LLVM 22 drops the size parameter
+      // Remove LifetimeHasSizeArg once intel llvm no longer has the extra
+      // argument
+      if (multi_llvm::LifeTimeHasSizeArg()) {
+        auto *ptr = CI->getOperand(1);
```
uwedolinsky, line 2059: Perhaps the `llvm::findAllocaForValue` function could be used here instead just in case the operand is more complex?

PR #966, opened 2025-08-29
Revert "[CI] Fix OpenCL-CTS to always use last good version"

hvdijk:
# Overview

Revert "[CI] Fix OpenCL-CTS to always use last good version"

# Reason for change

# Description of change

This reverts commit f2aed47ff9ed74cefa5384b7de5ada2e9a08f21d.

# Checklist

* Run [clang-format-20](https://clang.llvm.org/docs/ClangFormat.html) on all modified code.


.github/README.md
```diff
@@ -123,18 +123,7 @@ Further information on `runs-on:` can also be found in the [github docs](https:/
...
-        description: 'Git checkout ref for OpenCL Headers repo'
-        default: 'v2025.06.13'
-      icd_loader_ref:
-        description: 'Git checkout ref for OpenCL ICD Loader repo'
-        default: 'v2024.10.24'
-      opencl_cts_ref:
-        description: 'Git checkout ref for OpenCL-CTS repo'
-        default: 'v2025-04-21-00'
-```
+There are a number of inputs to this workflow which relate to Git checkout references in OpenCL repos. The default values for these can be found in [the workflow](workflows/create_opencl_cts_artefact.yml).
```
coldav, line 126: I thought we still wanted fixed versions for the cts artifact?

hvdijk: We do, I just avoided the duplication of the version specification in the README and linked to the workflow where those values are specified instead. (I've not yet tested that these versions work, this PR is still a draft.)

PR #975, opened 2025-09-04
Fix overrides for opencl cts for a test_group_functions test

coldav:
# Overview

risc-v was passing test_group_functions "* group and sub-group joint scan functions*" so add xfails for the other arches.


hvdijk:
I think it has also sometimes passed on x86_64 and/or AArch64, and rather than moving the overrides like this, we should be marking it `MayFail` instead.

PR #984, opened 2025-09-10
Avoid +v warnings running OpenCL CTS.

hvdijk:
# Overview

Avoid +v warnings running OpenCL CTS.

# Reason for change

We were setting CA_HOST_TARGET_FEATURES=+v to enable RVV, but for non-RISC-V it makes no sense and results in warnings.

# Description of change

This commit changes that so that we set it only for RISC-V.

# Checklist

* Run [clang-format-20](https://clang.llvm.org/docs/ClangFormat.html) on all modified code.

### DIFF
--- a/.github/actions/run_opencl_cts/action.yml
+++ b/.github/actions/run_opencl_cts/action.yml
@@ -15,6 +15,9 @@ inputs:
   split_index:
     description: "optional split index so this run can be done with different splits, current only 0 or 1"
     default: ''
+  cityrunner_args:
+    description: "optional extra arguments to pass to cityrunner"
+    default: ''
 
 runs:
   using: "composite"
@@ -77,4 +80,4 @@ runs:
           -s $CTS_CSV_FILE_PATH \
           -i $GITHUB_WORKSPACE/source/cl/scripts/$CTS_FILTER \
           -o override_combined.csv \
-          -e CA_HOST_TARGET_FEATURES=+v
+          ${{ inputs.cityrunner_args }}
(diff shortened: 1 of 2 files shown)


.github/actions/run_opencl_cts/action.yml
```diff
@@ -15,6 +15,9 @@ inputs:
   split_index:
     description: "optional split index so this run can be done with different splits, current only 0 or 1"
     default: ''
+  extra_env:
```
coldav, line 18: I'd probably call this extra_flags, like we've done elsewhere (or extra_params or whatever), as it coul be other things to -e.

.github/workflows/run_ock_external_tests.yml
```diff
@@ -403,6 +403,39 @@ jobs:
           llvm_version: ${{ inputs.llvm_version }}
           split_index: ${{ matrix.split_index }}
 
+  run_opencl_cts_riscv64_rvv:
```
coldav, line 406: personally not sure we should bother with a +v/non +v at this stage. but I'll approve if you want to add it.

.github/actions/run_opencl_cts/action.yml
hvdijk: Makes sense, I'll change that, but after CI is done so that the current run doesn't get cancelled.

.github/workflows/run_ock_external_tests.yml
hvdijk: We talked about this offline, this was based on the assumption that we were only testing `+v` elsewhere, we're not, we're rather inconsistent in what we test, but if we're going to make changes to what we test, that can be separate from fixing the issue of `+v` being applied to other architectures so I'll change that part back and save that for possibly a later PR.

PR #1008, opened 2025-09-30
[LLVM 22] Adjust test for new optimisation.

hvdijk:
# Overview

[LLVM 22] Adjust test for new optimisation.

# Reason for change

LLVM 22 is able to optimise a reduce.add of a splat into a multiplication.

# Description of change

Adjust one test to allow this.

# Checklist

* Run [clang-format-21](https://clang.llvm.org/docs/ClangFormat.html) on all modified code.


hvdijk:
This is no longer enough, we now also have a compilation failure due to an API change, but it would still be good to get this PR in first.

coldav (approved):

